PR #1, opened 2017-01-05
Proposal: Initial networking

fabiand:
Proposal of how to do the initial VM networking.

### DIFF
--- a/design-proposals/initial-networking.md
+++ b/design-proposals/initial-networking.md
@@ -13,7 +13,10 @@ On the other hand we want to connect VMs to the NIC(s) of the pod in which they 
 To keep it simple and work with what we have today, we just handle a single NIC (thus multiple should
 not make a difference).
 
-The broder scope is that: On the long run we expect that we can use the networking provided to pods.
+Another benefit of this proposal is that we will connect to the veth of the pod - thus all the networking infrastructure beyong this point is not relevant to us. Which means we are independnt of it for now. It could be delivered by weave, flannel, or something else.
+The drawback is that we do not - yet - have the tight control over the real network connectivity.
+
+The broader scope is that: On the long run we expect that we can use the networking provided to pods.
 This proposal allows us to start working with this assumption.
 A separate discussion is how to get multiple NICs into a pod, but this is rather a kubernetes discussion taking place i.e. here https://github.com/kubernetes/kubernetes/issues/27398.
 
--- a/design-proposals/initial-networking.md
+++ b/design-proposals/initial-networking.md
@@ -13,7 +13,7 @@ On the other hand we want to connect VMs to the NIC(s) of the pod in which they 
 To keep it simple and work with what we have today, we just handle a single NIC (thus multiple should
 not make a difference).
 
-Another benefit of this proposal is that we will connect to the veth of the pod - thus all the networking infrastructure beyong this point is not relevant to us. Which means we are independnt of it for now. It could be delivered by weave, flannel, or something else.
+Another benefit of this proposal is that we will connect to the veth of the pod - thus all the networking infrastructure beyond this point is not relevant to us. Which means we are independnt of it for now. It could be delivered by weave, flannel, or something else.
 The drawback is that we do not - yet - have the tight control over the real network connectivity.
 
 The broader scope is that: On the long run we expect that we can use the networking provided to pods.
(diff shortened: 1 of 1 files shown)


fabiand:
The only thing that I wonder about is if this works.

We connect a VM in a L2 fashion to a veth which is IPAM'ed / L3.
It depends on the veth backing infra (flannel, weave, ovn) if this will yield any connectivity between to VMs.

Idea: We could - like rancher does - to setup a dhcp server on the macvtap connecting to the vm, this dhcp server would advertise the ip of the veth of the pod.

design-proposals/initial-networking.md
```diff
@@ -0,0 +1,64 @@
...
+### Use-case
+
+The primary usecase for this proposal are VMs running in pods with libvirtd running on the host
+or in a DaemonSet.
+It is an easy way to get the VM simply connected to the NIC available inside the pod where
+it is intended to run.
+
+## API
+
+This proposal will enhance the `VM` Resoufce to also accept interfaces like:
```
rmohr: By that you mean, `virt-handler` will add it, right? because the cluster does not know the interface

design-proposals/initial-networking.md
```diff
@@ -0,0 +1,64 @@
...
+
+The `virt-controller` would transfer this into:
+
+```
+    <interface type='direct'>
+      <source dev='veth0@mypod' mode='vepa'/>
+    </interface>
+```
+
+The `dev` name is now referring to the veth endpoint `veth0@mypod` in th ehost's network namespace.
```
rmohr: s/th e/the /

design-proposals/initial-networking.md
```diff
@@ -0,0 +1,64 @@
...
+
+## API
+
+This proposal will enhance the `VM` Resoufce to also accept interfaces like:
+
+```
+"interfaces": [
+  {
+    "source": {
+      "dev": "eth0",
```
rmohr: How do you plan to get the device name from inside the pod?

Or in other words, how reliable is it that the first interface in the pod is always starting with `eth0`?

Could you name a resource for a convention there?

design-proposals/initial-networking.md
```diff
@@ -0,0 +1,64 @@
...
+VMs need network connectivity. This proposal is about a first start in this direction.
+
+The working assumption is that libvirtd is running with a host-view, thus in the host's
+network namespace.
+On the other hand we want to connect VMs to the NIC(s) of the pod in which they are running.
+
+To keep it simple and work with what we have today, we just handle a single NIC (thus multiple should
+not make a difference).
+
+Another benefit of this proposal is that we will connect to the veth of the pod - thus all the networking infrastructure beyong this point is not relevant to us. Which means we are independnt of it for now. It could be delivered by weave, flannel, or something else.
```
rmohr: Note that his implementation would be would be limited to solutions which use veth pairs. It is also possible to attach a macvlan to a host device and then directly move it into another namespace without veths between.

design-proposals/initial-networking.md
```diff
@@ -0,0 +1,64 @@
...
+This proposal will enhance the `VM` Resoufce to also accept interfaces like:
+
+```
+"interfaces": [
+  {
+    "source": {
+      "dev": "eth0",
+      "mode": "bridge"
+      },
+    "type": "direct"
```
rmohr: If (and only if) we can't know the name of the device in the container, another `type` might be a good thing and not naming a `dev` at all?

design-proposals/initial-networking.md
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+}
+```
+
+The `dev` value above is refering to the `dev` name _inside_ the pod.
+Thus the `virt-controller` needs to translate this pod specific name into the host sided endpoint.
```
fabiand: Mistake - should hanve been the handler.

design-proposals/initial-networking.md
sungwonh: beyong -> beyond

design-proposals/initial-networking.md
rbradford: s/Resoufce/Resource/

fabiand:
Closing this in favor of kubevirt/kubevirt#279

PR #2, opened 2017-01-15
Proposal on how to bring a VM into a container

rmohr:


rmohr:
@fabiand @stu-gott @admiyo, thoughts on this? I think we can come up with something better later if we find something. This is really a blocker regarding to bringing KubeVirt out of the Vagrant and Demo VM.

design-proposals/vm-in-a-container.md
```diff
@@ -0,0 +1,186 @@
...
+    <emulator>/usr/local/bin/myemulator</emulator>
+</devices>
+<qemu:commandline>
+   <qemu:env name='kubevirt.io.vm.uuid' value='1234-5678-1234-1234'/>
+   <qemu:env name='kubevirt.io.vm.name' value='testvm'/>
+</qemu:commandline>
+</domain>
+```
+
+Note that we have to specify the qemu namespace to use these features.
```
fabiand: You mean the target namespace the qemu instance should live in?

design-proposals/vm-in-a-container.md
```diff
@@ -0,0 +1,186 @@
...
+    file before it starts something in that container 
+
+The secret can be delivered to the container through 
+[Kubernetes secrets](https://kubernetes.io/docs/user-guide/secrets/). The
+binary can receive the secret via libvirts <qemu:commandline/> tag as
+environment variable:
+
+```xml
+<qemu:commandline>
+   <qemu:env name='kubevirt.io.secret' value='3098tFJoswfwkjp4'/>
```
fabiand: What about aligning the namespace: `kubevirt.io.vm.secret`?

design-proposals/vm-in-a-container.md
```diff
@@ -0,0 +1,186 @@
...
+operations. This imposes one big security risk: The binary can start
+applications in other namespaces and cgroups (even the one of the host).
+
+A few things can be done to reduce this risk:
+
+ 1. Only allow the qemu user to execute the binary
+ 2. Introduce a shared secret between the VM target container and kubevirt
+ 3. This secret can be unique per pod creation and only valid for a specific
+    amount of time
+ 4. The binary checks that  the secret is present in the kubernetes metadata
```
fabiand: "the kubernetes metadata file" what file are you referring to?

design-proposals/vm-in-a-container.md
```diff
@@ -0,0 +1,186 @@
...
+`/etc/kubevirt/qemu-kube.conf`). In there the path to `/proc` can be specified.
+On the specified location, the host `/proc` can be mounted.
+
+### Delivering the binary
+
+At the end, the binary and the libvirt hook need to be present in the mount
+namespace where libvirt is running. They should be delivered via an init
+container to the host via `hostDir` mount (if libvirt is running on the host),
+or via an `emptyDir` mount to a container (if libvirt is running in a
+container).
```
fabiand: What about shipping it directly in the libvirtd container?

design-proposals/vm-in-a-container.md
```diff
@@ -0,0 +1,186 @@
...
+
+There are not many problems to be expected regarding to permissions since the
+binary can
+
+ * have the suid bit
+ * have permission `-rwxr-xr-x`
+
+when we implement the secret properly.
+
+### NUMA
```
fabiand: Let's keep numa out of this for now.
It deserves it's own proposal I'd say.

design-proposals/vm-in-a-container.md
```diff
@@ -0,0 +1,186 @@
...
+hook](https://libvirt.org/hooks.html#qemu) will check if the Domain XML has an
+emulator tag which points to the right binary.
+
+### Flow
+
+Pseudocode:
+
+```
+# Getting the pid os container runtime independent
+pid = $(ps aux  | grep virt-launcher | grep "-kubevirt.vm.uuid 1234-5678-1234-1234 | tr -s ' ' | cut -d " " -f2)
```
fabiand: We can also set the secret on the launcher pod using an environemtn variable.
Then we can scan all /proc/*/environ files which is already a map, and find the correct process.

design-proposals/vm-in-a-container.md
```diff
@@ -0,0 +1,186 @@
...
+
+Note that this means that the secret is exposed in the Domain XML. If we think
+that having a secret lifespan of e.g. one minute is not secure enough, the
+secret could be encrypted.
+
+To further improve security, a [libvirt
+hook](https://libvirt.org/hooks.html#qemu) will check if the Domain XML has an
+emulator tag which points to the right binary.
+
+### Flow
```
fabiand: Would be nice if the flow would cover

1. controller part (setting the secret?)
2. launcher part
3. handler part
4. binary part

design-proposals/vm-in-a-container.md
fabiand: Being picky: It would probably not work with rkt stage2-lkvm runtime, because there the launcher would be run in a VM.

design-proposals/vm-in-a-container.md
```diff
@@ -0,0 +1,186 @@
...
+  executeQemuInPidNsAndCgroupContext(pid)
+} else {
+  fail()
+}
+```
+
+### Running it inside a container
+
+To run the binary from inside a container, it needs access to the host `/proc`
+directory. The recommended implementaiton is, to have a config file for the
```
fabiand: I think it only needs to access it if the container it is launched in has a pid namespace.

If the surrounding container is using the host's pid namespace, then /proc should be the host's /proc.

design-proposals/vm-in-a-container.md
rmohr: Would be ok too for the start, the disadvantage is, that we would need to rebuild the libvirt container then. So far I only had to change the emulator when kubevirt introduced new features and not when we changed the libvirt configuration. It seems to be more influenced on how we integrate with kubernetes than how we integrate with libvirt. If we ship it for example with the virt-handler daemon set as init container, we would not have to role out or restart libvirt. But both solutions would work.

design-proposals/vm-in-a-container.md
rmohr: You are right

design-proposals/vm-in-a-container.md
rmohr: Sounds good to me. I will make one for numa and cpu pinning.

design-proposals/vm-in-a-container.md
rmohr: Further, virt-handler controlls the additional qemu commandline and environment flags, and therefore how the emulator is called. So for me it is more like a part of virt-handler and that it is more important that it is kept in-sync with virt-handler than with libvirt. What do you think?

fabiand:
Closing this according to discussion offline discussions.

PR #5, opened 2017-02-05
designs: PersistentVolume storage

fabiand:


design-proposals/pvDiskStorage.md
```diff
@@ -0,0 +1,132 @@
...
+
+this is used by the Vm in the following way:
+
+```yaml
+kind: VM
+spec:
+  domain:
+    devices:
+      disks:
+      - type: PersistentVolumeClaim
```
admiyo: Is this the complete set of fields?  Document should have a section that describes each of the acceptable  fields and where they map to on dom.xml or in the k8s system

design-proposals/pvDiskStorage.md
```diff
@@ -0,0 +1,132 @@
...
+claim, to identify the `type` and connection deails.
+If the type is either iSCSI or RBD, then it can take the connection details
+and transform them into the correct libvirtd representation.
+
+Example for an iSCSI volume:
+
+```xml
+  <disk type='network' device='disk'>
+    <driver name='qemu' type='raw'/>
+    <source protocol='iscsi' name='iqn.2013-07.com.example:iscsi-nopool/2'>
```
admiyo: Worth investigating:  It might not show up as iSCSI in the VM, even if it is iSCSI in the claim.  The underlying Host might end up attaching, which may not be what we want.

design-proposals/pvDiskStorage.md
fabiand: As described in the paragraph above - We are adding this `type` `PersistentVolumeClaim` and a property `claimName` which is used to name the specific name.

design-proposals/pvDiskStorage.md
fabiand: The trick here is, that libvirt will use qemu's facilities to connect to the iscsi target. Thus the host is not involved, and to the VM it will look like a regular disk.

And both things are intended: Exclude the host - this helps to avoid all kind of udev issues we have with containers - and second make it transparent to the vm how the storage is implemnted.

design-proposals/pvDiskStorage.md
```diff
@@ -0,0 +1,132 @@
...
+
+1. User adds the PersistentVolumeClaim as described above to the VM
+   instance.
+2. The VM Pod is getting scheduled on a host, the `virt-handler`
+   identifies the Claim, translates it into the corresponding
+   libvirt representation and includes it in the domain xml.
+
+Note: The `virt-controller` does not do anything with the claim.
+
+In this flow, the claim is not used by the pod, the claim is only used
```
rmohr: Is the kubelet still mounting it to the pod, or is it completely ignored by the pod? I could not 100% determine from the text.

design-proposals/pvDiskStorage.md
rmohr: Does this mean that libvirt does also not have anything to do on the host when we migrate a VM?

design-proposals/pvDiskStorage.md
fabiand: In this proposal the kubelet/pod is completely ignoring the PVC.
The PVC is exclusively used to provide the connection details to qemu.

I'll try to clarify this in the text.

design-proposals/pvDiskStorage.md
fabiand: At least for this proposal yes - The storage which qemu would consume here does not require any interaction with the host - it's process/qemu only.

fabiand: At least for this part yes, libvirt would not need to do anything on the host to attach this storage, as they are process/qemu flags.

rmohr:
LGTM

fabiand:
@sgott updated the PR to reflect the recent disucssion:

1. Fix API example
2. Use `disk/source/name` for claim name storage

design-proposals/direct-pv-disks.md
```diff
@@ -0,0 +1,136 @@
...
+this is used by the Vm in the following way:
+
+```yaml
+kind: VM
+spec:
+  domain:
+    devices:
+      disks:
+      - type: PersistentVolumeClaim
+        - source:
```
stu-gott: Is it "source" or "diskSource"?

fabiand: kubevirt/kubevirt#95 is changing it to use source instead of diskSource.

It looked more consistent and less redudant.

design-proposals/direct-pv-disks.md
```diff
@@ -0,0 +1,136 @@
...
+          name: disk-01
+        - diskTarget:
```
stu-gott: Given changes from https://github.com/kubevirt/kubevirt/pull/95 does that mean this should then be "target"?

fabiand: :D oh yes!

PR #7, opened 2017-04-12
DRAFT Add a high-level-plane.md proposal

fabiand:
Currently KubeVirt is exposing a pretty low-level interface to running VMs.

For consumers of KubeVirt this is not pretty convenient (i.e. because you want to run Win 10, and not set all the relevant attributes in the spec).
To achieve this, this PR is suggesting the introduction of a high-level plane, to abstract from the lower level.

### DIFF
--- a/design-proposals/high-level-plane.md
+++ b/design-proposals/high-level-plane.md
@@ -48,6 +48,8 @@ FIXME Is there a bidirectional conncetion between high- and low-level VM Spec?
 
 ## Implementation
 
+The implemantation will reside in a separate controller, which can be an addon to KubeVirt.
+
 TBD
 * Separate controller
 * Watches for high-level specs
--- a/design-proposals/high-level-plane.md
+++ b/design-proposals/high-level-plane.md
@@ -10,6 +10,8 @@ But this low-level API might be too low-level for i.e. UIs to integrate. On the 
 
 The introduction of a high level plane also tries to prevent that too much logic is pushed into the lower level mechanics.
 
+There will be users for a high-level plane, but we assume that the low-level plane is still relevant, that is the reason why a separate API is suggested, instead of changing the existing one.
+
 ### Use-case
 
 The primary focus is to provide a convenient high-level API suitable for integration into UIs.
@@ -47,6 +49,7 @@ FIXME Is there a bidirectional conncetion between high- and low-level VM Spec?
 ## Implementation
 
 TBD
-Separate controller
-Watches for high-level specs
-Links to low level specs
+* Separate controller
+* Watches for high-level specs
+* Links to low level specs
+
--- a/design-proposals/high-level-plane.md
+++ b/design-proposals/high-level-plane.md
@@ -1,0 +1,52 @@
+# High-level plane
+
+Author: Fabian Deutsch \<fdeutsch@redhat.com\>
+
+## Introduction
+
+Today KubeVirt is exposing a pretty low-level interface to virtualization capabilities.
+
+But this low-level API might be too low-level for i.e. UIs to integrate. On the higher level its typically required to set policies, the lower layer then provides the mechanics to implement this policy. The suggested high-level plane is the middle layer, translating policies into the mechanics.
+
+The introduction of a high level plane also tries to prevent that too much logic is pushed into the lower level mechanics.
+
+### Use-case
+
+The primary focus is to provide a convenient high-level API suitable for integration into UIs.
+
+A couple of aspects will be the focus of this initial part:
+
+* VM abstraction (support stopped VMs, OS flavors, disk abstraction, network abstraction, )
+* VM templating
+* …
+
+## API
+
+TBD
+The API is probably just a more high-level VM Spec for now, it would exist in parallel to the low-level representation.
+Example:
+
+kind: VirtualMachine
+spec:
+  state: down
+  os:
+    flavor: Windows 10
+  disks:
+    - name: windows-disk-pvc-from-kube
+      type: disk
+    - name: windpws-iso-pvc-from-kube
+      type: cdrom
+  nics:
+    - network: foonet-from-kube
+      model: virtio
+  display:
+    heads: 2
+
+FIXME Is there a bidirectional conncetion between high- and low-level VM Spec?
+
+## Implementation
+
+TBD
+Separate controller
+Watches for high-level specs
+Links to low level specs


design-proposals/high-level-plane.md
```diff
@@ -0,0 +1,57 @@
...
+TBD
+The API is probably just a more high-level VM Spec for now, it would exist in parallel to the low-level representation.
+Example:
+
+kind: VirtualMachine
+spec:
+  state: down
+  os:
+    flavor: Windows 10
+  disks:
```
mykaul: Are we missing here buses? Order on the bus? How do you represent multiple disks on different virtio-scsi (or just scsi) controllers? How do I know if the cdrom is on SCSI or IDE bus?
I'm not sure I understand the level of abstraction usefulness here.

fabiand: The right level has to be determined. But the idea of the example above was to use the os.flavor value to determine the right buses, and to have a policy of how devices are mapped to buses.

I hope to see input from @michalskrivanek.

fabiand: Actually - It does not need to be determined. IIUIC oVirt Engine has a good understanding of what level of abstraction they want to have.

michalskrivanek: "os" is tricky to start with;-) I suppose we can/should use libosinfo(https://libosinfo.org/) nomenclature 

high level vs low level seem to make sense. I don't expect the high level plane cares about buses or order indeed. at low level the libvirt xml seems the right one. It's just a matter of what has a priority, how exactly is each plane used. On one hand it is a good idea to carry both together, on the other...synchronization - as they both overlap significantly.

design-proposals/high-level-plane.md
```diff
@@ -0,0 +1,57 @@
...
+      type: cdrom
+  nics:
+    - network: foonet-from-kube
+      model: virtio
+  display:
+    heads: 2
+
+FIXME Is there a bidirectional conncetion between high- and low-level VM Spec?
+
+## Implementation
```
rmohr: I am not sure if you need  any controllers or watchers. The focus is also not completely clear for me. Maybe you need a converter instead? Fetch, convert, post/put. I don't see the need for a controller or watch there ...

design-proposals/high-level-plane.md
```diff
@@ -0,0 +1,57 @@
+# High-level plane
+
+Author: Fabian Deutsch \<fdeutsch@redhat.com\>
+
+## Introduction
+
+Today KubeVirt is exposing a pretty low-level interface to virtualization capabilities.
+
+But this low-level API might be too low-level for i.e. UIs to integrate. On the higher level its typically required to set policies, the lower layer then provides the mechanics to implement this policy. The suggested high-level plane is the middle layer, translating policies into the mechanics.
```
rmohr: I don't like this. Having a VM store sounds ok to me, which has it's own api. The whole high-level thing sounds pretty ominous to me (That does not mean that we can't have a different format on such an endpoint than the low-level VM-spec).

From what we have talked about I though this should just be a an add-on to store VMs while they are not used by the cluster?

design-proposals/high-level-plane.md
rmohr: Having a "plane" there with extra controllers and watchers for such a "store" sounds wrong to me.

design-proposals/high-level-plane.md
rmohr: I would not recommend anything like a controller or watcher included in this. Just make a plane store (which might have some convenience simplifications).

@michalskrivanek do you by any chance have some examples of such flavours? @fabiand  mentioned that you are currently impmenting them in oVirt.

design-proposals/high-level-plane.md
michalskrivanek: @rmohr: well, we have user-exposed "OS type". I suggest to use libosinfo "names" and use combined information from libosinfo and other backend configuration/code/logic to determine how the low level representation should look like.
In oVirt we use engine-side db, osinfo config(not libosinfo), and logic on vdsm to ultimately create the right xml.

design-proposals/high-level-plane.md
fabiand: @rmohr try: `dnf install -y libosinfo && osinfo-query os`.

design-proposals/high-level-plane.md
rmohr: To clarify this:  I don't like the term. Having a bunch of "add-ons", like kubernetes has helm or the dashboard sounds like a better description for me.

design-proposals/high-level-plane.md
```diff
@@ -0,0 +1,57 @@
...
+
+## API
+
+TBD
+The API is probably just a more high-level VM Spec for now, it would exist in parallel to the low-level representation.
+Example:
+
+kind: VirtualMachine
+spec:
+  state: down
```
rmohr: @fabiand could you clarify if this proposal is about having add-ons or, having this add-on?

I think we need to have a clear concrete use-case (nothing fancy just a clear sentence) on why we want to have this extra api.
If we don't do that, it is not clear to me what it is for and to distinguish what should be in an add-on and what should go into kubevirt/kubevirt.

Some use-cases which I could think off:

 1. I want to store my VMs somewhere so that I can start them later on
 2. I want to see shut-off VMs, which should be running
 3. I want  to see crashed VMs, which should be running
 4. I want to see running VMs
 5. I want to have flavours or templates to make it easier to re-use VM definitions
 6. I want to browse online and offline VMs in different UIs on different scales

Possible distinction between add-on and core feature:

 1. vm store (add-on)
 2. probably a core feature, since the VM should be running, kubevirt should try to correct that mistake which means events and status updates on the core API. Seeing a difference between the wanted and the actual state should always be possible.
 3. like 2
 4. core feature
 5. Could be an add-on or a core feature
 6. Here both APIs might not be enough, depending on the UI (mobile, full blown app, ...) we probably need different APIs reflecting different usage concepts.

fabiand:
After re-reading this PR after a while I notice that it covers at least three things:

1. Suggestion to add a high-level plane, consisting of several add-ons to the existing low-level "runtime", to provide additional more high-level functionality
2. Suggestion to add a more high-level VM abstraction - as an add-on
3. Suggestion of a way to support persistent VMs

The thoughts on 1. can be considered as some background thoughts, eventually not belonging here but rather being discussed and eventually added to the architecture description.

Regarding 2. - This is likely a separate proposal, to define how a high-level abstraction can look (I think close to what the example in this proposla lays out) - but also how it's implemented.
Such a high-level representation would be the entry point for higher-level tools, thus I'd expect those tools to use the object to perform operations. The question is now to which degree and how this is implemented. I do expect that the high-level VM spec leads to a low-level VM instance (as we currently support). The question is how the relationship is looking.
Eventually it makes sense to just use the high-level VM definition to instantiate a low-level instance, and keep a reference to this one around. The low-level instance could then be used for monitoring etc. The question is if hot-plugging is done through the high-level or low-level object.

The third point is - in this line of thoughts - closely related to 2. Because we could keep the high-level definition around, without a running instance. This would allow us to model persistent domains.

fabiand:
Closing in favor of kubevirt/kubevirt#272

PR #8, opened 2017-05-05
Defaulting

rmohr:
Introduce a defaulting service, to simplify VM spec managment when doing cluster wide managment based on libvirt managed VMs.

### DIFF
--- a/design-proposals/defaulting.md
+++ b/design-proposals/defaulting.md
@@ -1,0 +1,122 @@
+# Defaulting Service Proposal
+
+## Current situation
+
+Libvirt allows you to post very minimal Domain specifications. It will then
+fill out a lot of device details for you. On cluster level this can lead to
+difficulties, since depending on the fact if the VM has ever run before, you
+will have to operate on different specs, and you have to incoroporate VM
+specification changes once a VM was started the first time.
+
+The following illustrates how immense the defaulting is.
+
+First let's have a look on a fairly minimal KubeVirt VM Specification:
+
+```yaml
+apiVersion: kubevirt.io/v1alpha1
+kind: VM
+metadata:
+  name: testvm
+spec:
+  domain:
+    devices:
+      interfaces:
+      - type: network
+        source:
+          network: default
+    memory:
+      unit: MB
+      value: 64
+    os:
+      type:
+        os: hvm
+    type: qemu
+```
+
+
+This can be translated to a quite small Domain XML:
+
+```xml
+<domain type="qemu" xmlns:qemu='http://libvirt.org/schemas/domain/qemu/1.0'>
+  <name>testvm</name>
+  <memory unit="MB">64</memory>
+  <os>
+    <type>hvm</type>
+  </os>
+  <devices>
+      <interface type="network">
+      <source network="default"></source>
+    </interface>
+   </devices>
+</domain>
+```
+
+After we have defined the domain in libvirt, the XML looks like this:
+
+```xml
+<domain type='qemu' xmlns:qemu='http://libvirt.org/schemas/domain/qemu/1.0'>
+  <name>testvm</name>
+  <uuid>a8335812-53d4-4e57-ac0b-72951c885a57</uuid>
+  <memory unit='KiB'>62500</memory>
+  <currentMemory unit='KiB'>62500</currentMemory>
+  <vcpu placement='static'>1</vcpu>
+  <os>
+    <type arch='x86_64' machine='pc-i440fx-2.7'>hvm</type>
+    <boot dev='hd'/>
+  </os>
+  <clock offset='utc'/>
+  <on_poweroff>destroy</on_poweroff>
+  <on_reboot>restart</on_reboot>
+  <on_crash>destroy</on_crash>
+  <devices>
+    <emulator>/usr/bin/qemu-system-x86_64</emulator>
+    <controller type='usb' index='0' model='piix3-uhci'>
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x01' function='0x2'/>
+    </controller>
+    <controller type='pci' index='0' model='pci-root'/>
+    <interface type='network'>
+      <mac address='52:54:00:2e:84:30'/>
+      <source network='default'/>
+      <model type='rtl8139'/>
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x02' function='0x0'/>
+    </interface>
+    <input type='mouse' bus='ps2'/>
+    <input type='keyboard' bus='ps2'/>
+    <memballoon model='virtio'>
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x03' function='0x0'/>
+    </memballoon>
+  </devices>
+</domain>
+```
+
+Finally, when a Domain was successfully started by Libvirt, Libvirt assigns an `alias` to every device in the XML:
+
+## The Problem
+
+ * Hotplug operations are not possible with the minimal VM spec of kubevirt, since we could match wrong devices.
+ * After VM restarts, different devices or addresses could be assigned to a VM by libvirt
+ * Libvirt assigns all that data on the host, either when defining the Domain, or after the Domain was started.
+   That requires asynchronous lookups, and when thinking about administrating offline VMs, it binds your domain specification management to the kubevirt runtime. happen, to get that data
+
+If your VMs are ephemeral (no migrations, recreating VMs vs. restarting VMs, no hotplug), you are not affected by that.
+
+## Proposed Solution
+
+### Defaulting service based on libvirt
+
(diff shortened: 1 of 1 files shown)


design-proposals/defaulting.md
```diff
@@ -0,0 +1,122 @@
...
+Finally, when a Domain was successfully started by Libvirt, Libvirt assigns an `alias` to every device in the XML:
+
+## The Problem
+
+ * Hotplug operations are not possible with the minimal VM spec of kubevirt, since we could match wrong devices.
+ * After VM restarts, different devices or addresses could be assigned to a VM by libvirt
+ * Libvirt assigns all that data on the host, either when defining the Domain, or after the Domain was started.
+   That requires asynchronous lookups, and when thinking about administrating offline VMs, it binds your domain specification management to the kubevirt runtime. happen, to get that data
+
+If your VMs are ephemeral (no migrations, recreating VMs vs. restarting VMs, no hotplug), you are not affected by that.
```
rmohr: In such cases it might even be annoying that the spec is so drastically changed by a defaulting. We could add a header which allows disabling this defaulting ...

design-proposals/defaulting.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Proposed Solution
+
+### Defaulting service based on libvirt
+
+Except for some runtime information, which is normally not included in migratable domain XMLs, libvirt fills in all defaults when defining a domain. We can leverage that with a simple REST based service ,which defines VMs in an isolated libvirt (running libvirt and the small REST service in a Pod). The service is a standalone component in KubeVirt and can then be integrated into two different flows:
+
+ 1. When doing a POST to `/apis/kubevirt.io/v1alpha1/namespaces/mynamespace/vms`, `virt-api` can do a roundtrip to that service and prefill the defaults. As the response of the POST to `virt-api`, you will get the completely prefilled VM specification. This updated spec contains all the necessary bits, to allow consistent hotplugging and keeping the VM definition consistent between redefines.
```
rmohr: Because of the ephemeral VM use-case, we might consider not to do this at all ...

berrange: Note that the way the defaults a filled in may vary depending on QEMU version and libvirt version.  So if you fill in defaults at the time the config is loaded into the API, your config will be fixed against that particular QEMU/libvirt version. This is undesirable in general - particularly if you have a cluster with hosts running at different libvirt/QEMU versions - eg if virt-api happened to fill in defaults based on QEMU version N, and you then try to launch on QEMU version N-1, it may well fail.

stu-gott: Do you have a concrete example of this?

design-proposals/defaulting.md
```diff
@@ -0,0 +1,122 @@
...
+ 2. Add an extra endpoint to `virt-api` which only purpose is to prefill the VM spec. The URI will be `/apis/kubevirt.io/v1.alpha1/vms/defaults` (non-namespaced). You can POST VM specifications there and get a fully populated VM spec back. With this service URL in place, mixing administrative tasks like hotplug, adding devices for later runs and changing the spec for later runs (e.g. memory) can be solved in a consistent way, without the need to ever do an asynchronous roundtrip to libvirt to build consistent specs. **This allows decoupling administrative tasks completely from the runtime**.
```
stu-gott: I suspect the URL should have had a version of "v1alpha1" without a dot.

Trailing whitespace.

design-proposals/defaulting.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+Finally, when a Domain was successfully started by Libvirt, Libvirt assigns an `alias` to every device in the XML:
+
+## The Problem
+
+ * Hotplug operations are not possible with the minimal VM spec of kubevirt, since we could match wrong devices.
+ * After VM restarts, different devices or addresses could be assigned to a VM by libvirt
+ * Libvirt assigns all that data on the host, either when defining the Domain, or after the Domain was started.
+   That requires asynchronous lookups, and when thinking about administrating offline VMs, it binds your domain specification management to the kubevirt runtime. happen, to get that data
```
stu-gott: This line seems to trail off.

design-proposals/defaulting.md
rmohr: @berrange, good point. I agree as far as such a service would impose a specific upgrade order, or would require scheduling prevention, so that it does not end up on a too old libvirt/QEMU host.

AFAIK oVirt does at the end pretty much the same thing like this defaulting service, by collecting the detailed device information from a started VM and stores it for later restarts of the VM. @michalskrivanek, correct? Is the described problem something you experienced?

berrange: I don't think that really solves it in a satisfactory manner. If you load your config into kubevirt, and it sets all the defaults based on the current QEMU. Then you upgrade QEMU + libvirt. All the configs are still going to be using the defaults associatd with the old QEMU + libvirt, and so likely miss out on improvements associated with the new QEMU+libvirt until you delete the config and load a new one. We should really only ever apply defaults at the time you start a guest, and those defaults should not be saved back into the master config stored in K8s API, but to a separate area, so if you stop & start the VM again later, it picks up the newer defaults once more.

michalskrivanek: The "built-in" libvirt functionality to fill in defaults is really only used for things we do not care about/ do not expose at oVirt level (PCI addresses, bridges). For everything else it's useless as the "default" VM is defined by other means (Inst types, Blank template, user choice of OS).
I wonder if there is any use for this defaulting for anything other than "low-level details".

User-wise you do want to separate that as the "high level" is defining how the VM should look like in terms of what users care about (graphics, 4 CPUs, USB) from what they do not care about but you need that, persisted (additional "mandatory" hw, stable pci addresses). Those are best done ad-hoc when you first run the VM(basically that's the case in oVirt)
Question is if it should be mixed all together...

rmohr: @berrange I agree that in many cases you exactly want the behaviour you described. I would also actually favour making this defaulting opt-in. 

But the oVirt VMs are the perfect example that there is the need for keeping the devices like they were generated the first time. Basically the VM also sees adresses, ... and changes to that settings can be disruptive for the OS in the VM.
Maybe we can narrow down the scope of the defaulter a little bit, and we could find parts which are defaulted but not visible from inside the VM. Then we can leave them out.

Also, to avoid misunderstandings, the described POST URL in 1 is only about VMs in the cluster which should run. This is not the place to store VMs for later runs (you can think of it a little bit like transient VMs in libvirt, except that they do not immediately disappear if the crash). So I completely agree to

> those defaults should not be saved back into the master config stored in K8s API, but to a separate area, so if you stop & start the VM again later, it picks up the newer defaults once more.

If you want to make use of these default for later runs is completely optional. The goal I had in mind is to allow VM stores on top of this KubeVirt runtime environment, to choose which level of consistency they want to provide to the outside world, without having to asynchronously collect defaults assigned by the lowest layers in the stack.

berrange: Yes, I understand the "vm" object currently only represents running VMs, but I'm anticipating that is going to need to change in future to represent the VM config in k8s, dependently of whether its currently running or not.

rmohr: @michalskrivanek I agree, this is only there for low-level details like (additional "mandatory" hw, stable pci addresses).
Other things like Instance types, Blank template, user choice of OS, could be handled before. I think that the defaulting endpoint in 2 could be a good supplement for them, so that you don't have to mix these two cases, where not necessary.

A possible higher level hotplug flow could look like this:

1) A user defines a VM in an arbitrary store
2) this store, populates the spec (internally, a copy of the posted definition) with the defulting service and keeps it in its' store
3) The fully populated spec is used for starting the VM
4) I hotplug 2 devices which should not be there again on the next VM start
5) I hotplug 1 device which should be kept on the next run.

To guarantee the same pci addresses to be assigned on the next run, we need to keep in mind, that two devices are not there anymore, and that the third device has to keep it's higher address from the last run. The defaulting endpoint allows you to easily calculate everything, without you ever having to do the actual hotplug on libvirt.

rmohr: @berrange well I am trying hard to make sure that this does not happen (this proposal is actually part of that attempt). Basically to allow a clear separation between administration and runtime. You can see if a VM is running or not. If it is not there or in Succeeded of Failed state when querying `/apis/kubevirt.io/v1alpha1/namespaces/mynamespace/vms`, but it should purely be about runtime information.

Now about VM administration in general:, I see that as a complete separate thing, with requirements, not necessarily beneficial for a runtime which should scale well and should be stable. I think both areas benefit most, if we can untangle them, and so far this defaulting, happening in libvirt, is the most interesting place, preventing me from doing this. This does not mean that the store could not be part of the k8s API, they should just not be in the same etcd directory (`/apis/kubevirt.io/v1alpha1/namespaces/mynamespace/vms`). It is definitely a valid option to have a VM store as another third party resource (but why not having them in e.g. a git repo instead, like it is popular for Kubernetes deployments).

What do you think?

rmohr:
Superseeded by https://github.com/kubevirt/kubevirt/pull/281

PR #21, opened 2019-03-19
Add initial info around KubeCon EU 2019

codificat:
Starting to plan a KubeVirt gathering during the upcoming KubeCon EU in Barcelona in May


codificat:
Forgot to merge that and Fabian beat me with b4d2b0c  so closing

PR #26, opened 2019-06-28
Sync owners with kubevirt.github.io

iranzo:


OWNERS
```diff
@@ -6,9 +6,11 @@ reviewers:
   - tripledes
   - joeldavis84
   - DirectedSoul1
```
jparrill: I think this is not necessary anymore

PR #29, opened 2019-07-01
Remove non logo file

iranzo:


iranzo:
Copied .gitignore from kubevirt.io

PR #44, opened 2020-03-13
Add yaspeller checks via pre-commit

iranzo:
Yaspeller got a PR merged so that it can be used directly.

This moves configuration to .yaspellerrc and makes pre-commit to check only markdown files in the commit


ptrnull:
Checked PR, looks fine, new changes in our code produced by a PR merged allowed us to use the source library directly.
/lgtm
/approve

PR #51, opened 2021-01-18
Propose a session about running VMs in Istio service mesh

rhrazdil:
Signed-off-by: Radim Hrazdil <rhrazdil@redhat.com>

### DIFF
--- a/events/2021-kubevirt-summit/kubevirt-with-istio.md
+++ b/events/2021-kubevirt-summit/kubevirt-with-istio.md
@@ -1,0 +1,30 @@
+# Title
+
+Kubevirt with Istio service mesh
+
+# Abstract
+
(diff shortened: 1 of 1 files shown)


codificat:
Many thanks for your proposal. This will be reviewed during this week.

codificat:

davidvossel (commented):
Due to having received more content than we expected, we had to quickly form a review team (thanks Red Hat and NVIDIA) to make difficult decisions about what presentations to accept.

**This submission is currently on a wait list,** which means that it’s possible that it will be accepted if an opening occurs as the schedule is being formed.

As a result, we’d like to keep this PR open in the event that someone backs out. Thank you for your submission and support of the KubeVirt community!

PR #77, opened 2021-01-28
move proposal Automating KubeVirt with Tekton Pipelines to the right directory

suomiy:
I am sorry, I put the proposal to the wrong directory @codificat @davidvossel


codificat:
ah, thanks for the correction!
/lgtm
/approve

PR #78, opened 2021-02-02
Add information about the KubeVirt v1 session

codificat:
This session is one of the "general interest" sessions planned for the Summit.

As the session program points to the proposal for details, this PR adds a description of that session so it can be linked the same way as the others.


codificat:
@davidvossel FYI.

hroyrh:
Looks good.
/lgtm
/approve

PR #105, opened 2021-04-29
Add check whether all org repos are present

dhiller:
Check for every repo in the org that its OWNERS file is present in this
file as a subprojects owners reference.

Signed-off-by: Daniel Hiller <dhiller@redhat.com>


mazzystr:
@dhiller  Are we ready for tag and merge?

dhiller:
> @dhiller Are we ready for tag and merge?

/remove-lifecycle stale

@mazzystr good question. TBH I totally forgot about this one. Do you think this has value? If yes, then I'll look at it again.

PR #125, opened 2021-08-27
Dedicated migration network proposal

jean-edouard:


design-proposals/dedicated-migration-network.md
```diff
@@ -0,0 +1,94 @@
...
+This problem is currently mitigated by throttling the number of concurrent migrations and using long timeouts.
+
+## Goals
+- Allow a cluster admin to setup and dedicate a secondary physical cluster network to migrations.
+- Keep the main network as the default medium for migrations.
+
+## Non Goals
+- Network configuration. A functional secondary physical network should be provided.
+- Network definition. A valid NetworkAttachmentDefinition should be written by the cluster admin and provided to the KubeVirt CR (Custom Resource).
+- Cluster components. A multi-network CNI like multus should be properly installed, as well as necessary plugins, and any required daemon (such as CNI DHCP) should be running on each node.
```
EdDev, line 18: The mentioning of `CNI DHCP` can now be dropped.
I think you can just mention "daemonsets" in general, avoiding details.

design-proposals/dedicated-migration-network.md
```diff
@@ -0,0 +1,94 @@
...
+In order for VMI migrations to succeed, there must be a fair amount of bandwidth available.
+If for example the memory of a migrating VMI changes faster than the current speed of the main network, the migration will never succeed.
+This is especially problematic during mass eviction events like platform upgrades.
+This problem is currently mitigated by throttling the number of concurrent migrations and using long timeouts.
+
+## Goals
+- Allow a cluster admin to setup and dedicate a secondary physical cluster network to migrations.
+- Keep the main network as the default medium for migrations.
+
+## Non Goals
```
EdDev, line 15: Per my understanding, this proposal does not cover QoS and the new network prioritization over the other networks.
I mention this, because this point is sometime essential for prioritizing the migration traffic over others.

Can we mention this as well in the non-goals?

design-proposals/dedicated-migration-network.md
```diff
@@ -0,0 +1,94 @@
...
+As an administrator of the cluster I want all my VMIs to be migrated over a dedicated network to provide a stable connection with a larger bandwidth comparing to the default k8s pod network
+
+## Definition of Users
+This feature is intended for cluster admins who wish to accelerate and/or control migration network usage.
+
+## Repos
+- kubevirt/kubevirt:
+  - add a CR option and handle it at the virt-handler level
+- kubevirt/kubevirtci:
+  - install whereabouts on CNAO lanes
```
EdDev, line 30: Can you add links here?

design-proposals/dedicated-migration-network.md
```diff
@@ -0,0 +1,94 @@
...
+
+# Design
+If a dedicated migration network is defined in the KubeVirt CR, migration target virt-handler should communicate their IP of that network instead of the main one.
+The source virt-handler will then use that IP to communicate with the target, which will get routed through the secondary network.
+
+## Cluster setup example
+- Every node has an eth1 connected to the secondary network that will be used for migrations
+- A DHCP server is running on that network
```
EdDev, line 38: This is the case with `whereabouts`?

Can you also add a section that describes the components that play a role here? Including `whereabouts`?
I guess the example needs to be updated with its usage, as I see below the mentioning of `CNI DHCP`.

design-proposals/dedicated-migration-network.md
```diff
@@ -0,0 +1,94 @@
...
+      "master": "eth1",
+      "mode": "bridge",
+      "ipam": {
+           "type": "dhcp"
+      }
+    }'
+```
+
+## API Examples
+In the KubeVirt CR, a field named dedicatedMigrationNetwork will be added under spec.configuration.migrations, like so:
```
EdDev, line 70: Can we see this field defined before the example? It will be a bit more formal then.

Also, perhaps consider dropping the "dedicated" prefix, it seems redundant.

design-proposals/dedicated-migration-network.md
```diff
@@ -0,0 +1,94 @@
...
+```
+spec:
+  configuration:
+    developerConfiguration:
+      featureGates:
+      - LiveMigration
+    migrations:
+      dedicatedMigrationNetwork: my-secondary-network
```
EdDev, line 78: Does it make sense to have it listed under a "networks" subtree?
This way we can extend it to other types of dedicated networks (e.g. storage).

design-proposals/dedicated-migration-network.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Repos
+- kubevirt/kubevirt:
+  - add a CR option and handle it at the virt-handler level
+- kubevirt/kubevirtci:
+  - install whereabouts on CNAO lanes
+
+# Design
+If a dedicated migration network is defined in the KubeVirt CR, migration target virt-handler should communicate their IP of that network instead of the main one.
+The source virt-handler will then use that IP to communicate with the target, which will get routed through the secondary network.
```
EdDev, line 34: I have some questions which will be nice to see posted in this section:
- How (from a high level) is the IP address supposed to be resolved?
- How will virt-handler know which interface is the dedicated one of the migration network? Based on the ID multus will assign it? Is this safe for future additional networks added?
- The network will use a single flat subnet, right?
- What will cause the traffic to be routed through the secondary interface? The directly connected subnet? A special route?
- Are MAC addresses managed in this solution (compared to VMI interfaces, where we do have an option to manage it)?

design-proposals/dedicated-migration-network.md
```diff
@@ -0,0 +1,94 @@
...
+make cluster-up
+```
+The functional test suite will first:
+- Create a NetworkAttachmentDefinition bridged to eth1, like the above, except using the `whereabouts` CNI plugin to avoid requiring a DHCP server.
+- Alter the KubeVirt CR to set that as the dedicated migration network
+- Wait for virt-handlers to respawn
+Then the functional tests will verify that migrations go through that network.
+
+# Future improvements
+- Multiple migration networks could be supported, with more granular settings for which VMIs migrate over which network
```
EdDev, line 94: Assuming this may indeed come as a future enhancement. Is the current suggested API supportive of this?

design-proposals/dedicated-migration-network.md
```diff
@@ -0,0 +1,94 @@
+# Overview
```
EdDev, line 1: It may be useful to mention for the less knowledgeable reader that the migration is conducted at the IP level between the virt-handlers and not directly from the virt-launchers. It is not that trivial to understand this part and readers may wrongly assume the secondary network is on the virt-launcher and not the virt-handler.

EdDev (commented):
I mainly tried to ask questions in order to add some more info and context into the proposal.

Thank you!

design-proposals/dedicated-migration-network.md
```diff
@@ -0,0 +1,129 @@
...
+In the KubeVirt CR, the ability to define network roles will be added under spec.configuration.network, like so:
+```
+spec:
+  configuration:
+    developerConfiguration:
+      featureGates:
+      - LiveMigration
+    network:
+      roles:
+      - networkAttachmentDefinition: my-secondary-network
```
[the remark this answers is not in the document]
EdDev: > In this case the evolution for this possible additional use-case would be possible without disturbance, but we can still focus on the main use-case. Further the API field naming gives very good hints regarding to what these fields may do.
> 
> If this section is then also placed in the migration section, also a more crisp representation may be possible without loosing meaning:

I understand the advantages here, this is a good argument.
Nevertheless, I think that there are also advantages in centralizing networking aspects under the same "roof".
I will try to outline my view on this with some scenarios.

---

Summarizing the advantages of the current proposal with a use-case beyond "migration":
- Use case (optional extensions):
  - Terminal (view) network: Currently, accessing the terminal (console/vnc) is done through the API server. A potential dedicated network for this task may make sense. The reason is very similar to the migration network, one wants  a dedicated network to work with in order to assure continuation of the service even if others are overloaded or some services are down.
  - It may be useful in the future to let the Kubevirt cluster operator (and not the k8s operator) to limit which networks can be used by the VM/s. One optional way to do this may be to list the "approved" networks. I'm just playing with the idea, to illustrate how this can evolve beyond the single need we have in front of us today.

- When defining a secondary network, there is some networking related knowledge needed from the operator. Operators have sometime domain expertise, like we do in our SIG separation. Today we already have potential network specific parameters that are required, the minimum being the NAD and from there the pod interface name. It is likely that we will need other details to control this interface (through CNI pod annotations), therefore more fields may be needed.
All this sums up with the obvious need to understand and know what a dedicated network means and what are its values. It is challenging IMO for one operator to know it all, and such a domain separation and centralization makes it more maintainable and extendable.

- Continuing the previous point, centralizing network services is a common organization in API/s, at least for other solutions that exist on the market. I guess the main point would sum up to be: How subjects and features should be organized. The approach I favor is per service domain and the other one is feature centric.
There is one nice advantage to all such systems, they cause a generic format for providing the service to whatever feature/sub-system that requires it.

- It is preferable not to need and change the API such that it will cause backward compatibility issues. We saw bad examples of this in k8s when IPv6 was not though of from day one (duplication and backward compatibility with services).

I find the current proposal better structured due to its centralization. Having network parts all over the configuration spec looks messy to me. I do have hard time imagining that this will be the only network option we will add, therefore my opinion on this.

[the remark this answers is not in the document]
rmohr: > * Terminal (view) network: Currently, accessing the terminal (console/vnc) is done through the API server. A potential dedicated network for this task may make sense. The reason is very similar to the migration network, one wants  a dedicated network to work with in order to assure continuation of the service even if others are overloaded or some services are down.

There are no plans to do that. VNC and console are administrative minimal access points. We btw. also have `port-forward` and `ssh` now. KubeVirt is not a out-of-the-box solution for e.g. remote desktop.

General Traffic shaping for workload ports, no matter if VMs or Pods is in general nice, but I personally don't see this happening in the KubeVirt CR.

> * It may be useful in the future to let the Kubevirt cluster operator (and not the k8s operator) to limit which networks can be used by the VM/s. One optional way to do this may be to list the "approved" networks. I'm just playing with the idea, to illustrate how this can evolve beyond the single need we have in front of us today.

This is something outside of the scope of kubevirt. What people are allowed to create is a matter of RBAC and access control where there is no need to separate between pods and VMs. Sending traffic over a network with VMs should not be more harmful to the cluster than sending it over pods.

Even if so, like above, I see this in the context of general purpose workload traffic shaping which I don't see happening in the KubeVirt CR.

> * When defining a secondary network, there is some networking related knowledge needed from the operator. Operators have sometime domain expertise, like we do in our SIG separation. Today we already have potential network specific parameters that are required, the minimum being the NAD and from there the pod interface name. It is likely that we will need other details to control this interface (through CNI pod annotations), therefore more fields may be needed.

I am not aware of those. Can you please elaborate? We need all kinds of realistic examples which we can get.

>   All this sums up with the obvious need to understand and know what a dedicated network means and what are its values. It is challenging IMO for one operator to know it all, and such a domain separation and centralization makes it more maintainable and extendable.
> * Continuing the previous point, centralizing network services is a common organization in API/s, at least for other solutions that exist on the market. I guess the main point would sum up to be: How subjects and features should be organized. The approach I favor is per service domain and the other one is feature centric.

If we can't create a critical mass of examples which will happen in kubevirt (and we are already a few years in development now and did not see this need emerge), I would continue to favour the feature centric approach and worst-case maintain the deprecation of a simple field which is entirely under our control.


>   There is one nice advantage to all such systems, they cause a generic format for providing the service to whatever feature/sub-system that requires it.

It would only be generic if we can expect that all feature related knobs would not be in that list. That would be unrealistic. For migrations we have a great example where it would actually be split the feature settings.

> * It is preferable not to need and change the API such that it will cause backward compatibility issues. We saw bad examples of this in k8s when IPv6 was not though of from day one (duplication and backward compatibility with services).

Yes this is however only a minimal issue if we would have to deprecate the field. It is pretty much irrelevant. I think you have to take comparable examples. All the code is under our control which relies on this. So it is safe to wait for use-cases to emerge, which they potentially will not, or only after a very long time period.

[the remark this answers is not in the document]
rmohr: > > * When defining a secondary network, there is some networking related knowledge needed from the operator. Operators have sometime domain expertise, like we do in our SIG separation. Today we already have potential network specific parameters that are required, the minimum being the NAD and from there the pod interface name. It is likely that we will need other details to control this interface (through CNI pod annotations), therefore more fields may be needed.
> 
> I am not aware of those. Can you please elaborate? We need all kinds of realistic examples which we can get.

Here for instance, if it turns out that we need such a section for workload ports to properly integrate within existing networks (e.g. to function at all), independent of traffic shaping, I could imagine for that feature (workload networks) an extra section where a list of additional config hints makes sense.  We can again ask us here if we want then to combine workload network (vm traffic) with administrative networks (e.g. migration network).

rmohr:
@jean-edouard what do you think about proceeding with just setting the network name for now like discussed originally? I think it is ok to deprecate the field if this PR can come up with enough use-cases. It then has the time to ripe while we can proceed with the feature.

jean-edouard:
> @jean-edouard what do you think about proceeding with just setting the network name for now like discussed originally? I think it is ok to deprecate the field if this PR can come up with enough use-cases. It then has the time to ripe while we can proceed with the feature.

Sounds good to me!

So in the CR:
```
spec:
  configuration:
    migrations:
      network: <NAD name>
```

I can easily make the change in this proposal and in the code today.

stu-gott:
@rmohr @jean-edouard @EdDev Are there still any points that still have some friction to them? What can be done to make forward progress on this?

stu-gott:
Ping? We've put a lot of effort into this so far. Is there a way to keep the positive momentum? @EdDev? @rmohr?

EdDev:
> Ping? We've put a lot of effort into this so far. Is there a way to keep the positive momentum? @EdDev? @rmohr?

@stu-gott , I have shared my opinion on this some time ago. @rmohr prefers a different path and I failed to convince him otherwise. Except the exposed API, all other points LGTM.
As I see it, since the proposal follows @rmohr request, it just needs his approval. I have not requested changes on the PR in order not to block it.

If there is anything specific I can help with beyond this, please let me know.

rmohr:
> @stu-gott , I have shared my opinion on this some time ago. @rmohr prefers a different path and I failed to convince him otherwise. Except the exposed API, all other points LGTM. As I see it, since the proposal follows @rmohr request, it just needs his approval. I have not requested changes on the PR in order not to block it.
> 
> If there is anything specific I can help with beyond this, please let me know.

If @EdDev has interest in continuing this proposal regarding to a dedicated API, I have no issues with that, if use-cases can be identified. He mentioned one case I am potentially not aware of.

I think that we don't need a proposal for this at all. If no one is interested in following up on what @EdDev had in mind we can just close this proposal and continue with the PR.

/close

rmohr:
@jean-edouard if you want to see this also merged as a proposal just reopen.

PR #134, opened 2021-09-16
nic-hotplug: provide design proposal document

maiqueb:
Signed-off-by: Miguel Duarte Barroso <mdbarroso@redhat.com>


design-proposals/nic-hotplug.md
```diff
@@ -0,0 +1,134 @@
...
+This would help bridge the gap between KubeVirt's feature set and what the user
+base expects from it.
+
+## Goals
+Adding network interfaces to running VMs.
+Removing networking interfaces from running VMs.
+A VM can have multiple interfaces connected to the same (secondary) network(s).
+
+## Non Goals
+Providing multiple connections to the primary cluster network (managed by
```
phoracek, line 18: It also would not cover SR-IOV hotplug, right? Since that depends not only on CNI but also on DP. To allow that, we would need to pick approach similar to storage hotplug.

design-proposals/nic-hotplug.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Definition of Users
+This feature is intended for cluster users who want to connect their existing
+Virtual Machines to additional existing networks.
+These users may - or may not - have permissions to create the aforementioned
+additional networks; if so, they would need to rely on the cluster
+administrator to provision it for them.
+
+## User Stories
+* as a user, I want to add a new network interface to a running VM.
```
phoracek, line 29: user story should contain the motivation / "because", could you explain why is it useful and why users can't list all the needed networks while defining the VM?

design-proposals/nic-hotplug.md
```diff
@@ -0,0 +1,134 @@
...
+to call the ADD / DEL for **all** networks listed in its
+`k8s.v1.cni.cncf.io/networks` annotation.
+
+As such, the first part of the design implementation must focus on multus; it
+must be refactored to work as a thick plugin, which enables it to be triggered
+not only when the pod's sandbox is created, but also on-demand.
+
+### Thickening multus
+A “thin CNI plugin” runs as a one-shot process, typically as a binary on disk
+that runsrun on a Kubernetes host machine.
```
phoracek, line 52: runs?

design-proposals/nic-hotplug.md
maiqueb: Doesn't that defeat the purpose of the motivation section ?

phoracek: You only argue there that it is an industry standard. I think it is important to rephrase why is hotplug needed / what is it used for, so reviewer of this proposal can decide whether the feature is something that makes sense in the scope of Kubernetes. But I guess you could mention it in Motivation instead of here

dougbtv: Possibly "because my workload cannot tolerate the VM rebooting in order to attach networks" or "because my legacy workload requires the dynamic attachment of L2 networks" or "because my workload statefully tracks the attachment of networks" are some that I could think of.

design-proposals/nic-hotplug/nic-hotplug.md
```diff
@@ -0,0 +1,142 @@
...
+
+The CNI shim will then be invoked by kubelet, and send the CNI ADD/DELETE
+commands to the server side of multus via the unix domain socket previously
+mentioned.
+
+Refer to the sequence diagram below for more information:
+
+![Create pod flow for thick plugin](create-pod-flow.png)
+
+### Further multus requirements
```
dougbtv, line 74: One requirement is that there's still a "thin plugin mode" that's available. Tomo noted that this is likely important to edge computing environments where there's limited memory/cpu, and the thick plugin implementation is likely to utilize a greater amount of resources (how much more I'm not sure, part of me thinks it won't be a lot more?)

One possibility here is that we create a `./v3` folder with the current implementation.

I believe we'll likely tag a `v4.0` for the change for thick plugin.

design-proposals/nic-hotplug/nic-hotplug.md
```diff
@@ -0,0 +1,142 @@
...
+Multus must also be updated to act only on specific networks (the ones being
+added or removed) rather than the full annotation list.
+
+Another conflicting behavior of multus is its interface naming scheme; it
+iterates all networks in the `k8s.v1.cni.cncf.io/networks` annotation, and for
+each creates an interface in the pod named `netX`, where `X=<net index> + 1`.
+The simplest way to work-around this limitation is to explicitly define the
+interface names being plugged into the pod, as documented in the
+[multus documentation](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/how-to-use.md#launch-pod-with-text-annotation-with-interface-name).
```
dougbtv, line 83: I think I'll add an item to the NPWG this week about the possibility of having a dynamically generated, and potentially predictable interface name, possibly as an option in Multus. Maybe it's possible that we hash a few values. I'll ask Kaloom what's done for kactus.

design-proposals/nic-hotplug/nic-hotplug.md
```diff
@@ -0,0 +1,142 @@
...
+It will listen to pod updates, and react when the `k8s.v1.cni.cncf.io/networks`
+have changed. It will then issue ADD operations for each new network, and DELETE
+operations for networks that were removed, to the responsible routine, within
+the same process.
+
+Refer to the sequence diagram below for more information:
+
+![Kubernetes native solution](controller-based-approach.png)
+
+This alternative seems to be better aligned to Kubernetes patterns, requires
```
dougbtv, line 103: +1, re: using controller pattern

design-proposals/nic-hotplug/nic-hotplug.md
maiqueb: Nice.

If we can create an issue out of that discussion, I'll reference it here.

design-proposals/nic-hotplug/nic-hotplug.md
maiqueb: I wouldn't like to engage in these talks without having characterized the system.

Having said that, I do think (intuitively) that it will require more memory / CPU than the thin alternative.

I'l add that requirement. Thanks.

design-proposals/nic-hotplug.md
maiqueb: Correct. It would deserve a dedicated design document. 

As long as we ensure this design does not make it impossible in the future for sriov, we are OK.

design-proposals/nic-hotplug.md
maiqueb: Thanks @dougbtv . 

I've added a few lines going over a subset of those. @phoracek is it clearer now ?

design-proposals/nic-hotplug.md
maiqueb: Done.

design-proposals/nic-hotplug.md
phoracek: Yeah, it looks great, thanks

design-proposals/nic-hotplug/nic-hotplug.md
```diff
@@ -0,0 +1,234 @@
...
+                ]
+            }
+        }'
+---
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: macvlan-conf-2
+  namespace: testns1
+spec:
```
phoracek, line 158:
```suggestion
  name: macvlan-conf-2
spec:
```

design-proposals/nic-hotplug/nic-hotplug.md
```diff
@@ -0,0 +1,234 @@
...
+
+The CNI shim will then be invoked by kubelet, and send the CNI ADD/DELETE
+commands to the server side of multus via the unix domain socket previously
+mentioned.
+
+Refer to the sequence diagram below for more information:
+
+![Create pod flow for thick plugin](create-pod-flow.png)
+
+### Further multus requirements
```
s1061123, line 82: From current multus user's point of view, we should keep current multus binary and add thick plugin in another one. As Doug mention, if we make thick plugin in same binary (with two 'mode') then current multus binary gets bigger than current. Multus is also used in edge case which requires small footprint and its user wants to use current thin plugin and they wants to keep current binary size.

So could you please add following into the requirements?

- keep current binary as thin plugin and thick plugin will be another binary
- try to keep current binary size for thin plugin (no new feature for thin plugin, well we may need a few code changes)
- multus provides both daemonset (thick plugin and thin plugin)

design-proposals/nic-hotplug/nic-hotplug.md
maiqueb: Good catch. I'll add this manually.

s1061123 (changes requested):

design-proposals/nic-hotplug/nic-hotplug.md
maiqueb: OK, I get it - thanks. Makes sense.

IIUC, we'd have to ship both executables on the docker images, and have two types of multus daemonset specs, each with different entrypoint commands. Right ?

s1061123: Right, two daemonset (multus-daemonset.yml and multus-daemonset-thick.yml) and one container image which contains both binary.

design-proposals/nic-hotplug.md
maiqueb: Likewise, macvtap interfaces - which are not usable on Openshift - are not in scope. 

I'm listing those as well, since this is an upstream doc.

design-proposals/nic-hotplug/nic-hotplug.md
```diff
@@ -0,0 +1,234 @@
...
+
+To transform multus into a thick plugin, it is needed to instantiate a long
+lived process - which will be the multus pod entrypoint - listening to a unix
+domain socket - this socket must be available both in the multus pod and the
+hosts's mount namespaces; as such, a bind mount to host this socket must be
+provided for the multus pod.
+
+The CNI shim will then be invoked by kubelet, and send the CNI ADD/DELETE
+commands to the server side of multus via the unix domain socket previously
+mentioned.
```
pperiyasamy, line 76: @maiqueb can you also mention about delegate CNI/IPAM volume mounts (config, storage, log files etc.) to be available inside multus pod ? I think this would vary across CNIs. For example, the ovs-cni would need /var/run/openvswitch/db.sock file.

design-proposals/nic-hotplug/nic-hotplug.md
```diff
@@ -0,0 +1,234 @@
...
+the same process.
+
+Refer to the sequence diagram below for more information:
+
+![Kubernetes native solution](controller-based-approach.png)
+
+This alternative seems to be better aligned to Kubernetes patterns, requires
+less code, and is also aligned with
+[Kaloom's multus fork](https://github.com/kaloom/kubernetes-kactus-cni-plugin/#how-the-podagent-communicate-the-additiondeletion-of-a-network-attachment-into-a-running-pod).
+
```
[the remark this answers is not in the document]
maiqueb: I do not follow the implication. What would be this new annotation's purpose ?

design-proposals/nic-hotplug/nic-hotplug.md
maiqueb: Done.

Please let me know if the info I've added is clear enough.

design-proposals/nic-hotplug/nic-hotplug.md
```diff
@@ -0,0 +1,353 @@
...
+less code, is aligned with
+[Kaloom's multus fork](https://github.com/kaloom/kubernetes-kactus-cni-plugin/#how-the-podagent-communicate-the-additiondeletion-of-a-network-attachment-into-a-running-pod), and finally, also follows [KubeVirt's razor](https://github.com/kubevirt/kubevirt/blob/main/docs/architecture.md#the-razor),
+since this solution also allows adding and removing network interfaces to/from
+pods.
+
+#### Expose interface creation / removal as a service
+TODO
+
+## KubeVirt
+In order to mimic the disk hotplug feature, the proposed API also follows the
```
phoracek, line 137: Could you remind the reader why is this better than making the interfaces section mutable? As a user, I would find it little more convenient.

I understand that this is what was done with disk hotplug. But since disk hotplug was the first PoC of this "hotplugging" API, maybe living with it provided some more insight that would make us reconsider? Just making sure we just don't take the subresource API for granted.

maiqueb: Done. Is it clearer now ? 

I'd rather have the same look and feel (no matter how clunky ...) of the API for everything hotplug related, rather than  APIs having entirely different looks; the former alternative is - imo - bound to confuse the users. 

We can take the opportunity to start a discussion on "who gets to update the vmi spec", but, I'd welcome if that happens in a different document.

phoracek: It's good now, thanks

design-proposals/nic-hotplug/nic-hotplug.md
maiqueb: This is a good point. 

We must ensure the delegate CNI runs in the host's mnt namespace - either that, or each individual user must know which things to bind mount into the multus pod when deploying it... Which sounds totally unreasonable to me.

I'm evaluating ways to do that, but it doesn't seem to be a deal breaker.

Thanks for raising this question.

design-proposals/nic-hotplug/nic-hotplug.md
```diff
@@ -0,0 +1,359 @@
...
+A "thin CNI plugin" runs as a one-shot process, typically as a binary on disk
+executed on a Kubernetes host machine.
+
+A "thick CNI Plugin", on the other hand, is a CNI component composed of two (or
+more) parts, usually composed of "shim", and a long lived process (daemon)
+resident in memory. The "shim" is a lightweight "thin CNI plugin" component that
+simply passes CNI parameters (such as JSON configuration, and environment
+variables) to the daemon component, which then processes the CNI request.
+
+To transform multus into a thick plugin, it is needed to instantiate a long
```
dougbtv, line 69: Unsure where else to comment this, but... something I'd like to have done as part of the thickening is to use a more-standard authentication to k8s api via client-go rather than what's in the current thin plugin -- the generated KUBECONFIG. This sometimes causes intermittent auth failures which flake out CI. I think if we rely on the service account auth tokens and use client-go, we'll clear that up.

design-proposals/nic-hotplug/nic-hotplug.md
maiqueb: I see two alternatives to ensure the delegate CNI runs in the host's mount namespace (and as a result has direct access to whatever things it might expect):
1. do not refactor multus as a thick plugin. Introduce a controller listening for pod network annotation updates, and have the controller spawn a new process (the "thin" CNI plugin) running on the host's PID / net / mount namespaces.
2. stick to the thick plugin refactor, and run the whole daemon side of it on the host's mount namespace.

I think the former alternative is preferable, since it prevents the controller from running directly on the host's mount namespace, thus limiting the attack surface.

Both options require the multus daemon to run in the host's PID namespace.

design-proposals/nic-hotplug/nic-hotplug.md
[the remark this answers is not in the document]
maiqueb: I think I see; you suggest introducing two new annotations, one that stores the revision for the `networks` annotation, and another to store the revision of the `network-status` annotation.

IIUC, the goal is to have them in-synch - i.e. both revision numbers should match. If your `network`s revision number is higher than your `network-status` annotation, the pod's networks are not in synch yet.

design-proposals/nic-hotplug/nic-hotplug.md
maiqueb: @pperiyasamy would appreciate your input :)

pperiyasamy: @maiqueb +1 for first approach for running the controller process on the host by reusing same kubeconfig generated from multus entrypoint.sh. now the question is how can we run a process on the host's PID / net / mount namespaces together from the pod ? is there a way to inject host mount namespace into container ? have you tried this before ?

maiqueb: Yes. 

Sharing the host's PID / net namespaces is configurable per pod via configuration: https://kubernetes.io/docs/concepts/policy/pod-security-policy/#host-namespaces

The mount namespace is a bit trickier, but doable. Check the latest push, I got a snippet of how to do it.

It exploits the "trick" that since we share the host's PID namespace already, process 1 in the container **is** process 1 on the host, and as such, we just choose to use that process mnt namespace instead of the pod's.
 
The real question is: are we allowed to share the host's PID namespace ?... Everything else is pretty much just a couple of syscalls.

pperiyasamy: @maiqueb Thanks! I had a doubt on sharing host's mnt namespace and your code snippet clarified. then this should work, I guess.

design-proposals/nic-hotplug/nic-hotplug.md
```diff
@@ -0,0 +1,407 @@
...
+Hot-plug / hot-unplug (add / remove) nics to running VMs is an industry
+standard available in multiple platforms, allowing the dynamic attachment of L2
+networks. This is useful when the workload (VM) cannot tolerate a restart when
+attaching / removing networks, of for scenarios where, for instance, the
+workload is created prior to the network.
+
+This would help bridge the gap between KubeVirt's feature set and what the user
+base expects from it.
+
+## Goals
```
s1061123, line 16: Could you please add 'container/Pod' in addition to VMs because multus is CNI ('container' network interface) plugin and multus need to support container...

design-proposals/nic-hotplug/nic-hotplug.md
```diff
@@ -0,0 +1,407 @@
...
+    - /sbin/init
+```
+
+The aforementioned update will trigger multus to start the CNI ADD flow for the
+network named `macvlan-conf-2`.
+
+### Hotplug for VMs
+TODO
+
+## Functional Testing Approach
```
s1061123, line 387: Could you please add Pods test case in testing for regression?

design-proposals/nic-hotplug/nic-hotplug.md
```diff
@@ -0,0 +1,407 @@
...
+To transform multus into a thick plugin, it is needed to instantiate a long
+lived process - which will be the multus pod entrypoint - listening to a unix
+domain socket - this socket must be available both in the multus pod and the
+hosts's mount namespaces; as such, a bind mount to host this socket must be
+provided for the multus pod.
+
+The CNI shim will then be invoked by kubelet, and send the CNI ADD/DELETE
+commands to the server side of multus via the unix domain socket previously
+mentioned.
+
```
s1061123, line 126: Is there any concerns for CMD_DEL? In container/Pod case, what is happen if the net-attach-def, which contains default route, is removed? How to get new default route?

maiqueb: Looking at kaloom's code, I see that they send `CNI_DEL` for each pod connected to the NAD that got removed: https://github.com/kaloom/kubernetes-podagent#deleting-a-network-attachment . IMO, that would orphan out the running pod default route. I guess it might not be that big of a deal, since - iiuc - you have to be an admin to remove NADs. 

I do not know what is more correct; to do that, **or** have a validating webhook that prevents the deletion of a network **whenever** there are pods attached to it. I feel inclined to say the latter is the preferable choice, but I'm not sure. 

I think the community has a say in this. We could take this to this week's community call.

design-proposals/nic-hotplug/nic-hotplug.md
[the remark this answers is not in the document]
maiqueb: I can't come up with anything better. I'll add these. 

Thanks for spotting this gap.

design-proposals/nic-hotplug/nic-hotplug.md
maiqueb: This is a document intended for KubeVirt, but since I'm using it also to describe plenty of behavioral changes for multus, it's only fair that I add pods as goals. 

Will add these, along with a short disclaimer that "adding / removing network interfaces to pods **is a requirement** for adding them to VMs".

design-proposals/nic-hotplug/nic-hotplug.md
maiqueb: FWIW, I see in [their docs](https://github.com/kaloom/kubernetes-podagent#adding-a-network-attachment) that:

> uses consistent device naming based on the network attachment name, it’s the first 12 characters of the md5 hash of the name prefixed with net, for green it’s net9f27410725ab

design-proposals/nic-hotplug/nic-hotplug.md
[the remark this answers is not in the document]
maiqueb: Done.

design-proposals/nic-hotplug/nic-hotplug.md
maiqueb: Done.

design-proposals/nic-hotplug/nic-hotplug.md
maiqueb: Done.

maiqueb (commented):
Updated the PR as a reaction to @s1061123 's comments in this [forced push](https://github.com/kubevirt/community/compare/d23cdaa78d3e67e093bf8b52543259a4e752f62f..cb998a8d5172ec827123eca663dbb8635d3e3ca9).

I'd appreciate another review round.

design-proposals/nic-hotplug/nic-hotplug.md
s1061123: multus has 'default-route' feature in network annotation, which allows user to change default route to net-attach-def network. As far as I know of, Kactus does not have this feature.

If user pod connects to 'net-attach1', 'net-attach2' and 'net-attach3' with 'cluster network' and pod's default route is to 'net-attach2', then if user changes pod network to remove 'net-attach2'. In this case, user want not only to remove default route, also to change default route to cluster network. It does not make it by just invoke `CMD_DEL`.

Could you please take into account it?

maiqueb: Oh I see, I had misunderstood your comment. 

What if the user want the default route in another network ? Reacting to NAD updates is not in scope of this feature for KubeVirt. IMO, the scenario you point out is a corner case of that. 

I think the fairest option here is to either document the limitation, or create a follow-up story in multus to address this scenario. 

Out of curiosity, how would this be done technically ? Re-run the primary/default CNI plugin indicating that it is now the default route ?

rmohr:
/approve

everything seems to be where it should be from the KubeVirt perspective.

PR #141, opened 2021-10-08
Design Proposal: AMD SEV support in KubeVirt

vasiliy-ul:
Initial design proposal targeting the support of confidential computing with AMD SEV in KubeVirt.

PoC PR (launching of SEV VMs, no attestation): https://github.com/kubevirt/kubevirt/pull/6166

### DIFF
--- a/design-proposals/amd-sev.md
+++ b/design-proposals/amd-sev.md
@@ -1,0 +1,180 @@
+# Overview
+
+[Secure Encrypted Virtualization (SEV)](https://developer.amd.com/sev/) is a feature of AMD's EPYC CPUs that allows the memory of a virtual machine to be encrypted on the fly. This document describes a design proposal for adopting the technology in KubeVirt.
+
+This is the first step towards enabling confidential computing in KubeVirt. The proposal focuses on the basic AMD SEV. Later it can be extended to cover the extensions: `ES` (Encryptes State) and `SNP` (Secure Nested Paging) as well as other technologies like `Intel TDX`.
+
+## Motivation
+
+Without encryption any information stored in RAM can be compromised. This problem is especially relevant for the cloud environments where users run their workloads. Malicious software deployed on the patform or the administrator can get access to a running VM and that may lead to data leakage.
+
+SEV attempts to address that and to reduce the risk of leaking sensitive information. It transparently encrypts the memory with a key that is unique for each VM. Additionally it also provides a measurement of the memory content that can be used by the owner of the VM to perform the attestation thus showing that the memory was encrypted correctly.
+
+## Goals
+
+- Provide a way for KubeVirt users to run a VM with SEV
+- Provide an interface to allow KubeVirt users to perform the attestation
+
+## Non Goals
+
+- Initialization of the SEV platform, keys provisioning and management
+
+    Potentially there can be other 'SEV users' on the platform apart from KubeVirt. Therefore initialization, keys provisioning and management is not included in the scope of this document at this step.
+
+- Automation of the attestation process
+
+    This directly affects user expeience but from the guest owner perspective KubeVirt is a part of the platform and therefore cannot be fully trusted.
+
+## Definition of Users
+
+The feature is intended for the users who operate VMs with sensitive data and who do not want to put additional 'trust' in the cloud platform.
+
+## User Stories
+
+- As a cluster administrator, in order to provide better confidentiality of the users' data, I want to allow running the VMs with memory encryption enabled.
+- As a cluster user, in order to not have to share my secrets with the administrator, I want to be able to run my VMs with AMD SEV functionality enabled.
+- As a cluster user, in order to prevent leaking of my data in case when the hypervisor software is compromised, I want to run my VMs with AMD SEV functionality enabled.
+
+## Repos
+
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+
+There are two major parts that need to be implemented. First the domain xml parameters need to be properly configured in order to match the launch requirements and run a VM with SEV. Then the second step implies providing the interfaces to allow a user to perform attestation of a running VM.
+
+## Launching SEV guests
+
+SEV parameters can be specified via the `launchSecurity` element of the domain xml:
+
+```xml
+<domain>
+  ...
+  <launchSecurity type='sev'>
+    <policy>0x0001</policy>
+    <cbitpos>47</cbitpos>
+    <reducedPhysBits>1</reducedPhysBits>
+    <dhCert>RBBBSDDD=FDDCCCDDDG</dhCert>
+    <session>AAACCCDD=FFFCCCDSDS</session>
+  </launchSecurity>
+  ...
+</domain>
+```
+
+To start a SEV VM there is a need to specify `cbitpos` and `reducedPhysBits`. Those parameters are hypervisor-dependent and can be obtained from the `<sev></sev>` element of the `virsh domcapabilities`:
+
+```xml
+<domainCapabilities>
+...
+  <features>
+    ...
+    <sev supported='yes'>
+      <cbitpos>47</cbitpos>
+      <reducedPhysBits>1</reducedPhysBits>
+    </sev>
+    ...
+  </features>
+</domainCapabilities>
+```
+
+The `policy` parameter allows setting the guest policy flags. Additionally the optional `dhCert` and `session` provide the guest owner's base64 encoded `DH` (Diffie-Hellman) key and the guest owner's base64 encoded `launch blob` respectively.
+
+There are also several prerequisites to run SEV guests:
+
+- All virtio devices need to be configured with the `iommu='on'` attribute in their `<driver>` configuration
+- All memory regions used by the VM must be `locked` for Direct Memory Access (DMA) and to prevent swapping
+
+## [Attestation](https://github.com/AMDESE/sev-tool#proposed-provisioning-steps)
+
+### Certificate chain validation
+
+In order to ensure that the platform is not compromised the guest owner first needs to validate the certificate chain provided by the platform owner. On a system where KubeVirt is running, the privileged `virt-handler` pod can export the required certificates in advance to a `kbevirt-private` directory. Whenever a new VM requiring SEV is launched, the directory can be mounted to the `virt-launcher` pod in `read-only` mode. A user can then invoke the `virtctl` tool to download the certificate chain for validation (for that an additional command will be needed, e.g. `virtctl fetchcertchain ...`).
+
+### Validating the measurement
+
+The attestation for the basic SEV is performed before a VM is booted. It requers a user to provide the launch blob and the certificate (via `session` and `dhCert` paramters in the libvirt domain xml). A VM is launched in paused state thus allowing to retrieve and validate the measurement. Once it is verified, user can inject the secret into the VM and resume the execution.
+
+To support the attestation in KubeVirt `virtctl` can be extended to accept additional commands: `launchmeasure` and `injectsecret`. The commands will eventually end up with calling the corresponding libvirt API:
+
+- [virDomainGetLaunchSecurityInfo](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainGetLaunchSecurityInfo)
+
+    The call returns the launch measurement for a SEV guest.
+
+- [sev-inject-launch-secret](https://listman.redhat.com/archives/libvir-list/2021-May/msg00196.html)
+
+    The API is still missing.
+
+## Known limitations
+
+- SEV is only going to work with OVMF (UEFI)
+- SEV-encrypted VMs cannot contain directly-accessible host devices (that is, PCI passthrough)
+- Live Migration is not supported with SEV at the moment
+- The ammount of running SEV VMs on a node is limited
+- libvirt currently does not provide an API to inject a secret in a running VM
+
+Some of the limitations will probably be removed in the future.
+
+## API Examples
+
+**Requsting SEV feature in the VMI spec**
+
+```yaml
+spec:
+  domain:
+    launchSecurity:
+      sev:
+        policy:
+          - EncryptedState
+```
+
+or alternatively
+
+```yaml
+spec:
+  domain:
+    launchSecurity:
+      sev:
+        policy:
+          encryptedState: true
+```
+
+**Fetching the certificate chain**
+
+```
+virtctl fetchcertchain VMI
+```
+
+**Querying launch measure**
+
+```
+virtctl launchmeasure VMI
+```
+
+**Injecting launch secret**
+
+```
+virtctl injectsecret VMI --secret <base64>
+```
+
+## Scalability
+
+ N/A
+
+## Update/Rollback Compatibility
+
+The feature should not impact update compatibility.
+
+## Functional Testing Approach
+
+Functional tests should be provided to cover the primary use-case scenarios:
+
+- Launching of a SEV VM
+- Attestation
+
+**Note**: SEV-capable hardware is required to run the end-to-end tests.
+
+# Implementation Phases
+
+- Support running KubeVirt VMs with basic SEV
+- Once supported by libvirt, implement the interfaces to allow attestation
+- Extend the implementation to support new features/technologies (ES, SNP, TDX, etc.) as soon as libvirt and qemu adopt those
--- a/design-proposals/amd-sev.md
+++ b/design-proposals/amd-sev.md
@@ -23,7 +23,7 @@ SEV attempts to address that and to reduce the risk of leaking sensitive informa
 
 - Automation of the attestation process
 
-    This directly affects user expeience but from the guest owner perspective KubeVirt is a part of the platform and therefore cannot be fully trusted.
+    This directly affects user expeience but from the guest owner perspective KubeVirt is a part of the platform and therefore cannot be fully trusted. To facilitate the attestation process a guest owner can use external tools like [sev-tool](https://github.com/AMDESE/sev-tool) or [sevctl](https://github.com/enarx/sevctl).
 
 ## Definition of Users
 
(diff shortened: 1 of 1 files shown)


rmohr:
/cc @vladikr 

since you already took a look on the POC PR too.

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,245 @@
...
+
+The guest owner should get the assets listed bellow from the platform owner:
+
+- the `Platform Diffie-Hellman (PDH)` key exported from the SEV hardware
+- the complete key chain (up to the `root of trust`) used to sign the `PDH`
+- the OVMF image
+
+On a node where KubeVirt is running, the privileged `virt-handler` pod can export the required assets in advance to a `kubevirt-private` directory. Whenever a new VM requiring SEV is launched, the directory can be mounted to the `virt-launcher` pod in `read-only` mode. A user can then invoke the `virtctl` tool to download the certificate chain for validation (for that an additional command will be needed, e.g. `virtctl fetchcertchain ...`).
+
+In order to ensure that the platform is not compromised the guest owner needs to validate the `PDH` key using the provided chain. For that `sev-tool --validate_cert_chain ...` can be used (AMD root certificate should be downloaded from the web for verification). Additionally the UEFI image should be considered 'trustable' by the guest owner.
```
vladikr, line 101: Where should this be done, in the guest OS?

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Step #2 Establishing secure channel with the platform
+
+To establish a secure communication channel with the platform a guest owner needs to provide a launch blob with the parameters of the session and the certificate derived from `PDH`. This is currently not in the scope of KubeVirt as it requires the guest owner to expose the secrets. It should be performed 'offline' by the guest owner (e.g. by using `sev-tool --generate_launch_blob ...`).
```
vladikr, line 105: This part indeed sounds very tricky. Not sure how users can achieve this. 

What if the extraction of launch blob be done in an init container that would use a provided secret (I assume this is required if I read this correctly... ) and pass it to the virt launcher via env. variable?

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+The implementation in `virt-launcher` in turn can 'postpone' the creation of the domain until the parameters are specified. Sidecar hooks functionality can be used to modify the domxml after a VMI has been scheduled. New `virtctl` commands (e.g. `virtctl launchblob ...`) can be introduced to support the process and to trigger propagation of the parameters.
+
+To run a SEV VM on a specific node the parameters can be specified directly in the VMI spec:
+
+```yaml
+spec:
+  nodeSelector:
+    kubernetes.io/hostname: "node01"
```
vladikr, line 127: This is also can be done with a limited success, as there may be other requirements that will prevent the scheduler from placing the VMI on this node.

design-proposals/amd-sev.md
vasiliy-ul: This is the 'preparation' step and it is done outside of the gest OS. Here the guest owner takes the certificates extracted from the platform and verifies that they are properly signed.

design-proposals/amd-sev.md
vasiliy-ul: Right. There is no guarantee here.

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+To run a VM without going through the attestation process at all the parameters can be omited:
+
+```yaml
+spec:
+  domain:
+    launchSecurity:
+      sev: {}
+```
```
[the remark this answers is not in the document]
vasiliy-ul: I think there is value in having just simple implementation. Even if a user cannot do the attestation at the moment its still beneficial that the VM workload is run in confidential environment as it increases the level of data protection.

The PR https://github.com/kubevirt/kubevirt/pull/6166 enables SEV support in KubeVirt but does not touch the attestation topic. Maybe we could focus on that one first and then gradually proceed with enhacning the functionality. I more than welcome such an approach.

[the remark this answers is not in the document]
rmohr: I consent here with @vladikr  that we should at least have a clear plan to add the attestation process relatively fritcion-less later. @vasiliy-ul do I understand the flow correctly

 * VM gets scheduled
 * Users fetch the host cred chain, generate the blob and inject it
 * VM starts in pause mode and the measurement takes place
 * Users verify the measurment and allow by injecting the final secret the actual start
 
I guess it would be great if we could come up with an overall-flow of something like this:

```
kubectl create -f def.yaml
kubectl virt sev attestate #goes through all the steps as above and informs users of the progress
```

It would be great if we can provide for all relevant steps in the flow proper subresources (not necessarily one subresource for each step).

I would love to see clear apis here where interaction is needed and no things like entering pods to read some files from somewhere.

In addition we can add additionl debugging tooling subcommands like

```
kubectl virt sev showchain
```

and so forth.

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,180 @@
...
+
+    The API is still missing.
+
+## Known limitations
+
+- SEV is only going to work with OVMF (UEFI)
+- SEV-encrypted VMs cannot contain directly-accessible host devices (that is, PCI passthrough)
+- Live Migration is not supported with SEV
+- The ammount of running SEV VMs on a node is limited
+- libvirt currently does not provide an API to inject a secret in a running VM
```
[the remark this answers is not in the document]
rmohr: I think I have a much better understanding now. Thanks.

design-proposals/amd-sev.md
vasiliy-ul: Yeah, maybe the explanation is a bit confusing. The launch blob is not extracted. Here the guest owner is responsible to geneate the blob. I think it is possible to pass certain parameters to virt-launcher but I dont think it will make things much easier. The procedure here implies also generation of several public/private key pairs and only the guest owner is supposed to have access to them (to private ones). If this task was to be offloaded to some init container then we would probably need to include it in the attestation as well.

The procedure is described in more details here https://github.com/enarx/sev/tree/main/docs/attestation#secure-channel-establishment. Its indeed tricky. In general I started thinking in the direction of having some external service for facilitating the attestation. Maybe KubeVirt could just provide the APIs to allow establishing this secure channel and injecting the secret afterwards. In such case the service would need to be 'trusted' by the guest owner but KubeVirt would in some sense stay away from the 'sensitive' data.

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,180 @@
...
+
+**Requsting SEV feature in the VMI spec**
+
+```yaml
+spec:
+  domain:
+    launchSecurity:
+      sev:
+        policy:
+          - EncryptedState
```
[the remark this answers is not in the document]
rmohr: Based on your verification process which you describe, it sounds to me like we could however involve virtctl to do the heavy lifting of the whole verification process while still providing clear REST apis which also virtcl uses so that people could come up with their own flows entirely. Is that correct?

design-proposals/amd-sev.md
vasiliy-ul: > With other words users do something like this?

Correct.

design-proposals/amd-sev.md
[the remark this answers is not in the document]
vasiliy-ul: > I consent here with @vladikr that we should at least have a clear plan to add the attestation process relatively fritcion-less later.

No objections here. Though for me it seems that the two topics (launching and attesting) can be handled somewhat independently. Simply launching a SEV VM (without attestation) is purely a technical process (it needs just some configuration tweaks, validations and VMI spec API). Attestation may affect the launching in a sense that e.g. with attestation we need to start the VM paused. But that is IMHO an independent flow and should not affect the scenario without attestation.

> @vasiliy-ul do I understand the flow correctly

Yes, that is absolutely correct. And I basically agree with the rest about the subresources and APIs. I will probably look more into that direction further.

design-proposals/amd-sev.md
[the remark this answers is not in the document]
vasiliy-ul: Right, I though I had mentioned it somewhere in the doc that the idea was to have virtctl commands backed by corresponding API. So that virtctl can be used as a 'client' to interact with VMs during attestation (and maybe to facilitate the process a bit). At the same time the APIs will provide some space for further improvements.

design-proposals/amd-sev.md
vasiliy-ul: Here a user needs to somehow get the platform certificates from node01 upfront. AFAIK that's all what is required for providing the needed parameters `dhCert` and `session`.

design-proposals/amd-sev.md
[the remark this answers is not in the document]
rmohr: Sounds good to me.

design-proposals/amd-sev.md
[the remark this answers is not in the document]
rmohr: > Right, I though I had mentioned it somewhere in the doc that the idea was to have virtctl commands backed by corresponding API. So that virtctl can be used as a 'client' to interact with VMs during attestation (and maybe to facilitate the process a bit). At the same time the APIs will provide some space for further improvements.


What confused me is this section:

```
On a node where KubeVirt is running, the privileged `virt-handler` pod can export the required assets in advance to a `kubevirt-private` directory. Whenever a new VM requiring SEV is launched, the directory can be mounted to the `virt-launcher` pod in `read-only` mode. A user can then invoke the `virtctl` tool to download the certificate chain for validation (for that an additional command will be needed, e.g. `virtctl fetchcertchain ...`).
```

I somehow had the impression that here `virtctl` would do some strange things to get the certificates. maybe just add there that `virtctl` will query `virt-handler` to get the certificate and it works for me.

design-proposals/amd-sev.md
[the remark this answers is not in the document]
vladikr: Sounds great. We can start with the simple launch flow and then come back a refine the attestation.

vasiliy-ul:
@vladikr , @rmohr Thanks for the feedback. I tried to address the comments by focusing more on the APIs. Maybe you could take one more look to ensure that's the right direction and we are on the same page? The APIs are not *fully* defined yet, just at high-level. For the complete definition with proper signatures and arguments I would first try to do a PoC PR with the attestation. I think that might give a better overview of how everything fits together (WDYT?). Also, any thought whether something is missing for the part of launching without the attestation?

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,253 @@
...
+
+or alternatively
+
+```yaml
+spec:
+  domain:
+    launchSecurity:
+      sev:
+        policy:
+          encryptedState: true
```
vladikr, line 205: Do we expect to have multiple policies?

vladikr (commented):
It looks much better to me. Thanks @vasiliy-ul

design-proposals/amd-sev.md
vasiliy-ul: Yes, there are multiple policy flags. They are defined in the [SEV spec](https://www.amd.com/system/files/TechDocs/55766_SEV-KM_API_Specification.pdf) (Chapter 3 Guest Policy). You can also find those in the [SEV implementation PR](https://github.com/kubevirt/kubevirt/pull/6166):

```go
type SEVPolicy struct {
	// Sharing keys with other guests is disallowed.
	// Defaults to false.
	// +optional
	NoKeysSharing *bool `json:"noKeysSharing,omitempty"`
	// SEV-ES is required.
	// Defaults to false.
	// +optional
	EncryptedState *bool `json:"encryptedState,omitempty"`
	// Sending the guest to another platform is disallowed.
	// Defaults to false.
	// +optional
	NoSend *bool `json:"noSend,omitempty"`
	// The guest must not be transmitted to another platform that is not in the domain.
	// Defaults to false.
	// +optional
	Domain *bool `json:"domain,omitempty"`
	// The guest must not be transmitted to another platform that is not SEV capable.
	// Defaults to false.
	// +optional
	SEV *bool `json:"sev,omitempty"`
}
```

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,253 @@
...
+
+In order to ensure that the platform is not compromised the guest owner needs to validate the `PDH` key using the provided chain. For that `sev-tool --validate_cert_chain ...` can be used (AMD root certificate should be downloaded from the web for verification). Additionally the UEFI image should be considered 'trustable' by the guest owner.
+
+### Step #2 Establishing secure channel with the platform
+
+To establish a secure communication channel with the platform a guest owner needs to provide the launch blob with the parameters of the session and the certificate derived from `PDH`. This can be suported in KubeVirt by `/sevsetupsession` VMI endpoint with the parameters `session` and `dhCert` (which will be mapped to the corresponding elements in the domain xml). New `virtctl sevsetupsession ...` command can be introduced to trigger propagation of the parameters.
+
+The implementation in `virt-launcher` in turn can 'postpone' the creation of the domain until the parameters are specified. Sidecar hooks functionality can be used to modify the domxml after a VMI has been scheduled.
```
rmohr, line 133: I am not sure if the sidecar sentence has any relevance here?

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Step #1 Validation of the platform
+
+The guest owner should get the assets listed bellow from the platform owner:
+
+- the `PDH` (Platform Diffie-Hellman) key exported from the SEV hardware
+- the complete key chain (up to the `root of trust`) used to sign the `PDH`
+- the OVMF binary
+
+KubeVirt can expose a VMI subresource endpoint `/sevfetchcertchain` which will return base64 encoded certificates. Also a new command can be added to virtctl tool to facilitate the process: `virtctl fetchcertchain ...`.
```
rmohr, line 123: coming up with a good api is definitely something which we can explre in the initial PoC. In general it may make sense to go with a `/sev/` subpath: `/sev/certchain` (GET, POST) for instance.

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,253 @@
...
+      sev:
+        policy:
+          encryptedState: true
+```
+
+**VMI subresource endpoints**
+
+The following VMI subresource endpoints can be introduced for the attestation process:
+
+- *Fetching the certificate chain*: `/sevfetchcertchain`
```
rmohr, line 212: A `/sev/` path element should make things nicer.

rmohr (commented):
This works for me.  @vasiliy-ul thanks. Maybe creata a `/sev/` path element in the rest calls.

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,253 @@
...
+    ```
+    virtctl sevfetchcertchain VMI
```
rmohr, line 214: Also on the commandline, I would recommend introducing a subcommand `sev`:

```
virtctl sev fetchcertchain VMI
```

design-proposals/amd-sev.md
vasiliy-ul: Removed for now.

rmohr:
/lgtm

from me. Looks good enough for me to sort out the rest on the PoC. @vladikr your turn :)

vladikr:
/lgtm
Looks good to me as well! Thank you, @vasiliy-ul !

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,253 @@
...
+SEV-capable nodes should be properly labeled. Then a corresponding node selector can be added to VM pods so they are scheduled correctly.
+
+## Attestation
+
+[Attestation](https://github.com/AMDESE/sev-tool#proposed-provisioning-steps) is an interactive process between `platform owner` and `guest owner`. It is carried out to establish a certain level of trust. To leverage the benefits of memory encryption a guest owner provides an encrypted VM disk. To decrypt the data and to run a VM with such a disk there is a need to `inject a secret` into the VM when the system is about to boot. The secret is known only by the guest owner. The goal of the attestation process is to validate that the platform is genuine and it really runs a non-compromised SEV hardware. Also it is needed to establish a private communication channel with the platform so the guest owner can securely provide the secret. Attestation consists of the following steps:
+
+- VM gets scheduled on a SEV-capable node
+- guest owner validates the platform on that specific node
+- guest owner establishes a secure channel with the platform by providing the session parameters
+- VM is started in a paused state
```
berrange, line 109: The multi-stage dance where the mgmt app has to separate VM startup into two phases, with user interaction in between scheduling a VM to a specific host and then later starting it on this host, is one of the major pain points in implementing support for measured boot with SEV.  It makes the SEV VM launch flow significantly different from a normal VM startup where scheduling the VM on a host and launching the VM are essentially part of the same logic operation.

The further complexity where by mgmt apps have to expose extra information about the host SEV capabilities, the guest measurement and add APIs fo secret injection are a further burden on the support for SEV VMs.

It further creates a burden on the guest owner, because to run trusted VMs, they need to use an attestation agent and this agent now needs to implement support for the different APIs exposed by many different cloud vendor, to avoid vendor lockin.

To try to address this extra burden on mgmt apps and users I have raised the question of whether libvirt can facilitate automation for almost all of this attestation related extra work. It would reduce the problem of enabling SEV attestation to merely providing a URL pointing to the attestation agent controlled by the guest owner, with libvirt then directly talking to this agent. 

https://listman.redhat.com/archives/libvir-list/2022-February/msg00857.html

With such a proposal the implementation of AMD SEV in KubeVirt would be significantly simpler IMHO, so would appreciate feedback from KubeVirt maintainers on whether this idea is appealing.

design-proposals/amd-sev.md
```diff
@@ -0,0 +1,253 @@
...
+
+The blob with the encrypted secret can be built by the guest owner using `sev-tool --package_secret ...`.
+
+## Known limitations
+
+- SEV is only going to work with OVMF (UEFI)
+- SEV-encrypted VMs cannot contain directly-accessible host devices (that is, PCI passthrough)
+- Live Migration is not supported with SEV at the moment
+- The ammount of running SEV VMs on a node is limited
+- libvirt currently does not provide an API to inject a secret in a running VM
```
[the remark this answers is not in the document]
vasiliy-ul: Right, thanks for bringing this up. It was written some time ago when the relevant API was not available. The update of the proposal is one of the TODO items in the tracking issue: https://github.com/kubevirt/kubevirt/issues/6991

design-proposals/amd-sev.md
vasiliy-ul: I think the proposed approach with handling attestation in libvirt is definitely valid. It may indeed simplify the implementation  here. KubeVirt will not need to expose all the APIs for handling the attestation. Though it will not eliminate the "two phases" (i.e. 1) scheduling on a node, verifying the certs, preparing the launch blob and then 2) launching paused, measuring, injecting secret, upause) but all that will happen behind the scenes without the need to involve the guest owner. 

The only thing that bothers me in that flow is the attestation server and how it fits into the SEV trust model. The guest owner will need to share the secret with this service. So I would assume it should be controlled solely by the guest owner. Maybe even running locally?... Just thoughts...

Anyway, I the idea with libvirt handling the attestation seems good to me. Would be great to see that moving forward.

Maybe other thought/opinions here?

berrange: > Though it will not eliminate the "two phases" (i.e. 1) scheduling on a node, verifying the certs, preparing the launch blob and then 2) launching paused, measuring, injecting secret, upause) but all that will happen behind the scenes without the need to involve the guest owner.

It would absolutely remove eliminate it from KubeVirt's POV too. It merely becomes a normal 'virDomainCreate' API call for KubeVirt as with any other non-SEV VM.   Libvirt simply blocks the virDomainCreate API impl from spawning QEMU until it has received the launch blob & certs back from the attestation agent.

> The only thing that bothers me in that flow is the attestation server and how it fits into the SEV trust model. The guest owner will need to share the secret with this service. So I would assume it should be controlled solely by the guest owner.

Yes the attestation server is something that has to be trusted by the guest owner. This means running it somewhere the guest owner considers to be trustworthy, which either means it has to be outside the untrustworthy cloud, or they need to somehow bootstrap the trust relationship for a 1st VM and then run their agent in this VM. The latter approach would mean that the cloud would actually still need to expose the info about SEV in its APIs, which is kinda defeats some of the benefits.

vasiliy-ul: Raised this in the KubeVirt mailing list https://groups.google.com/g/kubevirt-dev/c/a6XlT0qRNAY to have some attention from a broader audience.

rmohr: I would love to see some improvements on the flow.

> > The only thing that bothers me in that flow is the attestation server and how it fits into the SEV trust model. The guest owner will need to share the secret with this service. So I would assume it should be controlled solely by the guest owner.
> 
> Yes the attestation server is something that has to be trusted by the guest owner. This means running it somewhere the guest owner considers to be trustworthy, which either means it has to be outside the untrustworthy cloud, or they need to somehow bootstrap the trust relationship for a 1st VM and then run their agent in this VM. The latter approach would mean that the cloud would actually still need to expose the info about SEV in its APIs, which is kinda defeats some of the benefits.

Here are some potential pitfalls if we go down that path:

 * As already mentioned, a public accessible trusted place needs to exist:
   What if we deal with an air-gaped installation? There one typicall has access to clusters via e.g. a bastion, or with other means, typically going one direction (kubectl can reach the cluster, but not necessarily the other way round).
 * Libvirt has no network access in the kubevirt design. We could somehow work around that by routing through virt-handler, but having libvirt trying to reach external places, is not ideal from a kubevirt architecture

In general, from a k8s architecture perspective, if we talk about automating flows like this, a controller model may fit better. The controller watches the state of the VMI, goes through the atestation process and approves the start without further user interaction. The benefit of this is, that controllers only need access to the cluster in a similar way like `kubectl`. They can be deployed in the cluster (limited trust, but may still help to help preventing breaches on individual nodes), or it can be deployed anywhere on the world.

I understand that this would eliminate the benefit of a common protocol. However, if libvirt would decide to implement such a REST-based approach, we could still make use of it internally, or still use the other libvirt client calls, so we would not be in the way. I am just not sure if the reach-out approach from libvirt is the right pattern to be useful for all kinds of cloud-services and scenarios.

vasiliy-ul: > Libvirt has no network access in the kubevirt design.

Yes but there are already a couple of use-cases where KubeVirt proxies the data transfer: console/VNC or live migration. Communication with the atestation service can be proxied as well. In that case it should be sufficient to have access to the cluster via kubectl/virtctl. Or do you think it may have some implications?

> The controller watches the state of the VMI, goes through the atestation process and approves the start without further user interaction.

I think in case of the such a model a guest owner will have to share the secret with the controller. That in my opinion does not fit well with the idea behind SEV where a guest owner does not want to expose the secrets to the cloud provider/infra. I hardly see how it can be done without user interaction. Only maybe if there is an "attestation service" that will talk on behalf of the user with that controller. Or did I understand the idea wrong?

rmohr: > > Libvirt has no network access in the kubevirt design.
> 
> Yes but there are already a couple of use-cases where KubeVirt proxies the data transfer: console/VNC or live migration. Communication with the atestation service can be proxied as well. In that case it should be sufficient to have access to the cluster via kubectl/virtctl. Or do you think it may have some implications?

Yes that would be possible. However having the need to create port-forwards over the place til it reaches the user looks like not the right solution.


> > The controller watches the state of the VMI, goes through the atestation process and approves the start without further user interaction.
> 
> I think in case of the such a model a guest owner will have to share the secret with the controller. That in my opinion does not fit well with the idea behind SEV where a guest owner does not want to expose the secrets to the cloud provider/infra. I hardly see how it can be done without user interaction. Only maybe if there is an "attestation service" that will talk on behalf of the user with that controller. Or did I understand the idea wrong?

I agree with what you say. All I meant is, that a controller-like pattern on the user-side (or trusted remote side) may suite better than letting libvirt reach out. I just used `controller` in the sense of describing the pattern. This could for instance just be part of virtctl. At the end with some tool (even if it is just on the machine of the user) the secret has to be exposed.

rmohr: For more context, while I am not necessarily promoting the idea of creating additional k8s objects for SEV, it has some similarities to certificate signing request flows in k8s, where certain things, like joining new nodes, are blocked until someone or something examines the CSR and approves it explicitly.

berrange: > - As already mentioned, a public accessible trusted place needs to exist:
>    What if we deal with an air-gaped installation? There one typicall has access to clusters via e.g. a bastion, or with other means, typically going one direction (kubectl can reach the cluster, but not necessarily the other way round).
> - Libvirt has no network access in the kubevirt design. We could somehow work around that by routing through virt-handler, but having libvirt trying to reach external places, is not ideal from a kubevirt architecture

So when I say that libvirt would be able to connect to an attestation server, it doesn't necessarily require that libvirt have a direct outbound connection to the server. KubeVirt is not so unusual in not allowing the compute host to have arbitrary outbound network access, so it is a general concern. In KubeVirt case libvirt is admittedly even more restricted than some other clouds, but it isn't a show stopper.

Another concern is that the user may not have somewhere to run their attestation server that allows inbound network access. eg if running it on your laptop it can be behind NAT, so k8s would not be able to reach it.

A way to address all these issues would be for the cloud infrastructure to provide a transparent proxy that sits in between libvirt and the end user's attestation server. The proxy could sit on the internet perimeter of the cloud infra, such that the end users attestation service can establish a connection to it, and then libvirt can also establish a connection to it from the compute host, then it just forwards traffic in both directions. This would avoid the NAT issue for a guest owner running a remote attestation service, and also solve the issue of libvirt not having general outbound access. The KubeVirt case might involve multiple hops for the proxying. Ultimately this isn't all that different from exposing QEMU's VNC console, aside from the fact that libvirt reaches out, rather than a user reaching in eg we're just reversing one arrow in the comparison of:

   user -> proxy -> vnc

vs

  user  -> proxy <- libvirt SEV

I very much doubt the attestation service protocol would be anything tied to libvirt explicitly, as its design is really driven by what information needs to be provided by the two end points. So if we get a standardized protocol for attestation, it is certainly possible for KubeVirt to act as the endpoint on the cloud side, and make libvirt API calls to complete the work, if desired. 

Finally one thing to bear in mind is that what we're considering here has primarily only taken into account AMD SEV and AMD SEV-ES since these work in essentially the same way, the only difference being extra CPU register state in the SEV-ES case. 

With the newer AMD SEV-SNP the architecture is significantly changed. Instead of attestation being initiated on the host before guest CPUs run, with SNP it is initiated inside the guest after guest CPUs run. With such  a scheme attestation needs to take place as soon as possible during guest boot. With even the initrd arguably being inconveniently late, it is quite possible it will be triggered from inside the guest firmware (OVMF) before it loads the kernel. The guest firmware is can't be assumed to have network access to reach the guest owner's attestation service, so it is highly likely it will have to proxy info back into the host OS (with something like virtio-vsock or virtio-serial) whereupon QEMU and/or libvirt can pick it up and forward it on the real attestation service.

IOW, the architectural concerns raised here for SEV and SEV-ES wrt whether libvirt has ability to connect to a remote attestation server might be possible to avoid by KubeVirt taking more ownership of the exchange. When it comes to SEV-SNP though, KubeVirt's design might be forced, by the need to proxy from the guest back to the host. IOW the container in which libvirtd/qemu are running might have no choice but to provide some mechanism to connect back to an outside service.

NB, nothing about SEV-SNP is finalized, so this is just my best guess, but I feel we'll get something similar to this.

vasiliy-ul: Would it make sense to support both approaches? I dont see big conflict here. So KubeVirt may allow to do the attestation using the exported APIs as initially proposed in this PR (perhaps with a controller on top to automate the process and implement the `approve` flow). I think the use case mentioned here with bootstraping of a confidential VM with an attestation service will require smth like that. There might be other use cases as well. On the other hand, as soon as the proposed REST protocol is defined and the relevant bits are implemented in libvirt we could still support it in KubeVirt. That should give the users certain flexibility on how they do the attestation by also allowing them to bring in a 'standard' attestation service which follows that protocol.

PR #161, opened 2022-01-26
Update OWNERS: remove Chris and Pep

codificat:
@mazzystr and myself stepped down from community management.

This is to reflect that change in the repo's OWNERS file.

Should we add others?


cwilkers:
/lgtm
/approve
Welcome to emeritus status, we'll keep an honorary chair for you both!

PR #164, opened 2022-02-21
VM memory dump design proposal

ShellyKa13:
Signed-off-by: Shelly Kagan <skagan@redhat.com>

### DIFF
--- a/design-proposals/vm-memory-dump.md
+++ b/design-proposals/vm-memory-dump.md
@@ -5,7 +5,7 @@ Proposal for taking VM memory dump for analysis purposes.
 One of the methods of troubleshooting guest virtual machines is getting a memory dump and analyzing it. As this is a standard capability in most VM management systems, it is also required for kubevirt’s VMs.
 
 ## Goals
-Have a mechanism to get a VM memory dump.
+Have a mechanism to get a VM memory dump to be inspected with [Volatility3](https://github.com/volatilityfoundation/volatility3).
 
 ## Non Goals
 Memory dump is not a restorable output and currently will not be a part of the VM snapshot.
@@ -43,11 +43,11 @@ The VM subresource endpoint will recieve a rest request containing the pvc name 
 It will patch the vm status with a memoryDumpStatus stating the volume-name and Phase `Binding`.
 The memoryDumpStatus will look something like this:
 
-`
+```golang
 // MemoryDumpStatus represent the memory dump request status and info
 type MemoryDumpStatus struct {
-	// VolumeName is the name of the pvc that will contain the memory dump
-	VolumeName string `json:"volumeName"`
+	// ClaimName is the name of the pvc that will contain the memory dump
+	ClaimeName string `json:"claimName"`
 	// Phase represents the memory dump phase
 	Phase MemoryDumpPhase `json:"phase,omitempty"`
 	// TimeStamp represents the time the memory dump was completed
@@ -66,7 +66,7 @@ const (
 	// The memorydump is being unbound
 	MemoryDumpUnBinding MemoryDumpPhase = "Unbinding"
 )
-`
+```
 
 In the virt controller once getting the update of the memoryDump in state of binding it will bind the pvc to the vm by updating the vm volumes with the memoryDump pvc.
 After that the status of the memoryDump will be updated to `InProgress` which will be cause the volume to be mounted to the virt-launcher, same as done in the hotplug process, other then the end of the process that instead of attaching the volume to the VMI it will trigger the guest memory dump `virDomainCoreDump` command with `VIR_DUMP_MEMORY_ONLY` flag to be executed by virt-launcher.
@@ -115,7 +115,7 @@ items:
     Snip
     ....
     memoryDumpStatus:
-      volumeName: memory-dump
+      claimName: memory-dump
       phase: Completed
       timestamp: "2022-03-29T11:00:04Z"
 ```
@@ -142,6 +142,5 @@ New API should not affect updates / rollbacks.
 The implementation wil be split to several phases:
 * Add memory dump command in virt launcher server
 * Add virtctl command to execute memory-dump with existing pvc
-* Extend the virtctl command to support on-demand creation of PVC
 * Add remove memory dump virtctl command
-* Support block pvc for the dump (?)
+* Extend the virtctl command to support on-demand creation of PVC
--- a/design-proposals/vm-memory-dump.md
+++ b/design-proposals/vm-memory-dump.md
@@ -30,21 +30,22 @@ It will be possible to unbound the PVC and then do a memory dump to another PVC 
 #### trigger memory dump
 Run a virtctl command to get a memory dump. This new API will look as follows:
 
-`$ virtctl memoryDump my-vm --volume-name=memory-pvc`
+`$ virtctl memory-dump my-vm --volume-name=memory-pvc`
 
-In this case `memory-pvc` should already exist and be of a size big enough the contain the memory dump. A check will be made to make sure.
+In this case `memory-pvc` should already exist and be of a size big enough to contain the memory dump. A check will be made to make sure.
 It will be possible to ask for a PVC to be created with `--create` flag. In such case the required size will be calculated.
 
 ##### The process
 The trigger of memory dump will call a VM subresource that will add a memorydumpRequest to the VM status.
 The request will cause the binding of the PVC to the vm (if the pvc is not bounded yet) after that the pvc will be mounted to the virt-launcher that will trigger the guest memory dump `virDomainCoreDump` command with `VIR_DUMP_MEMORY_ONLY` flag to be executed by virt-launcher.
+(Check out [virDomainCoreDump](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainCoreDump) for reference)
 After the dump is complete it will unmount the pvc from virt-launcher.
 The PVC will remain bounded to the VM - will be kept in the VM volumes and have an updated status and timestamp in the VM status.
 
 #### remove memory dump from VM
 Run a virtctl command to remove a memory dump from a VM. This new API will look as follows:
 
-`$ virtctl removeMemoryDump my-vm`
+`$ virtctl remove-memory-dump my-vm`
 
 ##### The process
 The trigger of remove memory dump will call a VM subresource that will add a removeMemorydumpRequest to the VM status.
@@ -52,6 +53,8 @@ The request will unbound the PVC from the vm. The PVC will be deleted from the V
 
 ### Handle the memory dump
 After the memory dump is completed the user will be able to export this PVC and also if there is a snapshot supported storage class the PVC can be a part of the VM snapshot and then it can be unbounded and deleted.
+The output of the memory dump can be used for memory analysis with different tools for example [Volatility3](https://github.com/volatilityfoundation/volatility3) and maybe also [sleuthkit](https://www.sleuthkit.org/autopsy/)
+
 
 ## Update/Rollback Compatibility
 New API should not affect updates / rollbacks.
@@ -62,3 +65,4 @@ The implementation wil be split to several phases:
 * Add virtctl command to execute memory-dump with existing pvc
 * Extend the virtctl command to support on-demand creation of PVC
 * Add remove memory dump virtctl command
+* Support block pvc for the dump (?)
--- a/design-proposals/vm-memory-dump.md
+++ b/design-proposals/vm-memory-dump.md
@@ -73,6 +73,52 @@ After that the status of the memoryDump will be updated to `InProgress` which wi
 (Check out [virDomainCoreDump](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainCoreDump) for reference)
 After the dump is complete the timestamp and status of the memory dump will be updated and it will unmount the pvc from virt-launcher.
 The PVC will remain bound to the VM - will be kept in the VM volumes and have the updated status and timestamp in the VM status as mentioned.
+It will look something like that: (In the VMI there will be no information)
+
+```yaml
+---
+apiVersion: v1
+items:
+- apiVersion: kubevirt.io/v1
+  kind: VirtualMachine
+  name: my-vm
+  ....
+  Snip
+  ....
+    template:
+      spec:
+        domain:
+          devices:
+            disks:
+            - disk:
+                bus: virtio
+              name: datavolumedisk
+            - disk:
+                bus: virtio
+              name: cloudinitdisk
+        volumes:
+        - dataVolume:
+            name: my-dv
+          name: datavolumedisk
+        - cloudInitNoCloud:
+            userData: |-
+              #cloud-config
+              password: fedora
+              chpasswd: { expire: False }
+          name: cloudinitdisk
+        - memoryDump:
+            persistentVolumeClaim:
+                claimName: memory-dump-pvc
+          name: memory-dump
+  status:
+    ....
+    Snip
+    ....
+    memoryDumpStatus:
+      volumeName: memory-dump
+      phase: Completed
+      timestamp: "2022-03-29T11:00:04Z"
+```
 
 
 #### remove memory dump from VM
--- a/design-proposals/vm-memory-dump.md
+++ b/design-proposals/vm-memory-dump.md
@@ -1,0 +1,64 @@
+# Overview
+Proposal for taking VM memory dump for analysis purposes.
+
+## Motivation
+One of the methods of troubleshooting guest virtual machines is getting a memory dump and analyzing it. As this is a standard capability in most VM management systems, it is also required for kubevirt’s VMs.
+
+## Goals
+Have a mechanism to get a VM memory dump.
+
+## Non Goals
+Memory dump is not a restorable output and currently will not be a part of the VM snapshot.
+
+## User Stories
+As a Kubevirt user I would like to get a memory dump of a running VM so that I can later inspect its memory.
+
+## Repos
+[KubeVirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+To trigger a memory dump a virtctl memoryDump command will be added.
+The command will either use an existing PVC or create a new PVC on demand. This PVC will be bounded to the VM and will appear in the VM volumes spec with a fiting status.
+During the memory dump process this PVC will be mounted to the virt-launcher pod. After the mounting the guest memory will be dumped to that pvc and eventually it will be unmounted from the virt launcher.
+The PVC will remain bounded - in the VM spec. It is the users' responsibility to export/use the memory dump results before reusing the PVC for a new memory dump. As long as the PVC is bounded as a memory dump "container" each memory dump command will overwrite the previous content.
+It will be possible to track the last memory dump in the VM status with a new memory dump status that will be added there.
+It will be possible to unbound the PVC and then do a memory dump to another PVC that will be bounded instead. 
+
+## API
+
+### Get memory dump
+#### trigger memory dump
+Run a virtctl command to get a memory dump. This new API will look as follows:
+
+`$ virtctl memoryDump my-vm --volume-name=memory-pvc`
+
+In this case `memory-pvc` should already exist and be of a size big enough the contain the memory dump. A check will be made to make sure.
+It will be possible to ask for a PVC to be created with `--create` flag. In such case the required size will be calculated.
+
+##### The process
+The trigger of memory dump will call a VM subresource that will add a memorydumpRequest to the VM status.
+The request will cause the binding of the PVC to the vm (if the pvc is not bounded yet) after that the pvc will be mounted to the virt-launcher that will trigger the guest memory dump `virDomainCoreDump` command with `VIR_DUMP_MEMORY_ONLY` flag to be executed by virt-launcher.
+After the dump is complete it will unmount the pvc from virt-launcher.
+The PVC will remain bounded to the VM - will be kept in the VM volumes and have an updated status and timestamp in the VM status.
+
+#### remove memory dump from VM
+Run a virtctl command to remove a memory dump from a VM. This new API will look as follows:
+
+`$ virtctl removeMemoryDump my-vm`
+
+##### The process
+The trigger of remove memory dump will call a VM subresource that will add a removeMemorydumpRequest to the VM status.
+The request will unbound the PVC from the vm. The PVC will be deleted from the VM spec and the memoryDumpStatus will be removed.
+
+### Handle the memory dump
+After the memory dump is completed the user will be able to export this PVC and also if there is a snapshot supported storage class the PVC can be a part of the VM snapshot and then it can be unbounded and deleted.
+
+## Update/Rollback Compatibility
+New API should not affect updates / rollbacks.
+
+# Implementation Phases
+The implementation wil be split to several phases:
+* Add memory dump command in virt launcher server
+* Add virtctl command to execute memory-dump with existing pvc
+* Extend the virtctl command to support on-demand creation of PVC
+* Add remove memory dump virtctl command
--- a/design-proposals/vm-memory-dump.md
+++ b/design-proposals/vm-memory-dump.md
@@ -18,11 +18,11 @@ As a Kubevirt user I would like to get a memory dump of a running VM so that I c
 
 # Design
 To trigger a memory dump a virtctl memoryDump command will be added.
-The command will either use an existing PVC or create a new PVC on demand. This PVC will be bounded to the VM and will appear in the VM volumes spec with a fiting status.
+The command will either use an existing PVC or create a new PVC on demand. This PVC will be bound to the VM and will appear in the VM volumes spec with a fitting status.
 During the memory dump process this PVC will be mounted to the virt-launcher pod. After the mounting the guest memory will be dumped to that pvc and eventually it will be unmounted from the virt launcher.
-The PVC will remain bounded - in the VM spec. It is the users' responsibility to export/use the memory dump results before reusing the PVC for a new memory dump. As long as the PVC is bounded as a memory dump "container" each memory dump command will overwrite the previous content.
+The PVC will remain bound - in the VM spec. It is the users' responsibility to export/use the memory dump results before reusing the PVC for a new memory dump. As long as the PVC is bound as a memory dump "container" each memory dump command will overwrite the previous content.
 It will be possible to track the last memory dump in the VM status with a new memory dump status that will be added there.
-It will be possible to unbound the PVC and then do a memory dump to another PVC that will be bounded instead. 
+It will be possible to unbound the PVC and then do a memory dump to another PVC that will be bound instead.
 
 ## API
 
@@ -36,11 +36,44 @@ In this case `memory-pvc` should already exist and be of a size big enough to co
 It will be possible to ask for a PVC to be created with `--create` flag. In such case the required size will be calculated.
 
 ##### The process
-The trigger of memory dump will call a VM subresource that will add a memorydumpRequest to the VM status.
-The request will cause the binding of the PVC to the vm (if the pvc is not bounded yet) after that the pvc will be mounted to the virt-launcher that will trigger the guest memory dump `virDomainCoreDump` command with `VIR_DUMP_MEMORY_ONLY` flag to be executed by virt-launcher.
+The trigger of memory dump will call a VM subresource endpoint which will look like this:
+`/apis/subresources.kubevirt.io/v1alpha3/namespaces/<vm-namespace>/virtualmachines/<vm-name>/memorydump`
+
+The VM subresource endpoint will recieve a rest request containing the pvc name the user want to dump the memory to.
+It will patch the vm status with a memoryDumpStatus stating the volume-name and Phase `Binding`.
+The memoryDumpStatus will look something like this:
+
+`
+// MemoryDumpStatus represent the memory dump request status and info
+type MemoryDumpStatus struct {
+	// VolumeName is the name of the pvc that will contain the memory dump
+	VolumeName string `json:"volumeName"`
+	// Phase represents the memory dump phase
+	Phase MemoryDumpPhase `json:"phase,omitempty"`
+	// TimeStamp represents the time the memory dump was completed
+	TimeStamp *metav1.Time `json:"timestamp,omitempty"`
+}
+
+type MemoryDumpPhase string
+
+const (
+	// The memorydump is during pvc binding
+	MemoryDumpBinding MemoryDumpPhase = "Binding"
+	// The memorydump is in progress
+	MemoryDumpInProgress MemoryDumpPhase = "InProgress"
+	// The memorydump is completed
+	MemoryDumpCompleted MemoryDumpPhase = "Completed"
+	// The memorydump is being unbound
+	MemoryDumpUnBinding MemoryDumpPhase = "Unbinding"
+)
+`
+
+In the virt controller once getting the update of the memoryDump in state of binding it will bind the pvc to the vm by updating the vm volumes with the memoryDump pvc.
+After that the status of the memoryDump will be updated to `InProgress` which will be cause the volume to be mounted to the virt-launcher, same as done in the hotplug process, other then the end of the process that instead of attaching the volume to the VMI it will trigger the guest memory dump `virDomainCoreDump` command with `VIR_DUMP_MEMORY_ONLY` flag to be executed by virt-launcher.
 (Check out [virDomainCoreDump](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainCoreDump) for reference)
-After the dump is complete it will unmount the pvc from virt-launcher.
-The PVC will remain bounded to the VM - will be kept in the VM volumes and have an updated status and timestamp in the VM status.
+After the dump is complete the timestamp and status of the memory dump will be updated and it will unmount the pvc from virt-launcher.
+The PVC will remain bound to the VM - will be kept in the VM volumes and have the updated status and timestamp in the VM status as mentioned.
+
 
 #### remove memory dump from VM
 Run a virtctl command to remove a memory dump from a VM. This new API will look as follows:
@@ -49,10 +82,10 @@ Run a virtctl command to remove a memory dump from a VM. This new API will look 
 
 ##### The process
 The trigger of remove memory dump will call a VM subresource that will add a removeMemorydumpRequest to the VM status.
-The request will unbound the PVC from the vm. The PVC will be deleted from the VM spec and the memoryDumpStatus will be removed.
+The request will un the PVC from the vm. The PVC will be deleted from the VM spec and the memoryDumpStatus will be removed.
 
 ### Handle the memory dump
-After the memory dump is completed the user will be able to export this PVC and also if there is a snapshot supported storage class the PVC can be a part of the VM snapshot and then it can be unbounded and deleted.
+After the memory dump is completed the user will be able to export this PVC and also if there is a snapshot supported storage class the PVC can be a part of the VM snapshot and then it can be unbound and deleted.
 The output of the memory dump can be used for memory analysis with different tools for example [Volatility3](https://github.com/volatilityfoundation/volatility3) and maybe also [sleuthkit](https://www.sleuthkit.org/autopsy/)
 
 


design-proposals/vm-memory-snapshot.md
```diff
@@ -0,0 +1,151 @@
+# Overview
+
+Proposal for taking VM RAM snapshot which can be either a stand alone action, or alongside Online VMSnapshot.
```
dankenigsberg, line 3: Is `Online VMSnapshot` a known term? The capitalization looks a bit odd to me. Maybe you can add a link to where it is defined.

jobbler:
/lgtm
I had no issues understanding this proposal and process.

design-proposals/vm-memory-snapshot.md
ShellyKa13: hmm I guess not, it just seemed like to emphasize the action but I guess it is not proper :) ill update it

ShellyKa13:
@alicefr thanks! regarding your questions:
1) havent really dig deep into it, I checked how it is done in RHV, they have some formula to calculate it.. (https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.4.10.6/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/utils/VmOverheadCalculatorImpl.java#L135)
2) Maybe, I will check it before implementing to know if need to support attachment of block PVC too
3) It was considered. The idea is that probably if doing a one time analysis then you wont check for several memory dumps at the same time, and if it was done as part of the snapshot then it will also be snapshotted and saved as a volumesnapshot to be recovered so again doesnt seem to need more then one at the same time. But this is a proposal you are welcome to suggest otherwise :)

design-proposals/vm-memory-snapshot.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Goals
+Have a mechanism to take and get a VM memory snapshot.
+This mechanism can be either during the Online VMSnapshot process to get a complete snapshot of a VM: disks and Ram alike, or as a stand alone command to get the memory dump only.
+
+## Non Goals
+Currently the memory snapshot will not be used for VM hibernation, not in the VM restore process.
+
+## User Stories
+As a Kubeviert user I would like to take a snapshot of a running VM so that I can later inspect its disk and memory
```
jbpratt, line 16:
```suggestion
As a KubeVirt user I would like to take a snapshot of a running VM so that I can later inspect its disk and memory
```
spelling error

design-proposals/vm-memory-snapshot.md
```diff
@@ -0,0 +1,151 @@
...
+        name: cloudinitdisk
+```
+
+This can be added before creating the VM, or if the VM already exists need to add and restart the VM to make the pvc mounted.
+
+#### Add memory dump volume as a hotplug
+
+Use the hot-plug API to add a memory dump volume by adding a flag to the `virtctl addVolume` command:
+
+`$ virtctl addvolume my-vm --volume-name=memory-dump --memoryDumpVolume`
```
rmohr, line 95: I don't understand why this has to be tied to disks or volumes.

design-proposals/vm-memory-snapshot.md
rmohr, line 16: I would love to see more details stories outlined and how people are getting the dump.

In general, what would have been helpful in many cases would be a plain rest endpoint which allows streaming the dump to a local machine.

design-proposals/vm-memory-snapshot.md
mhenriks: The idea is that there is a "memory" volume type but the volume is not to be included in the list of disks since it should not be visible to the vm.  But could understand it not being a "volume" and just some other part of VM/VMI configuration.  If it is just a volume it may fit in easier with VirtualMachineExport proposal which exports volumes as http endpoints

But I think the PVC target for memory dumps should be configured somewhere on the VM/VMI.  Especially when you consider the hibernate use case.  Do you agree with that?

design-proposals/vm-memory-snapshot.md
mhenriks: The idea is to use this in concert with `VirtualMachineExport` proposal that exports all volumes as http endpoints

design-proposals/vm-memory-snapshot.md
```diff
@@ -0,0 +1,151 @@
...
+As mentioned the memory snapshot can be taken either in the Online VMSnapshot or as a stand alone command.
+To trigger the memory snapshot during the Online VMSnapshot one will add a parameter to the VMSnapshot yaml.
+To trigger a memory snapshot as a stand alone command a virtctl memoryDump command will be added.
+
+** Before doing either option a pvc should be mounted to the virt-launcher pod. A volume source type will be added to indicate that this is a volume without a matching disk since the VMI doesn't need to access the memory dump, only the virt launcher. This volume source will be able to be added to the VM spec as a volume or by using the Hot-plug API that will be extended to adding this volume without a disk.
+Only one memoryDump volume will be for the VM, the content will be overwritten in every memory dump call.
+
+## API
+
+### New volume source
```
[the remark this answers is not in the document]
mhenriks: I think memoryDump best makes most sense as an imperative operation.  You call the subresource API, the operation is executed as quickly as possible, and when the call returns you know whether it succeeded or not.  That all requires the target of the dump be mounted in the virt launcher pod.  Do you agree?

If the dump target is not mounted like the rest of the other PVC based volumes (static or hotplug).  How would it be configured?  A new resource that references the VM/VMI?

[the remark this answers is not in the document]
ShellyKa13: @rmohr In order for the dump to be included in the snapshot it made sense that the pvc that will contain the dump will be a part of the VM specifically some kind of volume, the pvc is attached to the virt-launcher pod like a volume it just not being mounted to the vmi as a disk also. Also for hibernation the pvc need to be a part of the vm even after it is stopped, so it seems like it should to be as part of the vm.
@mhenriks the operation will not return after the completion of the operation since from what I understand if it possible that the dump will be big and will take some time, so the plan is to send the command and then wait till the status of the dump volume will be updated to completed the operation.

design-proposals/vm-memory-snapshot.md
ShellyKa13: As I mentioned in the previous comment the dump needs to be a part of the VM cause it should be available also after VM shutdown so it cant only be a part of the VMI status for example.
It was considered to the the create of the pvc and binding as part of the dump command but since it should also happen in the snapshot it seems like it will add too much latency to the snapshot process.

design-proposals/vm-memory-snapshot.md
```diff
@@ -0,0 +1,151 @@
...
+    name: my-vm
+  includeMemory: true
+```
+
+#### How to handle it
+The trigger of memory dump will call a vmi subresource that will trigger `virDomainCoreDump` command with `VIR_DUMP_MEMORY_ONLY` flag to be executed by virt-launcher. Since this can take some time it will be an asynchronous operation. The memoryDump volume in the VMI-volume-status will change to `memoryDumpInProgress` and when completed it will change to `memoryReady` to indicate that the memory dump is ready to use.
+
+If the memory dump was triggered by memoryDump command then the user will be able to watch the progress by looking at the VMI memorydump volume status.
+
+In the Online VMSnapshot the memoryDump will be triggered after the vmfreeze. Taking into account that for a big memory dump it can take more time for the snapshot to be completed.
```
rthallisey, line 128: What is vmfreeze?  Is the VM paused? Or is it an fsfreeze?

design-proposals/vm-memory-snapshot.md
```diff
@@ -0,0 +1,151 @@
...
+## Update/Rollback Compatibility
+New API should not affect updates / rollbacks.
+
+# Implementation Phases
+Will split the design to several phases:
+* Add new volume source type that is only mounted to virt-launcher
+* Update hotplug API to be able to hotplug memoryDump volume.
+* Add subresource which dump the VM memory to a known path in virt-launcher - i.e the pvc mount point.
+* Add update of memoryDump volume status according to the progress of the memory dump to be able to know when it is done.
+* Add virtctl command to execute getMemoryDump
```
rthallisey, line 142: What's the end goal for the memory?  We now have it in a volume, now what?

design-proposals/vm-memory-snapshot.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Get memory dump
+#### trigger memory dump
+After adding the pvc to which the memory dump should be saved to, the user can trigger the memory dump in two ways:
+
+1. Run a virtctl command to get a memory dump. This new API will look as follows:
+
+`$ virtctl memoryDump my-vm`
+
+2. Get a memory dump during Online VMSnapshot. This will be done by adding to the VMSnapshot yaml a parameter to indicate that we also want a memory dump. For example:
```
rthallisey, line 108: Do we need to account for devices?

davidvossel (commented):
If we're moving in the direction of capturing VM memory, I'd like for us to gain an understanding of how such a feature might eventually work into a suspend/resume sort of feature similar to this, [1]

As a user, this suspend/resume like functionality is interesting because it allows someone to preserve a running VM while releasing the cpu/mem resources associated with it. I could for example pre-warm a VM at a specific spot in the execution, then restore a pre-loaded VM. This is a feature that has been discussed for years within KubeVirt, but never formally documented.

We certainly don't have to implement suspend/resume right away, i'd just like for us to explore (and know) how this VM memory snapshot fits into that use case... and if it doesn't fit into that use case then we might want to re-evaluate what we're doing.

For example, what would be a future workflow for someone wanting to suspend/resume a VM using this feature? 

1. https://cloud.google.com/compute/docs/instances/suspend-resume-instance

design-proposals/vm-memory-snapshot.md
rmohr: You can also have stuff in the VM status. But I really don't see how adding "special" volumes is in any way convenient for users and consistent with the spec, and I also don't see the need for it without further explanations.


I think you can try to tackle this from a few different angles.

Firs, if a memory dump is really done on the VM/VMI it is very likely not a declarative action. Adding special volumes does not make it more declarative. So you anyway already have an imperative trigger to do the actual dump.

I then see multiple ways, depending on how much work you want to happen on the VM or the VMI. Like, if you do an imperative /dump call on the VM, you could in that call either specify a PVC, or let the VM controller just create a PVC and record the procgress in the status.

The VM controller could then trigger on the VMI level a /dump call where it insructs the VMI to dump into that PVC. Everything can be visibly tracked on the VM and VMI status and the latest snapshot can be found any time from the VM status.

This is of course just one possibility. What do you think?

design-proposals/vm-memory-snapshot.md
ShellyKa13: is it fsfreeze. for the dump command I understand that the vm will be suspended and after unsuspended hence it will take care for the consistency of the dump and the wrapping fsfreeze will take care for the consistency of the dump with the disks.

design-proposals/vm-memory-snapshot.md
ShellyKa13: for now need to find a manual way of getting the dump and analyzing it with debugging tools. in the future there should be able to export it as a volume from https point as Michael mentioned. or have it as part of the vm snapshot and restore from it. this proposal is a tech preview only for the memory dump part.

design-proposals/vm-memory-snapshot.md
ShellyKa13: not sure what you mean

design-proposals/vm-memory-snapshot.md
awels: We looked at doing something similar to disk hotplug and the trigger for virt-handler to add a volume to the virt-launcher pod is a new volume being added to the VMI object. This is where the idea of a non disk volume comes from. We basically need a spot to deposit the memory dump and a volume seems like a natural spot. Now the mechanism which we use to add the volume is up for debate, we just started with something similar to hotplug.

rmohr: If I am not mistaken, when we do hot plug we add a volume because we want a volume, and when we want it gone we remove the volume. This is declarative and we are not adding a pseudo volume from an architectural perspective. Or maybe I misunderstood what you meant?

I think I laid out a few other possibilities, did you also look into something like them?

awels: I suggested initially to do the entire process as part of a single command. Mount the volume, do the memory dump, unmount the volume. Which is a prescriptive operation. But if we think of the add on functionality of taking a memory dump together with a VM snapshot, or a hibernate type scenario, those are both declarative operations, and having a prescriptive operation be part of those declarative operations might not be the best idea. Which is why are trying to see if it makes sense (or is possible) for this to be a declarative operation.

rmohr: VM snapshot can definitely be done in a declarative state from a VMSnapshot object point of view. I would not do a live snapshot on a VM in a declarative way. I don't even think that is possible.

For hibernate, yes hibernate can be expressed declarative on the VM, and it can use in the background common flows with a "debug" memory dump.

design-proposals/vm-memory-snapshot.md
rthallisey: If a mediated device is attached to the VM and we take an Online Snapshot, do we get the device's memory?

design-proposals/vm-memory-snapshot.md
rthallisey: I think we're are aligned, but can you add the expectation that the VM is suspended/paused to the doc?

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,64 @@
...
+
+## API
+
+### Get memory dump
+#### trigger memory dump
+Run a virtctl command to get a memory dump. This new API will look as follows:
+
+`$ virtctl memoryDump my-vm --volume-name=memory-pvc`
+
+In this case `memory-pvc` should already exist and be of a size big enough the contain the memory dump. A check will be made to make sure.
```
jbpratt, line 35:
```suggestion
In this case `memory-pvc` should already exist and be of a size big enough to contain the memory dump. A check will be made to make sure.
```

jbpratt (commented):
:cat: :+1:

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,64 @@
...
+It will be possible to track the last memory dump in the VM status with a new memory dump status that will be added there.
+It will be possible to unbound the PVC and then do a memory dump to another PVC that will be bounded instead. 
+
+## API
+
+### Get memory dump
+#### trigger memory dump
+Run a virtctl command to get a memory dump. This new API will look as follows:
+
+`$ virtctl memoryDump my-vm --volume-name=memory-pvc`
```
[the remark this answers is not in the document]
jbpratt: there are a few more than `migrate-cancel` such as `soft-reboot`, `port-forward`, `permitted-devices`, and `image-upload`

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,64 @@
...
+
+In this case `memory-pvc` should already exist and be of a size big enough the contain the memory dump. A check will be made to make sure.
+It will be possible to ask for a PVC to be created with `--create` flag. In such case the required size will be calculated.
```
alicefr, line 36: Which storage class do you plan to use if multiple are defined? The default one? Or should this be configurable? 
Can we also use block-based PVCs or you should also check that the PVC is filesystem mode?

design-proposals/vm-memory-dump.md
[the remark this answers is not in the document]
ShellyKa13: weird I edited my comment to say that there are several from each format and it didn't edit it.

design-proposals/vm-memory-dump.md
ShellyKa13: I think the logic should be similar to the one done in image-upload in terms of the creation of the pvc from what I recall the default storage class is used.
Regarding block I did test doing a dump to a block pvc the dump succeeded so it is possible Im just not sure how the pvc will be handled after the dump in regards of using the output written on the block device. I think as a first phase I will do filesystem only. Ill mention it in the proposal.

iholder-redhat:
Maybe it's just me - but I don't fully understand what does a memory dump mean exactly or how it is represented.

Is it represented by text? binary? Is the something standard format which I'm not aware of? How would the data be analyzed? For which cases is it most valuable?

Again, might be me lacking knowledge here.

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,64 @@
...
+#### remove memory dump from VM
+Run a virtctl command to remove a memory dump from a VM. This new API will look as follows:
+
+`$ virtctl removeMemoryDump my-vm`
+
+##### The process
+The trigger of remove memory dump will call a VM subresource that will add a removeMemorydumpRequest to the VM status.
+The request will unbound the PVC from the vm. The PVC will be deleted from the VM spec and the memoryDumpStatus will be removed.
+
+### Handle the memory dump
```
dankenigsberg, line 53: Can you add a URL to tools that may be used to debug the dump? I think it would help readers who may not know much about what dumps are and what you do with them.

/cc @iholder-redhat

ShellyKa13: yeah sure I'll add some links

ShellyKa13: @iholder-redhat @dankenigsberg added a link to the libvirt command that will be used and a link to two tools that can be used for debugging. Hope it's enough

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,68 @@
...
+## Update/Rollback Compatibility
+New API should not affect updates / rollbacks.
+
+# Implementation Phases
+The implementation wil be split to several phases:
+* Add memory dump command in virt launcher server
+* Add virtctl command to execute memory-dump with existing pvc
+* Extend the virtctl command to support on-demand creation of PVC
+* Add remove memory dump virtctl command
+* Support block pvc for the dump (?)
```
mhenriks, line 68: Can't imagine that being possible

ShellyKa13: as I responded to @alicefr I did test it manually, attaching a block pvc hard coded to virt launcher and did virsh dump to the block device and it completed successfully... I havent tried to copy it or doing with it anything so Im not sure if the result is viable lol

mhenriks: Yeah, not sure how you would know and/or track the size of the dump on the block device in order to copy it out or make the dump usable

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,68 @@
...
+#### trigger memory dump
+Run a virtctl command to get a memory dump. This new API will look as follows:
+
+`$ virtctl memory-dump my-vm --volume-name=memory-pvc`
+
+In this case `memory-pvc` should already exist and be of a size big enough to contain the memory dump. A check will be made to make sure.
+It will be possible to ask for a PVC to be created with `--create` flag. In such case the required size will be calculated.
+
+##### The process
+The trigger of memory dump will call a VM subresource that will add a memorydumpRequest to the VM status.
```
mhenriks, line 39: Example of the subresource api being called and what the VM status would look like will be helpful

design-proposals/vm-memory-dump.md
ShellyKa13: valid point

design-proposals/vm-memory-dump.md
ShellyKa13: @mhenriks added a new commit with more details about the subresource api. please have a look

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,99 @@
...
+Run a virtctl command to get a memory dump. This new API will look as follows:
+
+`$ virtctl memory-dump my-vm --volume-name=memory-pvc`
+
+In this case `memory-pvc` should already exist and be of a size big enough to contain the memory dump. A check will be made to make sure.
+It will be possible to ask for a PVC to be created with `--create` flag. In such case the required size will be calculated.
+
+##### The process
+The trigger of memory dump will call a VM subresource.
+The VM subresource endpoint will recieve a request containing the pvc name the user want to dump the memory to.
```
arnongilboa, line 40: Can you please extend on the VM subresource ep?

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,99 @@
...
+It will patch the vm status with a memoryDumpStatus stating the volume name and Phase Binding.
+The memoryDumpStatus will look something like this:
+
+`
+// MemoryDumpStatus represent the memory dump request status and info
+type MemoryDumpStatus struct {
+	// VolumeName is the name of the pvc that will contain the memory dump
+	VolumeName string `json:"volumeName"`
```
arnongilboa, line 48: No namespace needed?

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,99 @@
...
+	MemoryDumpInProgress MemoryDumpPhase = "InProgress"
+	// The memorydump is completed
+	MemoryDumpCompleted MemoryDumpPhase = "Completed"
+	// The memorydump is being unbound
+	MemoryDumpUnBinding MemoryDumpPhase = "Unbinding"
+)
+`
+
+In the virt controller once getting the update of the memoryDump in state of ing it will bind the pvc to the vm by updating the vm volumes with the memoryDump pvc.
+After that the status of the memoryDump will be updated to in progress which will be cause the volume to be mounted to the virt-launcher, same as done in the hotplug process, other then the end of the process that instead of attaching the volume to the VMI it will trigger the guest memory dump `virDomainCoreDump` command with `VIR_DUMP_MEMORY_ONLY` flag to be executed by virt-launcher.
```
arnongilboa, line 70: in progress = `InProgress`

design-proposals/vm-memory-dump.md
ShellyKa13: what more should be extended on? I explained about what it will update and trigger what is missing?

design-proposals/vm-memory-dump.md
ShellyKa13: this struct is not yet set in stone, it is possible to add to it namespace or at pointer to the pvc. doesn't affect the idea of implementation.

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,99 @@
...
+	// The memorydump is in progress
+	MemoryDumpInProgress MemoryDumpPhase = "InProgress"
+	// The memorydump is completed
+	MemoryDumpCompleted MemoryDumpPhase = "Completed"
+	// The memorydump is being unbound
+	MemoryDumpUnBinding MemoryDumpPhase = "Unbinding"
+)
+`
+
+In the virt controller once getting the update of the memoryDump in state of ing it will bind the pvc to the vm by updating the vm volumes with the memoryDump pvc.
```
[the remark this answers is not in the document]
ShellyKa13: yep, I did some vim replacing tricks and I guess it skipped an instance :)

design-proposals/vm-memory-dump.md
arnongilboa: simply how it looks in the VM CR status perspective

ShellyKa13: Not sure I understand the endpoint is not shown in the VM CR, its a a path with an ending /memorydump that redirects the request and eventually updates the VM CR status with the MemoryDumpStatus with values of volume name and binding status which I wrote about and shown the struct

arnongilboa: Ok, that explanation made it much clearer to me, so consider adding it.

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,101 @@
...
+	// The memorydump is being unbound
+	MemoryDumpUnBinding MemoryDumpPhase = "Unbinding"
+)
+`
+
+In the virt controller once getting the update of the memoryDump in state of binding it will bind the pvc to the vm by updating the vm volumes with the memoryDump pvc.
+After that the status of the memoryDump will be updated to `InProgress` which will be cause the volume to be mounted to the virt-launcher, same as done in the hotplug process, other then the end of the process that instead of attaching the volume to the VMI it will trigger the guest memory dump `virDomainCoreDump` command with `VIR_DUMP_MEMORY_ONLY` flag to be executed by virt-launcher.
+(Check out [virDomainCoreDump](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainCoreDump) for reference)
+After the dump is complete the timestamp and status of the memory dump will be updated and it will unmount the pvc from virt-launcher.
+The PVC will remain bound to the VM - will be kept in the VM volumes and have the updated status and timestamp in the VM status as mentioned.
```
mhenriks, line 75: What does this look like in the volumes list?

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,101 @@
...
+It will be possible to track the last memory dump in the VM status with a new memory dump status that will be added there.
+It will be possible to unbound the PVC and then do a memory dump to another PVC that will be bound instead.
+
+## API
+
+### Get memory dump
+#### trigger memory dump
+Run a virtctl command to get a memory dump. This new API will look as follows:
+
+`$ virtctl memory-dump my-vm --volume-name=memory-pvc`
```
mhenriks, line 33: `--volume-name` optional if already in the volumes list or no?

ShellyKa13: yes maybe, I'll see when implementing if it works out

design-proposals/vm-memory-dump.md
ShellyKa13: @mhenriks add a commit

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,147 @@
...
+	MemoryDumpCompleted MemoryDumpPhase = "Completed"
+	// The memorydump is being unbound
+	MemoryDumpUnBinding MemoryDumpPhase = "Unbinding"
+)
+`
+
+In the virt controller once getting the update of the memoryDump in state of binding it will bind the pvc to the vm by updating the vm volumes with the memoryDump pvc.
+After that the status of the memoryDump will be updated to `InProgress` which will be cause the volume to be mounted to the virt-launcher, same as done in the hotplug process, other then the end of the process that instead of attaching the volume to the VMI it will trigger the guest memory dump `virDomainCoreDump` command with `VIR_DUMP_MEMORY_ONLY` flag to be executed by virt-launcher.
+(Check out [virDomainCoreDump](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainCoreDump) for reference)
+After the dump is complete the timestamp and status of the memory dump will be updated and it will unmount the pvc from virt-launcher.
```
awels, line 74: The unmount has to happen in the virt-handler pod, so we need a mechanism to communicate that we are complete from the virt-launcher to the virt-controller, so the virt-controller can then update the status which in turn triggers the virt-handler to unmount the volume.

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,147 @@
...
+	TimeStamp *metav1.Time `json:"timestamp,omitempty"`
+}
+
+type MemoryDumpPhase string
+
+const (
+	// The memorydump is during pvc binding
+	MemoryDumpBinding MemoryDumpPhase = "Binding"
+	// The memorydump is in progress
+	MemoryDumpInProgress MemoryDumpPhase = "InProgress"
```
awels, line 63: I would probably add 3 more phases here to indicate that we are mounting to the volume in the virt-launcher pod and that we are unmounting it. The third is a dump to volume complete, which can trigger the virt-handler to start the unmounting.

design-proposals/vm-memory-dump.md
awels: Well technically you can just dd the entire block device out to a file and use that, but I would stick with filesystem.

design-proposals/vm-memory-dump.md
```diff
@@ -73,6 +73,52 @@ After that the status of the memoryDump will be updated to `InProgress` which wi
...
+            name: my-dv
+          name: datavolumedisk
+        - cloudInitNoCloud:
+            userData: |-
+              #cloud-config
+              password: fedora
+              chpasswd: { expire: False }
+          name: cloudinitdisk
+        - memoryDump:
+            persistentVolumeClaim:
```
mhenriks, line 110: Will there possibly be other other than a persistentVolumeClaim?  If not, then maybe just `memoryDump.claimName`?

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,147 @@
...
+                claimName: memory-dump-pvc
+          name: memory-dump
+  status:
+    ....
+    Snip
+    ....
+    memoryDumpStatus:
+      volumeName: memory-dump
```
mhenriks, line 118: should this be `claimName`?

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,147 @@
...
+          name: cloudinitdisk
+        - memoryDump:
+            persistentVolumeClaim:
+                claimName: memory-dump-pvc
+          name: memory-dump
+  status:
+    ....
+    Snip
+    ....
+    memoryDumpStatus:
```
mhenriks, line 117: This is the last memory dump then?

design-proposals/vm-memory-dump.md
mhenriks: What is the purpose of including the memory dump pvc in the list of volumes?  I can understand it being there while a dump is in progress, but whey after?

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,147 @@
...
+	MemoryDumpInProgress MemoryDumpPhase = "InProgress"
+	// The memorydump is completed
+	MemoryDumpCompleted MemoryDumpPhase = "Completed"
+	// The memorydump is being unbound
+	MemoryDumpUnBinding MemoryDumpPhase = "Unbinding"
+)
+`
+
+In the virt controller once getting the update of the memoryDump in state of binding it will bind the pvc to the vm by updating the vm volumes with the memoryDump pvc.
+After that the status of the memoryDump will be updated to `InProgress` which will be cause the volume to be mounted to the virt-launcher, same as done in the hotplug process, other then the end of the process that instead of attaching the volume to the VMI it will trigger the guest memory dump `virDomainCoreDump` command with `VIR_DUMP_MEMORY_ONLY` flag to be executed by virt-launcher.
```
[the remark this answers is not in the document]
ShellyKa13: right :), I didnt want to get in so many details in the design

design-proposals/vm-memory-dump.md
ShellyKa13: You can look at the PR I created for the memory dump, the action in of the memorydump triggers an event on the domain which updates the timestamp and completion of the memorydump. in a future PR this update of state in the volume status in the vmi will triger the update of that status on the vm status which will then remove the memorydump from the vmi which will cause the unmount

design-proposals/vm-memory-dump.md
ShellyKa13: I think it will only be pvc, so maybe the name is enough. will be decided in the implementation

design-proposals/vm-memory-dump.md
ShellyKa13: sure

design-proposals/vm-memory-dump.md
ShellyKa13: yep I mentioned it in the design, there will be information on the last dump, the user will be responsible to handle it before taking another dump that will ovewrite the previous if its on the same pvc

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,147 @@
...
+	// The memorydump is in progress
+	MemoryDumpInProgress MemoryDumpPhase = "InProgress"
+	// The memorydump is completed
+	MemoryDumpCompleted MemoryDumpPhase = "Completed"
+	// The memorydump is being unbound
+	MemoryDumpUnBinding MemoryDumpPhase = "Unbinding"
+)
+`
+
+In the virt controller once getting the update of the memoryDump in state of binding it will bind the pvc to the vm by updating the vm volumes with the memoryDump pvc.
```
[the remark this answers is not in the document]
ShellyKa13: not exactly, Im not updating any ownership or annotation on the PVC, its just adding it to the vm spec volumes and memorydumpstatus which mentions that this is the pvc for the memorydump and it will remain as such until removing it - unbinding it.. if you have a better word for it I'm open..

design-proposals/vm-memory-dump.md
ShellyKa13: hmmm, the idea is that if it is in the vm spec it can be a part of the vmsnapshot since it is counted as part of the volumes. If you think we can just handle it specially in the vmsnapshot and your export doesnt also require it to be part of the vm volumes list then maybe in the vm spec volumes list we dont need to update it there at all and the memorydumpstatus is enough

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,147 @@
...
+
+## User Stories
+As a Kubevirt user I would like to get a memory dump of a running VM so that I can later inspect its memory.
+
+## Repos
+[KubeVirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+To trigger a memory dump a virtctl memoryDump command will be added.
+The command will either use an existing PVC or create a new PVC on demand. This PVC will be bound to the VM and will appear in the VM volumes spec with a fitting status.
```
duyanyan, line 21: If memory dump to an existing PVC, we need create the PVC before starting the dump, right? if yes, how to know the size for the PVC that will be enough for dump?

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,147 @@
...
+      volumeName: memory-dump
+      phase: Completed
+      timestamp: "2022-03-29T11:00:04Z"
+```
+
+
+#### remove memory dump from VM
+Run a virtctl command to remove a memory dump from a VM. This new API will look as follows:
+
+`$ virtctl remove-memory-dump my-vm`
```
duyanyan, line 127: Do we need to specify the --volume-name para when removing the memory dump?

ShellyKa13: no since there is only one memory dump pvc at a given time so it is known which should be removed

design-proposals/vm-memory-dump.md
ShellyKa13: right, there is a formula to calculate the size of the memory used by a vm (example of the calculation done in oVirt: https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.4.10.6/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/utils/VmOverheadCalculatorImpl.java#L135)

design-proposals/vm-memory-dump.md
```diff
@@ -0,0 +1,147 @@
+# Overview
+Proposal for taking VM memory dump for analysis purposes.
+
+## Motivation
+One of the methods of troubleshooting guest virtual machines is getting a memory dump and analyzing it. As this is a standard capability in most VM management systems, it is also required for kubevirt’s VMs.
+
+## Goals
+Have a mechanism to get a VM memory dump.
```
fabiand, line 8: A memory dump including all disks attached to the VM is needed for some forensic tools IIUIC.

design-proposals/vm-memory-dump.md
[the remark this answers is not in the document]
awels: Maybe associating? When I see binding, I imagine a relationship like between a PVC and PV

design-proposals/vm-memory-dump.md
fabiand, line 8: Let's include the specific forensics tools to the goals section.

design-proposals/vm-memory-dump.md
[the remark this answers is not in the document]
awels: Okay

design-proposals/vm-memory-dump.md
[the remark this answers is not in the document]
ShellyKa13: yeah Adam also used associate to describe this relationship, but it sounded weird to me a phase of associating

design-proposals/vm-memory-dump.md
ShellyKa13: as far as I understand adding the disks to the dump is not the goal, the tools I think we are focused on is volatility3 which doesnt need the disks it uses the memory dump as is

design-proposals/vm-memory-dump.md
mhenriks: Okay this is fine, but the snapshot will not generate a dump (at least initially), right?

ShellyKa13: no it will not (not initially at least)

davidvossel (commented):
/approve

makes sense to me!

One thing to note about the implementation. When virDomainCoreDump is called by the virt-handler reconcile loop, We likely need to make sure that's an asynchronous call, meaning that it doesn't block the handler reconcile thread waiting for the dump to write to disk. 

This could be accomplished by executing virDomainCoreDump in a goroutine withing virt-launcher, then reporting back the status of the dump on the DomainXML's metadata (see how live migration status is returned on domainxml) which will trigger the VMI's reconcile loop again in hander.

so something like this in handler
- reconcile 1, invoke snapshot
- virt-launcher gorouting to execute virDomainCoreDump
- reconcile 1 complets
- virDomainCoreDump completes, write info th domain XML related to mem dump
- reconcile 2 caused by new domainxml change
- reconcile 2 sees core dump completes and writes to VMI status

PR #165, opened 2022-03-02
design-proposals: Refactor and extend the Flavors API 

lyarwood:
KubeVirt's VirtualMachine API contains many advanced options for tuning
a virtual machine performance that goes beyond what typical users need
to be aware of. Users are unable to simply define the storage/network
they want assigned to their VM and then declare in broad terms what
quality of resources and kind of performance they need for their VM.

This design proposal sets of various changes to existing
VirtualMachineFlavor CRDs and the introduction of a new set of
VirtualMachinePreference CRDs to hopefully simplify the choices a user
has to make when defining a VirtualMachine to get the resources,
performance and guestOS characteristics they desire.

### DIFF
--- a/design-proposals/flavors-and-preferences.md
+++ b/design-proposals/flavors-and-preferences.md
@@ -1,0 +1,1059 @@
+# Overview
+
+A common pattern for IaaS is to have abstractions separating the resource sizing and performance of a workload from the user defined values related to launching their custom application. This pattern is evident across all the major cloud providers (also known as hyperscalers) as well as open source IaaS projects like OpenStack. AWS has [instance types](https://aws.amazon.com/ec2/instance-types/), GCP has [machine types](https://cloud.google.com/compute/docs/machine-types#custom_machine_types), Azure has [instance VM sizes](https://docs.microsoft.com/en-us/azure/virtual-machines/sizes) and OpenStack has [flavors](https://docs.openstack.org/nova/latest/user/flavors.html).
+
+Let’s take AWS for example to help visualize what this abstraction enables. Launching an EC2 instance only requires a few top level arguments, the disk image, instance type, keypair, security group, and subnet: 
+
+```bash
+$ aws ec2 run-instances --image-id ami-xxxxxxxx \
+                        --count 1 \
+                        --instance-type c4.xlarge \
+                        --key-name MyKeyPair \
+                        --security-group-ids sg-903004f8 \
+                        --subnet-id subnet-6e7f829e
+```
+
+When creating the EC2 instance the user doesn't define the amount of resources, what processor to use, how to optimize the performance of the instance, or what hardware to schedule the instance on. Instead all of that information is wrapped up in that single `--instance-type c4.xlarge` CLI argument. `c4` denoting a specific performance profile version, in this case from the `Compute Optimized` family and `xlarge` denoting a specific amount of compute resources provided by the instance type, in this case 4 vCPUs,	7.5 GiB of RAM, 750 Mbps EBS bandwidth etc.
+
+While hyperscalers can provide predefined types with performance profiles and compute resources already assigned IaaS and virtualization projects such as OpenStack and KubeVirt can only provide the raw abstractions for operators, admins and even vendors to then create instances of these abstractions specific to each deployment.
+
+## Problem Statement
+
+KubeVirt's [`VirtualMachine`](https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachine) API contains many advanced options for tuning a virtual machine performance that goes beyond what typical users need to be aware of. Users are unable to simply define the storage/network they want assigned to their VM and then declare in broad terms what quality of resources and kind of performance they need for their VM.
+
+Instead, the user has to be keenly aware how to request specific compute resources alongside all of the performance tunings available on the [`VirtualMachine`](https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachine) API and how those tunings impact their guest’s operating system in order to get a desired result.
+
+The [partially implemented currently v1alpha1 `Virtual Machine Flavors` API](https://github.com/kubevirt/kubevirt/commit/6dc548459fa17d7f9601cbc251088d2b70a2a96a) was an attempt to provide operators and users with a mechanism to define resource buckets that could be used during VM creation. At present this implementation provides a cluster-wide [`VirtualMachineClusterFlavor`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineclusterflavor) and a namespaced [`VirtualMachineFlavor`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineflavor) CRDs. Each containing an array of [`VirtualMachineFlavorProfile`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineflavorprofile) that at present only encapsulates CPU resources by applying a full copy of the [`CPU`](http://kubevirt.io/api-reference/main/definitions.html#_v1_cpu) type to the [`VirtualMachineInstance`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachineinstance) at runtime.
+
+This approach has a few pitfalls such as:
+
+* Using embedded profiles within the CRDs
+
+* Tight coupling between the resource definition of a `VirtualMachineFlavorProfile` and workload preferences it might also contain that are required to run a specific `VirtualMachine`.
+
+* This results in a reliance on the user selecting the correct [`VirtualMachineFlavorProfile`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineflavorprofile) for their specific workload.
+
+* Not allowing a user to override some viable `VirtualMachineInstanceSpec` attributes at runtime
+
+## User Stories
+
+* As an operator, user or vendor I would like to provide flavors that define the resources available to a VM along with any schedulable *or* performance related attributes.
+
+* As an operator, user or vendor I would like to separately provide preferences that relate to ensuring that a specific guest OS within the VM is able to run correctly.
+
+* As an user I would like to be presented with and select from pre-defined options that encapsulate the resources, performance and any additional preferences required to run my workload.
+
+## Goals
+
+* Simplify the choices a user has to make in order to get a runnable `VirtualMachine` with the required amount of resources, desired performance and additional preferences to correctly run a given workload.
+
+## Non-Goals
+
+The following items have been discussed alongside `Virtual Machine Flavor` and `Virtual Machine Preferences` but are not within the current scope of this design proposal:
+
+* Introspection of imported images to determine the correct guest OS related `VirtualMachinePreferences` to apply.
+
+* Using image labels to determine the correct guest OS related `VirtualMachinePreferences` to apply.
+
+* Remove the need to define [`Disks`](http://kubevirt.io/api-reference/main/definitions.html#_v1_disk) within [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec) when providing [`Volumes`](http://kubevirt.io/api-reference/main/definitions.html#_v1_volume) within a [`VirtualMachineInstanceSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachineinstancespec).
+
+* Remove the need to define [`Interfaces`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachineinstancenetworkinterface) within [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec) when providing [`Networks`](http://kubevirt.io/api-reference/main/definitions.html#_v1_network) within a [`VirtualMachineInstanceSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachineinstancespec).
+
+* Versioning of these CRDs through [Controller Revisions](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#controllerrevision-v1-apps) to ensure the generated [`VirtualMachineInstanceSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachineinstancespec) of a `VMI` doesn't change over time if the CRDs are modified. This will be documented in a separate design proposal.
+
+## Repos
+
+* [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+
+As stated above the ultimate goal of this work is to provide the end user with a simple set of choices when defining a `VirtualMachine` within KubeVirt. To achieve this the existing [`VirtualMachineFlavor`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineflavor) CRDs will be heavily modified and extended to better encapsulate resource, performance or schedulable attributes of a VM.
+
+A quick note on the original `Flavor` implementation before we continue. At present the [`FlavorMatcher`](http://kubevirt.io/api-reference/main/definitions.html#_v1_flavormatcher) type is used lookup a given flavor when admitting the initial `VirtualMachine` resource or when later starting the `VirtualMachine`. During the latter the flavor is applied to the associated `VirtualMachineInstance` object by the `VM controller` before the actual request to create the `VirtualMachineInstance` resource is submitted to the API.
+
+For now this aspect of the design will stay but could be changed in future to allow the `Flavor` to be applied to the `VirtualMachineInstance` *after* submission to the API. This would allow users creating `VirtualMachine` and `VirtualMachineInstance` resources to have the same overall experience while ensuring things are expanded before any underlying `VM` launches.
+
+For now however this proposal will start with the removal of the embedded `VirtualMachineFlavorProfile` type within the CRDs, this will be replaced with a singular `VirtualMachineFlavorSpec` type per flavor. The decision to remove `VirtualMachineFlavorProfile` has been made as the concept isn't prevalent within the wider Kubernetes ecosystem and could be confusing to end users. Instead users looking to avoid duplication when defining flavors will be directed to use tools such as [`kustomize`](https://kustomize.io/) to generate their flavors. This tooling is already commonly used when defining resources within Kubernetes and should afford users plenty of flexibility when defining their flavors either statically or as part of a larger GitOps based workflow.
+
+`VirtualMachineFlavorSpec` will also include elements of `CPU`, `Devices`, `HostDevices`, `GPUs`, `Memory` and `LaunchSecurity` defined fully below. Users will be unable to override any aspect of the flavor (for example, `vCPU` count or amount of `Memory`) within the `VirtualMachine` itself, any attempt to do so resulting in the `VirtualMachine` being rejected.
+
+A new set of `VirtualMachinePreference` CRDs will then be introduced to define any remaining attributes related to ensuring the selected guestOS can run. As the name suggests the `VirtualMachinePreference` CRDs will only define preferences, so unlike a flavor if a preference conflicts with something user defined within the `VirtualMachine` it will be ignored. For example, if a user selects a `VirtualMachinePreference` that requests a `preferredDiskBus` of `virtio` but then sets a disk bus of `SATA` for one or more disk devices within the `VirtualMachine` the supplied `preferredDiskBus` preference will not be applied to these disks. Any remaining disks that do not have a disk bus defined will however use the `preferredDiskBus` preference of `virtio`.
+
+## Implementation Phases
+
+* Remove [`VirtualMachineFlavorProfile`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineflavorprofile) from [`VirtualMachineFlavor`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineflavor) and [`VirtualMachineClusterFlavor`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineclusterflavor).
+
+* Introduce a new `Spec` attribute and new `VirtualMachineFlavorSpec` type  itself with `CPU`, `Memory`, `GPUs` and `HostDevices` attributes to [`VirtualMachineFlavor`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineflavor) and [`VirtualMachineClusterFlavor`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineclusterflavor).
+
+* Introduce new `VirtualMachinePreference` and `VirtualMachineClusterPreference` CRDs to the Flavors API and `VirtualMachinePreferenceSpec` type to define guestOS related preferences such as preferred device buses and models.
+
+* Allow for `CPU` preferred topology policy to be defined within `VirtualMachinePreferenceSpec` through a `PreferredTopology` attribute.
+
+* Extend [`VirtualMachineSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachinespec) to allow an optional reference to `VirtualMachinePreference` or `VirtualMachineClusterPreference`.
+
+## [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec) mappings to `VirtualMachinePreferenceSpec` and `VirtualMachineFlavorSpec`
+
+The following basic rules are applied when deciding which attributes of the [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec) will be mapped to the new `VirtualMachinePreferenceSpec` and `VirtualMachineFlavorSpec` types:
+
+* Resource, schedulable and performance related attributes of the [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec) reside within `VirtualMachineFlavorSpec`
+
+* Guest OS related preferences reside within `VirtualMachinePreferenceSpec`
+
+For example, the number of and topology of vCPUs presented to the guest are defined within `VirtualMachineFlavorSpec` but the preferred topology of these vCPUs resides within  `VirtualMachinePreferenceSpec`.
+
+* [DomainSpec](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec)
+  * [CPU](http://kubevirt.io/api-reference/main/definitions.html#_v1_cpu)
+    * cores 
+      * `VirtualMachineFlavorSpec.CPU.Guest` 
+        * Number of vCPUs to expose to the guest
+      * `VirtualMachinePreferenceSpec.CPU.PreferredTopology`
+        * Guest visible CPU topology to expose, `preferSockets` , `preferThreads` or `preferCores` (default).
+    * sockets
+      * `VirtualMachineFlavorSpec.CPU.Guest` 
+        * Number of vCPUs to expose to the guest
+      * `VirtualMachinePreferenceSpec.CPU.PreferredTopology`
+        * Guest visible CPU topology to expose, `preferSockets`, `preferThreads`  or `preferCores` (default).
+    * threads
+      * `VirtualMachineFlavorSpec.CPU.Guest` 
+        * Number of vCPUs to expose to the guest
+      * `VirtualMachinePreferenceSpec.CPU.PreferredTopology`
+        * Guest visible CPU topology to expose, `preferSockets`, `preferThreads`  or `preferCores` (default).
+    * dedicatedCpuPlacement
+      * `VirtualMachineFlavorSpec.CPU.DedicatedCpuPlacement`
+    * [features](http://kubevirt.io/api-reference/main/definitions.html#_v1_cpufeature)
+      * `VirtualMachineFlavorSpec.CPU.Features`
+    * isolateEmulatorThread
+      * `VirtualMachineFlavorSpec.CPU.IsolateEmulatorThread`
+    * model
+      * `VirtualMachineFlavorSpec.CPU.Model`
+    * [NUMA](http://kubevirt.io/api-reference/main/definitions.html#_v1_watchdog)
+      * `VirtualMachineFlavorSpec.CPU.NUMA`
+    * [Realtime](http://kubevirt.io/api-reference/main/definitions.html#_v1_realtime)
+      * `VirtualMachineFlavorSpec.CPU.Realtime`
+  * [Devices](http://kubevirt.io/api-reference/main/definitions.html#_v1_devices)
+    * autoattachGraphicsDevice
+      * `VirtualMachinePreferenceSpec.Devices.PreferredAutoattachGraphicsDevice`
+    * autoattachMemBalloon
+      * `VirtualMachinePreferenceSpec.Devices.PreferredAutoattachMemBalloon`
+    * autoattachPodInterface
+      * `VirtualMachinePreferenceSpec.Devices.PreferredAutoattachPodInterface`
+    * autoattachSerialConsole
+      * `VirtualMachinePreferenceSpec.Devices.PreferredAutoattachSerialConsole`
+    * blockMultiQueue
+      * `VirtualMachineFlavorSpec.Devices.BlockMultiQueue`
+    * clientPassthrough
+      * None
+    * disableHotplug
+      * `VirtualMachinePreferenceSpec.Devices.PreferredDisableHotplug`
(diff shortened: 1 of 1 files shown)


design-proposals/preferences.md
```diff
@@ -0,0 +1,584 @@
+# Overview
+
+A common pattern for IaaS is to have an abstraction separating the sizing and performance characteristics of a virtual machine from the user defined values related to launching their custom application. This pattern is evident across all the major cloud providers as well as open source IaaS projects like OpenStack. AWS calls this abstraction an [instance type](https://aws.amazon.com/ec2/instance-types/), GCP calls it a [machine type](https://cloud.google.com/compute/docs/machine-types#custom_machine_types). Azure calls it an [instance VM size](https://docs.microsoft.com/en-us/azure/virtual-machines/sizes), and OpenStack has [flavors](https://docs.openstack.org/nova/latest/user/flavors.html) and [image properties](https://docs.openstack.org/glance/latest/admin/useful-image-properties.html).
+
```
dominikholler, line 4: What is the pendant of googles machine families and azures or aws instance types in the presented approach, and what how is the sizing in the approach modeled?

design-proposals/preferences.md
```diff
@@ -0,0 +1,584 @@
...
+In OpenStack for example this issue is mostly avoided through the use of [image properties](https://docs.openstack.org/glance/latest/admin/useful-image-properties.html). This is where the boot image of an instance (`VMI`) is decorated with properties in order to override or augment preferences defined by [flavors](https://docs.openstack.org/nova/latest/user/flavors.html).
+
+As this isn't possible within `KubeVirt` another mechanism is required to separate the quantitative compute resources defined by flavors and qualitative preferences we also want to provide.
+
+## User Stories
+
+* As an operator or user I would like to provide flavors that define the compute resources available to a VM.
+
+* As an operator or user I would like to provide preferences that relate to the performance characteristics of a VM within these flavors.
```
dominikholler, line 23: In which way they relate?

design-proposals/preferences.md
```diff
@@ -0,0 +1,584 @@
...
+
+* As an operator, user or third party (application or distribution vendor) I would like to independently provide preferences that relate to the performance characteristics of a VM outside of a flavor.
```
dominikholler, line 25:
```suggestion
* As an operator, user or third party (application or distribution) guest VM vendor I would like to independently provide preferences that relate to the performance characteristics of a VM outside of a flavor.
```

design-proposals/preferences.md
```diff
@@ -0,0 +1,584 @@
...
+
+* Introducing an `optional` policy for some of the preferences as done with OpenStack [Flavor extra-specs](https://docs.openstack.org/nova/latest//configuration/extra-specs.html#hw:cpu_thread_policy). It isn't clear how we would handle scheduling failures and retries here within KubeVirt/k8s.
+
+* Versioning of [`VirtualMachineFlavor`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineflavor),  [`VirtualMachineClusterFlavor`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineclusterflavor), `VirtualMachinePreferences` and `VirtualMachineClusterPreferences`.
+
+* Extracting [`Resources`](http://kubevirt.io/api-reference/main/definitions.html#_k8s_io_api_core_v1_resourcerequirements) from the [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec).
+
+* Adding more supported [`Extended Resources`](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#extended-resources) such as [`GPUs`](http://kubevirt.io/api-reference/main/definitions.html#_v1_gpu) and [`HostDevices`](http://kubevirt.io/api-reference/main/definitions.html#_v1_hostdevice) etc.
```
dominikholler, line 41: It would be nice, if the proposal could show an idea, how this could be modeled, because I guess this is tightly coupled with the way the VM families/types are modeled.

design-proposals/preferences.md
```diff
@@ -0,0 +1,584 @@
...
+  DevicesDefaults *DevicesDefaults `json:"devicesDefaults,omitempty"`
+  
+}
+```
+
+### `VirtualMachinePreferencesSpec`
+
+This is a copy of `DomainSpec` with every field marked as optional. The aim being to provide any preference or tunable using the same topology as a normal `DomainSpec` when provided to a `VirtualMachine`.
+
+The following items however will cause the `VirtualMachinePreferencesSpec` to be rejected if provided:
```
dominikholler, line 125: When which input would be rejected by which api call?

design-proposals/preferences.md
```diff
@@ -0,0 +1,584 @@
...
+
+  // PreferredCdromBus optionally defines the preferred bus for Cdrom Disk devices.
+  //
+  // +optional
+  PreferredCdromBus string `json:"preferredCdromBus,omitempty"`
+
+  // PreferredDedicatedIoThread optionally enables dedicated IO threads for Disk devices.
+  //
+  // +optional
+  PreferredDedicatedIoThread *bool `json:"preferredDedicatedIoThread,omitempty"`
```
dominikholler, line 250: Isn't this about the resources provided to the VM?

design-proposals/preferences.md
```diff
@@ -0,0 +1,584 @@
...
+type VirtualMachineFlavorProfile struct {
+
+  // Name specifies the name of this custom profile.
+  Name string `json:"name"`
+
+  // Default specifies if this VirtualMachinePreferencesProfile is the default for the VirtualMachineFlavor.
+  // Zero or one profile can be set to default.
+  //
+  // +optional
+  Default bool `json:"default,omitempty"`
```
dominikholler, line 310: Does this mean the whole API call is invalid, if multiple items with Default=true are sent or received?

dominikholler (changes requested):

design-proposals/preferences.md
lyarwood: I can add some examples if that would help?

design-proposals/preferences.md
lyarwood: See the examples at the end where I suggest two options for providing the preferences, either embedded in the flavor or alongside within the VM definition.

design-proposals/preferences.md
lyarwood: ACK.

design-proposals/preferences.md
lyarwood: Yes, however I've dropped profiles from the proposal moving forward.

design-proposals/preferences.md
lyarwood: Right I've left a TODO to determine which side things like this need to fall in the latest revision but I might expand this into a section covering it in a little more detail.

design-proposals/preferences.md
lyarwood: I've dropped L41 after a discussion last week about extended resources, I wasn't aware that GPUs etc can be expressed as resource requests.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,408 @@
...
+        * virtiofs None
+      * [Gpus](http://kubevirt.io/api-reference/main/definitions.html#_v1_gpu)
+        * deviceName `VirtualMachineFlavorSpec.Devices.GPUs`
+        * name None
+        * tag `VirtualMachineFlavorSpec.Devices.GPUs`
+        * virtualGPUOptions `VirtualMachinePreferencesSpec.Devices.GPUs.virtualGPUOptions`
+      * [HostDevices](http://kubevirt.io/api-reference/main/definitions.html#_v1_hostdevice)
+        * deviceName `VirtualMachineFlavorSpec.Devices.HostDevices`
+        * name None
+        * tag  `VirtualMachineFlavorSpec.Devices.HostDevices`
```
dominikholler, line 133: I am unsure.

dominikholler (commented):
Great improvement. There are some DomainSpec mapping I am unsure, due to my lack of understanding how the particular feature works, but I agree with all the others I did not comment.

design-proposals/flavors-and-preferences.md
lyarwood: ACK, I thought providing a tag would help users identify devices easier but we can add this in later if requried.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,408 @@
...
+      * [Inputs](http://kubevirt.io/api-reference/main/definitions.html#_v1_input)
+        * bus `VirtualMachinePreferencesSpec.Devices.PreferredInputBus`
+        * Name None
+        * Type `VirtualMachinePreferencesSpec.Devices.PreferredInputType`
+      * [Interfaces](http://kubevirt.io/api-reference/main/definitions.html#_v1_interface)
+        * bootOrder None
+        * bridge None
+        * dhcpoptions `VirtualMachinePreferencesSpec.Devices.Interfaces.PreferredDhcpOptions`
+        * macaddress None
+        * [macvtap](http://kubevirt.io/api-reference/main/definitions.html#_v1_interfacemacvtap) `VirtualMachinePreferencesSpec.Devices.Interfaces.Macvtap`
```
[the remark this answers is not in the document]
lyarwood: ACK removed.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,408 @@
...
+    * isolateEmulatorThread `VirtualMachineFlavorSpec.CPU.isolateEmulatorThread`
+    * Model `VirtualMachineFlavorSpec.CPU.Model`
+    * [NUMA](http://kubevirt.io/api-reference/main/definitions.html#_v1_watchdog) `VirtualMachineFlavorSpec.CPU.NUMA`
+    * [Realtime](http://kubevirt.io/api-reference/main/definitions.html#_v1_realtime) `VirtualMachineFlavorSpec.CPU.Realtime`
+  * [Devices](http://kubevirt.io/api-reference/main/definitions.html#_v1_devices)
+    * autoattachGraphicsDevice `VirtualMachinePreferencesSpec.Devices.autoattachGraphicsDevice`
+    * autoattachMemBalloon `VirtualMachinePreferencesSpec.Devices.autoattachMemBalloon`
+    * autoattachPodInterface `VirtualMachinePreferencesSpec.Devices.autoattachPodInterface`
+    * autoattachSerialConsole `VirtualMachinePreferencesSpec.Devices.autoattachSerialConsole`
+    * blockMultiQueue `VirtualMachineFlavorSpec.Devices.blockMultiQueue`
```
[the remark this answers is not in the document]
lyarwood: Specifically about what?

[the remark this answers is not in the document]
dominikholler: I am unsure if the blockMultiQueue consumes ressources from the host or from the VM.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,475 @@
...
+        * Name
+          * None
+        * Type
+          * `VirtualMachinePreferencesSpec.Devices.PreferredInputType`
+      * [Interfaces](http://kubevirt.io/api-reference/main/definitions.html#_v1_interface)
+        * bootOrder
+          * None
+        * bridge
+          * None
+        * dhcpoptions
```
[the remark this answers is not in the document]
phoracek: No. I have never seen anybody using it.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,475 @@
...
+* Introspection of imported images to determine the correct guest OS related `VirtualMachinePreferences` to apply.
+
+* Using image labels to determine the correct guest OS related `VirtualMachinePreferences` to apply.
+
+## Repos
+
+* [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+
```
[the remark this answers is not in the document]
lyarwood: Done.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,475 @@
...
+* Remove the need to define [`Interfaces`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachineinstancenetworkinterface) within [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec) when providing [`Networks`](http://kubevirt.io/api-reference/main/definitions.html#_v1_network) within a [`VirtualMachineInstanceSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachineinstancespec).
+
+
+## [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec) mappings to `VirtualMachinePreferencesSpec` and `VirtualMachineFlavorSpec`
+
+The following basic rules are applied when deciding which attributes of the [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec) will be mapped to the new `VirtualMachinePreferencesSpec` and `VirtualMachineFlavorSpec` types:
+
+* Virtual resource and performance related preferences reside within `VirtualMachineFlavorSpec`
```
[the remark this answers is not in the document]
lyarwood: Done.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,475 @@
...
+      * `VirtualMachineFlavorSpec.CPU.isolateEmulatorThread`
+    * model
+      * `VirtualMachineFlavorSpec.CPU.Model`
+    * [NUMA](http://kubevirt.io/api-reference/main/definitions.html#_v1_watchdog)
+      * `VirtualMachineFlavorSpec.CPU.NUMA`
+    * [Realtime](http://kubevirt.io/api-reference/main/definitions.html#_v1_realtime)
+      * `VirtualMachineFlavorSpec.CPU.Realtime`
+  * [Devices](http://kubevirt.io/api-reference/main/definitions.html#_v1_devices)
+    * autoattachGraphicsDevice
+      * `VirtualMachinePreferencesSpec.Devices.autoattachGraphicsDevice`
```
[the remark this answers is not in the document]
lyarwood: ACK I'm going to go with `Preferred` ahead of each name in the next revison.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,475 @@
...
+  name: rhel9
+spec:
+  flavor:
+    kind: VirtualMachineFlavor
+    name: c.large
+  preferences:
+    kind: VirtualMachinePreferences
+    name: linux.rhel9
+  networks:
+    - name: private
```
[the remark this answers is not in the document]
lyarwood: ACK

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,475 @@
...
+
+* Extend [`VirtualMachineSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachinespec) to allow an optional reference to `VirtualMachinePreferences` or `VirtualMachineClusterPreferences`.
+
+* Remove the need to define [`Disks`](http://kubevirt.io/api-reference/main/definitions.html#_v1_disk) within [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec) when providing [`Volumes`](http://kubevirt.io/api-reference/main/definitions.html#_v1_volume) within a [`VirtualMachineInstanceSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachineinstancespec).
+
+* Remove the need to define [`Interfaces`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachineinstancenetworkinterface) within [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec) when providing [`Networks`](http://kubevirt.io/api-reference/main/definitions.html#_v1_network) within a [`VirtualMachineInstanceSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachineinstancespec).
```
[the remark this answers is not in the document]
lyarwood: I'm going to drop these two points back into the non-goal section as they are only going to complicate an already complicated design doc IMHO. Definitively something we can address separately and even in parallel to this effort IMHO.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,475 @@
...
+        * `VirtualMachinePreferencesSpec.Devices.RNG`
+      * [Sound](http://kubevirt.io/api-reference/main/definitions.html#_v1_sounddevice)
+        * model
+          * `VirtualMachinePreferencesSpec.Devices.Sound.PreferredModel`
+        * name
+          * None
+      * useVirtioTransitional
+        * `VirtualMachinePreferencesSpec.Devices.useVirtioTransitional`
+      * [Watchdog](http://kubevirt.io/api-reference/main/definitions.html#_v1_watchdog)
+        * [I6300esb](http://kubevirt.io/api-reference/main/definitions.html#_v1_i6300esbwatchdog)
```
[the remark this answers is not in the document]
lyarwood: ACK I'll drop it for now.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,475 @@
...
+    * [smm](http://kubevirt.io/api-reference/main/definitions.html#_v1_featurestate)
+      * `VirtualMachinePreferencesSpec.Features.Smm`
+  * ioThreadsPolicy
+    * `VirtualMachineFlavorSpec.ioThreadsPolicy`
+  * [launchSecurity](http://kubevirt.io/api-reference/main/definitions.html#_v1_launchsecurity)
+    * [sev](http://kubevirt.io/api-reference/main/definitions.html#_v1_sev)
+      * `VirtualMachineFlavorSpec.LaunchSecurity.SEV`
+  * [Machine](http://kubevirt.io/api-reference/main/definitions.html#_v1_machine)
+    * type
+      * `VirtualMachinePreferencesSpec.Machine.PreferredType`
```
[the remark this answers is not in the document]
lyarwood: I don't think it is, how would we know the supported versions ahead of launching virt-launcher that itself contains a version of QEMU capable of launching specific versions of machine types?

design-proposals/flavors-and-preferences.md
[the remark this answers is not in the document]
lyarwood: I've moved this back to the flavor given the recent changes.

design-proposals/flavors-and-preferences.md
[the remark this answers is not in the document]
fabiand: virt handler is launching an instance of launcher today, to discover node details.

But I just checked, and no: No MT info as labels.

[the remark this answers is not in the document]
lyarwood: ACK apologies I see the code now, smells like a valid RFE to at least list the supported types and at least the latest version?

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,482 @@
...
+
+* Extend [`VirtualMachineSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachinespec) to allow an optional reference to `VirtualMachinePreference` or `VirtualMachineClusterPreference`.
+
+## [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec) mappings to `VirtualMachinePreferenceSpec` and `VirtualMachineFlavorSpec`
+
+The following basic rules are applied when deciding which attributes of the [`DomainSpec`](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec) will be mapped to the new `VirtualMachinePreferenceSpec` and `VirtualMachineFlavorSpec` types:
+
+* Resource, schedulable and performance related attributes of the DomainSpec reside within `VirtualMachineFlavorSpec`
+
+* Guest OS related Preference reside within `VirtualMachinePreferenceSpec`
```
fabiand, line 76: Not sure where the comment belongs, but: Because it's preferences, I assume that a user is able to define some domain value in VMI.spec.domain or rather VM.spec.template.spec.domain, like autoAttachGraphicalDisplay to set his custom specified value.
If this is not set, then the value from the preference would be taken.

Is this assumption correct?

lyarwood: ACK yeah I'll expand the initial design section to contain this and some related points, thanks!

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,492 @@
...
+* Guest OS related preferences reside within `VirtualMachinePreferenceSpec`
+
+For example, the number of and topology of vCPUs presented to the guest are defined within `VirtualMachineFlavorSpec` but the preferred topology of these vCPUs resides within  `VirtualMachinePreferenceSpec`.
+
+* [DomainSpec](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec)
+  * [CPU](http://kubevirt.io/api-reference/main/definitions.html#_v1_cpu)
+    * cores 
+      * `VirtualMachineFlavorSpec.CPU.Guest` 
+        * Number of vCPUs to expose to the guest
+      * `VirtualMachinePreferenceSpec.CPU.PreferredTopology`
```
fabiand, line 91: I suppose this should go to flavor as well, reasoning: it's a hardware resource we expose.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,492 @@
...
+        * Guest visible CPU topology to expose, `preferSockets` or `preferCores` (default).
+    * sockets
+      * `VirtualMachineFlavorSpec.CPU.Guest` 
+        * Number of vCPUs to expose to the guest
+      * `VirtualMachinePreferenceSpec.CPU.PreferredTopology`
```
fabiand, line 96: Same here.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,492 @@
...
+        * Guest visible CPU topology to expose, `preferSockets` or `preferCores` (default).
+    * threads
+      * `VirtualMachineFlavorSpec.CPU.Guest` 
+        * Number of vCPUs to expose to the guest
+      * `VirtualMachinePreferenceSpec.CPU.PreferredTopology`
```
fabiand, line 101: Same here.

design-proposals/flavors-and-preferences.md
lyarwood: I had a conversation with @vladikr (keep me honest here please) about this a while ago and at the time I was advocating for the same given that's what we also did in OpenStack with extra specs such as [`hw:cpu_sockets`](https://docs.openstack.org/nova/latest/configuration/extra-specs.html#hw:cpu_sockets), [`hw:cpu_cores`](https://docs.openstack.org/nova/latest/configuration/extra-specs.html#hw:cpu_cores) and [`hw:cpu_threads`](https://docs.openstack.org/nova/latest/configuration/extra-specs.html#hw:cpu_threads).

I think the only reason we went  with PreferredTopology in the VirtualMachinePreferenceSpec was due to the Windows use case that keeps coming up where it prefers sockets over cores. So having it as a preference removes possible duplication from the flavors.

iholder-redhat (commented):
Thanks @lyarwood! Great to see this progressing!
left some questions :smiley: 

Also, a thought:
If someone wants to have a flavor that configures some spec element that is yet to be implemented, there's eventually a need to change the code, which might take a while. Maybe we can introduce another generic option in Flavors to provide json patches. I think this might be valuable either for users waiting for a new Flavor feature to come out or to users with very special needs that are not part of the Flavor API. WDYT?

lyarwood:
> Thanks @lyarwood! Great to see this progressing! left some questions smiley
> 
> Also, a thought: If someone wants to have a flavor that configures some spec element that is yet to be implemented, there's eventually a need to change the code, which might take a while. Maybe we can introduce another generic option in Flavors to provide json patches. I think this might be valuable either for users waiting for a new Flavor feature to come out or to users with very special needs that are not part of the Flavor API. WDYT?

That's a valid idea but it smells a little of feature creep at this point. I'd rather go with the simple hardcoded approach for now and potentially extend the design with something like this later once the need arises.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,475 @@
...
+        * pciAddress
+          * None
+        * ports
+          * None
+        * [slirp](http://kubevirt.io/api-reference/main/definitions.html#_v1_interfaceslirp) 
+          * None
+        * [sriov](http://kubevirt.io/api-reference/main/definitions.html#_v1_interfacesriov) 
+          * None
+        * Tag
+          * None
```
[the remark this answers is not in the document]
vladikr: By default, no tags will be generated, since the original idea was to let users tag only the relevant devices with a tag that would be meaningful to the guest user (i.e "fastNic").
I think generating these tags using the device name, in a consistent way, would work. We should also have a way to not generate the tags at all.

lyarwood:
> Another thought I had in mind - what about the ability to construct hierarchies of flavors to allow re-use of configurations? For example, I assume that flavors like `c4.xlarge` and `c4.xmedium` would have a lot in common. Does it make sense to being able to have a "base" flavor that other would further expend?

Yeah I wanted to go that route initially but after some back and fourth outside of this review I've been won over by the idea of pushing users towards tools like  [`kustomize`](https://kustomize.io/) to avoid this duplication. There should be a note about this in the current proposal, I'm also working on an example [repo](https://github.com/lyarwood/vmdefs/tree/preferences) alongside the proposal.

iholder-redhat:
> > Another thought I had in mind - what about the ability to construct hierarchies of flavors to allow re-use of configurations? For example, I assume that flavors like `c4.xlarge` and `c4.xmedium` would have a lot in common. Does it make sense to being able to have a "base" flavor that other would further expend?
> 
> Yeah I wanted to go that route initially but after some back and forth outside of this review I've been won over by the idea of pushing users towards tools like [`kustomize`](https://kustomize.io/) to avoid this duplication. There should be a note about this in the current proposal, I'm also working on an example [repo](https://github.com/lyarwood/vmdefs/tree/preferences) alongside the proposal.

This is valid point indeed.
Thanks!

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,1023 @@
...
+### `VirtualMachineFlavorSpec`
+
+```go
+// VirtualMachineFlavorSpec
+//
+// +k8s:openapi-gen=true1
+type VirtualMachineFlavorSpec struct {
+	CPU *CPUFlavor `json:"cpu"`
+
+	Memory *MemoryFlavor `json:"memory"`
```
fabiand, line 314: Nit pick, and not relevant in the design, but for API documentation this field should get some comment as well.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,1023 @@
...
+
+```go
+// CPUFlavor
+//
+// +k8s:openapi-gen=true
+type CPUFlavor struct {
+
+	// Number of vCPUs to expose to the guest.
+	// The resulting CPU topology being derived from the optional PreferredCPUTopology attribute of CPUPreferences.
+	Guest uint32 `json:"guest"`
```
fabiand, line 350: Same here: We should include a unit in the field name

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,1023 @@
...
+	DedicatedCPUPlacement bool `json:"dedicatedCPUPlacement,omitempty"`
+
+	// NUMA allows specifying settings for the guest NUMA topology
+	// +optional
+	NUMA *v1.NUMA `json:"numa,omitempty"`
+
+	// IsolateEmulatorThread requests one more dedicated pCPU to be allocated for the VMI to place
+	// the emulator thread on it.
+	// +optional
+	IsolateEmulatorThread bool `json:"isolateEmulatorThread,omitempty"`
```
fabiand, line 372: I wonder if it would be safer to use a struct here in order to keep the door open to reflect different emulator thread isolation policies in future.
@vladikr thoughts?

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,1023 @@
...
+### `MemoryFlavor`
+
+```go
+// FlavorMemory
+//
+// +k8s:openapi-gen=true
+type MemoryFlavor struct {
+
+	// Guest allows to specifying the amount of memory which is visible inside the Guest OS.
+	Guest *resource.Quantity `json:"guest,omitempty"`
```
fabiand, line 389: A Unit is appreciated here as well

design-proposals/flavors-and-preferences.md
fabiand: OTOH - Maybe we should stick aligned to the VMI API, as we'd need the change there as well.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,1023 @@
...
+}
+```
+
+### `CPUPreferences`
+
+```go
+// +k8s:openapi-gen=true
+type PreferredCPUTopology string
+
+const (
```
fabiand, line 485: In the code we want to have API documentation for these consts

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,1023 @@
...
+
+`VirtualMachineFlavorSpec` will also include elements of `CPU`, `Devices`, `HostDevices`, `GPUs`, `Memory` and `LaunchSecurity` defined fully below. Users will be unable to override any aspect of the flavor (for example, `vCPU` count or amount of `Memory`) within the `VirtualMachine` itself, any attempt to do so resulting in the `VirtualMachine` being rejected.
+
+A new set of `VirtualMachinePreference` CRDs will then be introduced to define any remaining attributes related to ensuring the selected guestOS can run. As the name suggests the `VirtualMachinePreference` CRDs will only define preferences, so unlike a flavor if a preference conflicts with something user defined within the `VirtualMachine` it will be ignored. For example, if a user selects a `VirtualMachinePreference` that requests a `preferredDiskBus` of `virtio` but then sets a disk bus of `SATA` for one or more disk devices within the `VirtualMachine` the supplied `preferredDiskBus` preference will not be applied to these disks. Any remaining disks that do not have a disk bus defined will however use the `preferredDiskBus` preference of `virtio`.
```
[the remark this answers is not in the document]
lyarwood: Possibly but you could then argue the same for `VirtualMachineFlavor` that at present only has an impact on the `DomainSpec`.

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,1023 @@
...
+
+* [DomainSpec](http://kubevirt.io/api-reference/main/definitions.html#_v1_domainspec)
+  * [CPU](http://kubevirt.io/api-reference/main/definitions.html#_v1_cpu)
+    * cores 
+      * `VirtualMachineFlavorSpec.CPU.Guest` 
+        * Number of vCPUs to expose to the guest
+      * `VirtualMachinePreferenceSpec.CPU.PreferredTopology`
+        * Guest visible CPU topology to expose, `preferSockets` , `preferThreads` or `preferCores` (default).
+    * sockets
+      * `VirtualMachineFlavorSpec.CPU.Guest`
```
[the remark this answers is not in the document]
lyarwood: I was following the example from [Memory](http://kubevirt.io/api-reference/main/definitions.html#_v1_memory) but I can make both more specific if the difference between the flavor and `VirtualMachineInstanceSpec` isn't an issue.

design-proposals/flavors-and-preferences.md
lyarwood: ACK I'll add these in the series I have posted and update the design proposal.

design-proposals/flavors-and-preferences.md
lyarwood: ACK.

design-proposals/flavors-and-preferences.md
lyarwood: ACK, as above if it's fine to differ from the VM API I can do that here.

design-proposals/flavors-and-preferences.md
lyarwood: ACK

design-proposals/flavors-and-preferences.md
```diff
@@ -0,0 +1,1023 @@
...
+
+This approach has a few pitfalls such as using embedded profiles within the CRDs, relying on the user to select the correct Flavor or [`VirtualMachineFlavorProfile`](http://kubevirt.io/api-reference/main/definitions.html#_v1alpha1_virtualmachineflavorprofile) that will allow their workload to run correctly, not allowing a user to override some viable attributes at runtime etc.
```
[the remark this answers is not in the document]
lyarwood: ACK.

lyarwood:
@fabiand @davidvossel @iholder-redhat given the series is now approved (and should hopefully be LGTM'd again shortly after a rebase) I was hopeful that we could also land this initial version of the design doc. As with the series I'll be coming back to this document and updating it with any future changes to the design before we hit beta.

davidvossel (commented):
/lgtm

PR #166, opened 2022-03-07
Template Generation Design Proposal

acardace:
Signed-off-by: Antonio Cardace <acardace@redhat.com>

### DIFF
--- a/design-proposals/template-generation.md
+++ b/design-proposals/template-generation.md
@@ -1,0 +1,159 @@
+# Overview
+
+This design defines an approach to create Virtual Machine templates starting from an existing VM.
+
+## Motivation
+
+Creating a 'blueprint' of an existing VM to further tweak and use as a basis golden VM template is a common operation, also it is a tool that can be leveraged by UIs to improve the UX of the project.
+
+## Goals
+
+* Create a generic template of an existing VM.
+
+## Non Goals
+
+* Designing a VM-cloning mechanism, which copies disk data or create a snapshot of the live memory of a VM.
+
+## User Stories
+
+* As a cluster user, I want to create a re-usable VM spec to use as the basis for my VMs workloads starting from an existing VM.
+* As a cluster user, I want to create different workloads based on the same VM image by just tweaking some template parameters.
+
+## Repos
+
+The idea is to implement this design by adding a controller in kubevirt/ssp-operator.
+
+# Design
+
+The template generation API introduces a new CRD called the 'VirtualMachineTemplateRequest' object. This CRD when posted to the cluster is an explicit request to create a VM template derived from an existing VM.
+
+The controller watching `VirtualMachineTemplateRequest` objects will see the request and post a template of the request VirtualMachine in the specified format.
+
+## VirtualMachineTemplateRequest API
+
+The VirtualMachineTemplateRequest specification in Golang:
+
+```go
+// VirtualMachineTemplateRequestSource is the reference VM that will be used to generate the template
+type VirtualMachineTemplateRequestSource struct {
+	// VMName is the name of the reference VM that will be used to generate the template
+	VMName string `json:"vmName"`
+	// VMNamespace is the namespace of the VM referenced by VMName
+	VMNamespace string `json:"vmNamespace"`
+}
+
+// VirtualMachineTemplateRequestTarget is the target VM template that will be created
+type VirtualMachineTemplateRequestTarget struct {
+	// OpenShiftTemplate is the OpenShift template format
+	OpenShiftTemplate *OpenShiftTemplate `json:"openshiftTemplate,omitempty"`
+}
+
+// VirtualMachineTemplateRequestSpec contains the VirtualMachineTemplateRequest specification
+type VirtualMachineTemplateRequestSpec struct {
+	// Source is the reference VM from which the template will be created
+	Source VirtualMachineTemplateRequestSource `json:"source"`
+	// Target is the target VM template that will be created
+	Target VirtualMachineTemplateRequestTarget `json:"target"`
+}
+
+type OpenShiftTemplate struct {
+	// Name is the name of the generated template
+	Name string `json:"name,omitempty"`
+}
+```
+
+The first implemented format will be the [OpenShift Template(https://docs.openshift.com/container-platform/4.9/openshift_images/using-templates.html#templates-writing_using-templates)] format.
+
+In order to support multiple formats, the controller will have to be extended to generate the correct output for those formats.
+
+## API Examples
+
+Given the following VM
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: vm-ephemeral
+spec:
+  template:
+    domain:
+      devices:
+        disks:
+        - disk:
+            bus: virtio
+          name: containerdisk
+      resources:
+        requests:
+          memory: 128Mi
+    volumes:
+    - containerDisk:
+        image: registry:5000/kubevirt/cirros-container-disk-demo:devel
+      name: containerdisk
+```
+
+After posting this VirtualMachineTemplateRequest object
+
+```yaml
+apiVersion: ssp.kubevirt.io/v1alpha1
+kind: VirtualMachineTemplateRequest
+metadata:
+  name: virtualmachinetemplaterequest-sample
+spec:
+  vmName: vmi-ephemeral
+  templateName: vmi-ephemeral-template
+  format: template.openshift.io/v1
+```
+
+The following template will be created, beware this is an example of the OpenShift template format.
+
+```yaml
+apiVersion: v1
+kind: Template
+metadata:
+  name: vmi-ephemeral-template
+  labels: ${LABELS}
+objects:
+- apiVersion: kubevirt.io/v1
+  kind: VirtualMachine
+  metadata:
+    generateName: vmi-ephemeral
+  template:
+    spec:
+      domain:
+        devices:
+          disks:
+          - disk:
+              bus: ${DISK0_BUS}
+            name: containerdisk
+        resources:
+          requests:
+            memory: ${REQUESTS_MEMORY}
+      volumes:
+      - containerDisk:
+          image: ${VOLUME0_IMAGE}
+        name: containerdisk
+parameters:
+- description: disk0-bus
+  name: DISK0_BUS
+  value: virtio
+- description: requests-memory
+  name: REQUESTS_MEMORY
+  value: 128Mi
+- description: volume0-image
+  name: VOLUME0_IMAGE
+  value: registry:5000/kubevirt/cirros-container-disk-demo:devel
+```
+
+## Template generation
+
+The template generation mechanism follows some basic rules:
+
+* every leaf field (but name fields) in the `VirtualMachineInstanceSpec` will be parameterized and given a default value according to the reference VM which has been created from.
+* the VM name will be converted into a `generateName` field.
+* the `dataVolumeTemplates` field inside a `VirtualMachineSpec` will be preserved as is (as it's already templated).
+* the `LABELS` parameter will always be created, its content will be merged with the existing labels of the `VirtualMachine` object.
+
+## Further developments
+
+* Add `virtctl` support for creating templates from the cmdline.


design-proposals/template-generation.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+## Template generation
+
+The template generation mechanism follows some basic rules:
+
+* every leaf field (but name fields) in the `VirtualMachineInstanceSpec` will be parameterized and given a default value according to the reference VM which has been created from.
+* the VM name will be converted into a `generateName` field.
+* the `dataVolumeTemplates` field inside a `VirtualMachineSpec` will be preserved as is (as it's already templated).
+* the `LABELS` paramater will always be created, its content will be merged with the existing labels of the `VirtualMachine` object.
```
jean-edouard, line 127: Typo: s/paramater/parameter/

design-proposals/template-generation.md
```diff
@@ -0,0 +1,127 @@
...
+
+```yaml
+apiVersion: ssp.kubevirt.io/v1alpha1
+kind: VirtualMachineTemplateRequest
+metadata:
+  name: virtualmachinetemplaterequest-sample
+spec:
+  vmName: vmi-ephemeral
+  templateName: vmi-ephemeral-template
+  Format: OpenShift4
```
jean-edouard, line 77: `format` should probably be lowercase
Shouldn't we target an API version instead of an OpenShift version? Maybe we can even use the API version verbatim, like `template.openshift.io/v1`?

design-proposals/template-generation.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Non Goals
+
+* Designing a VM-cloning mechanism, which copies disk data or create a snapshot of the live memory of a VM.
+
+## User Stories
+
+* As a cluster user, I want to create a re-usable VM spec to use as the basis for my VMs workloads starting from a running VM.
+* As a cluster user, I want to create different workloads based on the same VM image by just tweaking some template parameters.
+
```
jean-edouard, line 21: In the template, there's a `repos` section right about there. I think it would make sense to specify where the work would happen (ssp-operator?).

design-proposals/template-generation.md
```diff
@@ -0,0 +1,127 @@
...
+      resources:
+        requests:
+          memory: 128Mi
+    volumes:
+    - containerDisk:
+        image: registry:5000/kubevirt/cirros-container-disk-demo:devel
+      name: containerdisk
+```
+
+After posting this VirtualMachineTemplateRequest object
```
jean-edouard, line 67: Probably out of scope (but then let's add it to non-goals), do we want a more user-friendly way to create templates from VMs, with something like `virtctl`?

design-proposals/template-generation.md
```diff
@@ -0,0 +1,127 @@
...
+The controller watching `VirtualMachineTemplateRequest` objects will see the request and post a template of the request VirtualMachine in the specified format.
+    
+## VirtualMachineTemplateRequest API
+
+The VirtualMachineTemplateRequest specification fields:
+
+* **VMName**: the name of a VirtualMachine living in the cluster which will be the reference from which the template will be created.
+* **VMNamespace**: the namespace of the reference VM. This field is optional if not specified the VM will be looked up in the namepace in which VirtualMachineTemplateRequest was created.
+* **TemplateName**: the name of the template object that will be generated, if omitted it will be auto-generated.
+* **Format**: the format of the output template.
```
jean-edouard, line 35: Optional?

jean-edouard (commented):
Solid design proposal, I agree with the idea.
Comments in-line. Thanks!

design-proposals/template-generation.md
acardace: Added!

design-proposals/template-generation.md
acardace: fixed, thanks!

design-proposals/template-generation.md
acardace: Do you mean the `format` field? I'd leave it as required, why would you make it optional?

design-proposals/template-generation.md
acardace: Very nice idea! Definitely, something to implement after the design has been approved, but I'd leave it out of this design document if you agree.

design-proposals/template-generation.md
acardace: The idea is to have something rather general because while tomorrow only OCP might be supported there should be no restrictions in extending the controller to support more template formats outside of OCP.

I like the idea of using the API group and version! I'll change it.

design-proposals/template-generation.md
jean-edouard: It's fine either way, I just think it would be nice to specify that it's a required field, but I guess if you don't say anything it means "required"?

acardace: That was my assumption, but since it wasn't clear I added it in the description.

design-proposals/template-generation.md
```diff
@@ -0,0 +1,131 @@
...
+- description: volume0-image
+  name: VOLUME0_IMAGE
+  value: registry:5000/kubevirt/cirros-container-disk-demo:devel
+```
+
+## Template generation
+
+The template generation mechanism follows some basic rules:
+
+* every leaf field (but name fields) in the `VirtualMachineInstanceSpec` will be parameterized and given a default value according to the reference VM which has been created from.
```
xpivarc, line 128: (except for name fields?)

design-proposals/template-generation.md
```diff
@@ -0,0 +1,131 @@
...
+* the VM name will be converted into a `generateName` field.
```
xpivarc, line 129: I guess this is only true when the **VMName** is not specified

design-proposals/template-generation.md
```diff
@@ -0,0 +1,131 @@
...
+* the `dataVolumeTemplates` field inside a `VirtualMachineSpec` will be preserved as is (as it's already templated).
+* the `LABELS` parameter will always be created, its content will be merged with the existing labels of the `VirtualMachine` object.
```
xpivarc, line 131: What about `annotations`?

design-proposals/template-generation.md
xpivarc, line 128: This sounds like a generic solution from which every format can do specific modifications. I guess if we go with the suggested embedded VM object we could already embed VM with "parametrized" leafs. 

One question comes to my mind because I am not sure if it is straightforward. 
Let's take volumes/disks and let's say VM has a disk of pvc type. Now, what is the expected behavior here? 
1. Should we just have the same volume with the default pointing to the same pvc(which would lead to VMs using same disk)?
2. Should we use `datavolumesTemplates` section to generate similar disks?
3. The disk is immutable(read containerDisk) and we don't care.

I think this needs further clarification in the design if I didn't miss anything. We can debate if similar complication exist for networks.

design-proposals/template-generation.md
xpivarc: Can we add section with implementation phases/expected improvements and mention this there?

design-proposals/template-generation.md
```diff
@@ -0,0 +1,131 @@
...
+  name: REQUESTS_MEMORY
+  value: 128Mi
+- description: volume0-image
+  name: VOLUME0_IMAGE
+  value: registry:5000/kubevirt/cirros-container-disk-demo:devel
+```
+
+## Template generation
+
+The template generation mechanism follows some basic rules:
```
xpivarc, line 126: We recently added flavors to core kubevirt API. What is your vision of the flavor in your design? (Can we use them for possible suggested parameters?)

xpivarc (commented):
Great start, I think we just need to describe little bit more what is the expected behavior.

design-proposals/template-generation.md
```diff
@@ -0,0 +1,131 @@
...
+* As a cluster user, I want to create a re-usable VM spec to use as the basis for my VMs workloads starting from a running VM.
+* As a cluster user, I want to create different workloads based on the same VM image by just tweaking some template parameters.
+
+## Repos
+
+The idea is to implement this design by adding a controller in kubevirt/ssp-operator.
+
+# Design
+
+The template generation API introduces a new CRD called the 'VirtualMachineTemplateRequest' object. This CRD when posted to the cluster is an explicit request to create a VM template derived from a running VM.
```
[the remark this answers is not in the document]
acardace: Sorry, I think I used the wrong term, I meant an existing VM in the cluster, not necessarily a running one.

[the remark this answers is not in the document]
acardace: I'll remove the reference to 'running'.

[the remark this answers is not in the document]
xpivarc: I think my question still applies. One can still modify the VM, right?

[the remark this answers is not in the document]
acardace: Yes of course, but the template is a snapshot at a given time so the template won't be updated automatically.
We can reference the VM unambiguously by referencing the `metadata.generation` field in a label added to the template, what do you think?

design-proposals/template-generation.md
```diff
@@ -0,0 +1,131 @@
...
+The VirtualMachineTemplateRequest specification fields:
+
+* **VMName**: the name of a VirtualMachine living in the cluster which will be the reference from which the template will be created (required field).
+* **VMNamespace**: the namespace of the reference VM. If not specified the VM will be looked up in the namepace in which VirtualMachineTemplateRequest was created (optional field).
+* **TemplateName**: the name of the template object that will be generated, if omitted it will be auto-generated (otional field).
+* **Format**: the format of the output template (required field).
+
+The first implemented format will be the [OpenShift Template(https://docs.openshift.com/container-platform/4.9/openshift_images/using-templates.html#templates-writing_using-templates)] format.
+
+In order to support multiple formats, the controller will have to be extended to generate the correct output for those formats.
```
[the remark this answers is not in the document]
acardace: Sure, but if you agree this is just an implementation detail.

design-proposals/template-generation.md
acardace: Sorry didn't know how to better put it into words, what I mean is that leaf fields called `name` won't be parametrized as they are just names and not real VM parameters.

design-proposals/template-generation.md
acardace: `VMName` is a required field and it references the VM from which the template will be generated, here I'm talking about the `metadata.name` field itself which will be converted into a `metadata.generateName`.

design-proposals/template-generation.md
acardace: OpenShift templates don't have a way (as far as I know) to do the same for annotations, that's why I didn't include them, you may argue that this design is supposed to be generic and it is but since the first format to be implemented will be OCP I've not added it as a tradeoff in favor of simplicity.

design-proposals/template-generation.md
```diff
@@ -0,0 +1,131 @@
...
+kind: VirtualMachineTemplateRequest
+metadata:
+  name: virtualmachinetemplaterequest-sample
+spec:
+  vmName: vmi-ephemeral
+  templateName: vmi-ephemeral-template
+  format: template.openshift.io/v1
+```
+
+The following template will be created
```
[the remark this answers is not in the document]
acardace: Sure.

design-proposals/template-generation.md
```diff
@@ -0,0 +1,131 @@
...
+
+This design defines an approach to create Virtual Machine templates starting from an existing VM.
+
+## Motivation
+
+Creating a 'blueprint' of an existing VM to further tweak and use as a basis golden VM template is a common operation, also it is a tool that can be leveraged by UIs to improve the UX of the project.
+
+## Goals
+
+* Create a generic template of an existing VM.
```
fabiand, line 11: What is a `template` in Kubernetes?
Ain't there different template methods? Which of them is targeted by this proposal?

design-proposals/template-generation.md
```diff
@@ -0,0 +1,131 @@
...
+
+After posting this VirtualMachineTemplateRequest object
+
+```yaml
+apiVersion: ssp.kubevirt.io/v1alpha1
+kind: VirtualMachineTemplateRequest
+metadata:
+  name: virtualmachinetemplaterequest-sample
+spec:
+  vmName: vmi-ephemeral
```
fabiand, line 79: What about differentiating source and target more:

```
from:
  vmName: vmi-ephemeral
to:
  openshiftTemplate:
    name: vmi-ephemeral-template
``

The benefit of a dedicated struct instead of a format value is, is that the API is "future proof", we can easily add format specific fields in future and keepin backwards compatibility.

File: design-proposals/template-generation.md [hunk 15]
```diff
@@ -0,0 +1,131 @@
...
+
+The following template will be created
+
+```yaml
+apiVersion: v1
+kind: Template
+metadata:
+  name: vmi-ephemeral-template
+  labels: ${LABELS}
+objects:
```
fabiand, line 92: A source VM can include my stateful things
- secrets
- configmaps
- pvcs
- cloud-init
- sysprep
- …

What will happen to each of them?

We should find a generic rule of how stateful elements are converted into a template.

Maybe worth also to understand how oVirt does this - @borod108 might know this

design-proposals/template-generation.md
acardace: Sure thing.

design-proposals/template-generation.md
fabiand: Agreed. Maybe good to somehow hint at this by i.e. including a sentence around .. different mechanisms, because there is no default K8s way?

design-proposals/template-generation.md
stu-gott: I would find this very surprising. I would expect that requesting a template to be made would make a template only--no other objects. especially anything with storage impact.

design-proposals/template-generation.md
xpivarc: I think there are existing template tools that could benefit from generic template API that would serve for metadata purposes. Examples I could think of or find are:
1. [Blog](https://livewyer.io/blog/2021/02/23/kubevirt-showcase-a-kubernetes-cluster-within-a-kubernetes-cluster/ )- Using Helm for templating VM (See [implementation](https://github.com/livewyer-ops/kubevirt-demo/tree/v1.0.0/examples/ubuntu/multi-vmis))
2. [Kustomize](https://kustomize.io/) - You could want to create a kustomize "template". You would pick a VM (Lets say you are using in testing stage for new app) and create a version with parameters better suited production deployment.

Ansible or mentioned Openshift templates (and other tools) could greatly benefit from this. The benefit I see is that there would be one API that knows what needs to be replaced or stay untouched.

design-proposals/template-generation.md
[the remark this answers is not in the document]
xpivarc: I would therefore ask to highlight this is extensible and the above sentence is describing one of the approaches.

design-proposals/template-generation.md
xpivarc: I think [this](https://www.ovirt.org/develop/release-management/features/virt/template-versions.html) describes mentioned ovirt mechanism.

design-proposals/template-generation.md
xpivarc: I would probably write "except name fields which doesn't have any functional impact".

design-proposals/template-generation.md
xpivarc: Didn't notice it. Thanks

design-proposals/template-generation.md
xpivarc: I think ruling out annotation can affect functionality. See 
```
// This annotation is used to inject ignition data
	// Used on VirtualMachineInstance.
	IgnitionAnnotation           string = "kubevirt.io/ignitiondata"

`// This annotation is to keep virt launcher container alive when an VMI encounters a failure for debugging purpose
	KeepLauncherAfterFailureAnnotation string = "kubevirt.io/keep-launcher-alive-after-failure"`
```

PR #170, opened 2022-03-25
VM backend storage design proposal

jean-edouard:


design-proposals/vm-backend-storage.md
```diff
@@ -0,0 +1,68 @@
...
+
+## User Stories
+As a VM user, I want to be able to store the encryption key of my VMs partitions inside the vTPM (example: BitLocker)
+
+## Repos
+kubevirt/kubevirt
+
+# Design
+The storage used for this would be 1 PVC for the whole cluster.  
+This PVC would be 1GiB in size and it would be shared across virt-handlers.
```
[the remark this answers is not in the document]
jean-edouard: Yes I think so. Is that a concern, or just something I should be more specific about?
Thank you for the review!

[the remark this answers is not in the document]
alicefr: You might not have a storage class that support RWX PVCs

design-proposals/vm-backend-storage.md
```diff
@@ -0,0 +1,68 @@
...
+This feature is intended for end-users who wish to preserve the state of some emulated device across reboots.
+
+## User Stories
+As a VM user, I want to be able to store the encryption key of my VMs partitions inside the vTPM (example: BitLocker)
+
+## Repos
+kubevirt/kubevirt
+
+# Design
+The storage used for this would be 1 PVC for the whole cluster.
```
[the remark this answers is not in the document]
jean-edouard: The idea of 1 PVC per VM was brought up at the KubeVirt community meeting, and discarded for its high disk usage.
It was stated that storage backends tend to allocated a minimum of 1GiB per PV, which would be a massive waste of space if there was 1 per VM...

design-proposals/vm-backend-storage.md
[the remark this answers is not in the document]
alicefr: I'm not sure if this works. How do you support multiple users? Do you have a PVC per user?

design-proposals/vm-backend-storage.md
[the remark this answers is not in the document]
alicefr: @jean-edouard this is a very vague idea, but why don't we use [csi-hostpath](https://github.com/kubevirt/hostpath-provisioner) for doing a proper partitioning? At least this CSI driver is something maintained by KubeVirt community. 

@awels do we have a minimum size for the PVC generated by the hostprovisioner?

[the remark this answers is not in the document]
awels: So the hostpath provisioner generates PVs with capacity of the total size, which should match most PVC request sizes (note the difference between capacity and request sizes). Since the hostpath provisioner does not have anyway of enforcing size it just reports the total size (of which not everything will be available). The only provisioner I am aware of that allocates in PVs in 1Gi chunks is the cinder provisioner. ceph allows smaller sizes as well.

design-proposals/vm-backend-storage.md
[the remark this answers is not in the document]
jean-edouard: Sorry, by "user" I meant "cluster admin".
However, the discussion below might steer us back to 1 PVC per VM, which would alleviate the need for a CR option.

design-proposals/vm-backend-storage.md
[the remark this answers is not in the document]
jean-edouard: @awels that sounds good! In that case, we should be able to just create 1 per VM, simplifying things a lot. On migration, I assume we can move the hostpath to the target node?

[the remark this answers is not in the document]
awels: I am missing a piece here, when you say migration, you don't mean live migration, since that is not possible without RWX storage.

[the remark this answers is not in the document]
awels: Okay so if you want to live migrate a VM, you need RWX storage so it can be mounted on 2 nodes at once. RWO by definition cannot be mounted on 2 nodes at once. hostpath provisioner kind of by definition can only generate RWO volumes since it is just creating directories on a host. So you cannot live migrate.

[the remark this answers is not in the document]
jean-edouard: It's important to note that the storage is only for virt-launcher, it's not exposed to the guest, and libvirt doesn't know it exists.
So, theoretically, we can handle it however we want, like in a way that never has it mounted on 2 nodes at once...
The simplest thing would be to make a duplicate of it on migration, if that's a possible operation.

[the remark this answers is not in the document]
alicefr: > It's important to note that the storage is only for virt-launcher, it's not exposed to the guest, and libvirt doesn't know it exists. So, theoretically, we can handle it however we want, like in a way that never has it mounted on 2 nodes at once... The simplest thing would be to make a duplicate of it on migration, if that's a possible operation.

It will be interesting to understand how libvirt treats the vTPMs during migration. If it is copied over the network like local storage (in KubeVirt an example is the delta with containerdisks) or it expects that the data are available also on the destination node. If there is such a mechanism, then you can achieve live migration by creating a second PVC in the destination node and let libvirt copying the information. From your description, the data seems to be relatively small so I guess it shouldn't have a big impact on the performance.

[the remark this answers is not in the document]
jean-edouard: Yeah, the file is just copied over. It probably does it while the guest is down to avoid worrying about (unlikely) deltas, and since the file is so small, but I don't know for sure.
Creating a second PVC sounds just fine, we'd just have to figure out how to maintain the link between VM object and PVC.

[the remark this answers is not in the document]
alicefr: maybe you could set the ownerReference on the PVC to the VM object? If you do so, the PVC is automatically deleted when the VM object is deleted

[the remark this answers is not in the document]
jean-edouard: Great idea! Thanks for all the info/ideas, I will revise the proposal soon to include that.

design-proposals/vm-backend-storage.md
```diff
@@ -0,0 +1,73 @@
...
+
+## User Stories
+As a VM user, I want to be able to store the encryption key of my VMs partitions inside the vTPM (example: BitLocker)
+
+## Repos
+kubevirt/kubevirt
+
+# Design
+The storage used for this would be 1 PVC per VM, and would be created on first start of a VM that needs backend storage.  
+This PVC would be RWO, about 2 MiB in size (still TBD), and use [csi-hostpath](https://github.com/kubevirt/hostpath-provisioner).
```
xpivarc, line 32: This point is not clear to me. While we can copy most of the data to the new PVC on the migration, we simply would be limited by storage topology on restarts. I think we need to express these constrain or choose a different path.

design-proposals/vm-backend-storage.md
```diff
@@ -0,0 +1,73 @@
...
+## User Stories
+As a VM user, I want to be able to store the encryption key of my VMs partitions inside the vTPM (example: BitLocker)
+
+## Repos
+kubevirt/kubevirt
+
+# Design
+The storage used for this would be 1 PVC per VM, and would be created on first start of a VM that needs backend storage.  
+This PVC would be RWO, about 2 MiB in size (still TBD), and use [csi-hostpath](https://github.com/kubevirt/hostpath-provisioner).  
+The PVC would be added to the virt-launcher compute container as a volume, but never assigned to the libvirt domain (the guest VM).
```
xpivarc, line 33: I would suggest altering this sentence, but it will not be available to VirtualMachine. I think it is more common than libvirt domain...

design-proposals/vm-backend-storage.md
```diff
@@ -0,0 +1,73 @@
...
+As a VM user, I want to be able to store the encryption key of my VMs partitions inside the vTPM (example: BitLocker)
+
+## Repos
+kubevirt/kubevirt
+
+# Design
+The storage used for this would be 1 PVC per VM, and would be created on first start of a VM that needs backend storage.  
+This PVC would be RWO, about 2 MiB in size (still TBD), and use [csi-hostpath](https://github.com/kubevirt/hostpath-provisioner).  
+The PVC would be added to the virt-launcher compute container as a volume, but never assigned to the libvirt domain (the guest VM).  
+The PVC would be mounted in a directory at the root of the filesystem. Appropriate symbolic links will be created before starting the libvirt domain.
```
xpivarc, line 34: Can you add an example of such symlinks and maybe explain what is the motivation for it?

design-proposals/vm-backend-storage.md
```diff
@@ -0,0 +1,73 @@
...
+The PVC would be owned-by the VM, and therefore will get deleted when its associated VM is deleted.
+
+## API Examples
+Currently, the TPM device API doesn't have any option.
```
xpivarc, line 38: I think we should state here that there will be no dedicated API field for this and features such as persistent TPM will transparently use the storage.

Edit: Will there be kubevirt option to provide the storage class for the storage?

design-proposals/vm-backend-storage.md
```diff
@@ -0,0 +1,73 @@
...
+The PVC would be added to the virt-launcher compute container as a volume, but never assigned to the libvirt domain (the guest VM).  
+The PVC would be mounted in a directory at the root of the filesystem. Appropriate symbolic links will be created before starting the libvirt domain.
+The PVC would be owned-by the VM, and therefore will get deleted when its associated VM is deleted.
+
+## API Examples
+Currently, the TPM device API doesn't have any option.  
+We will add a `persistent` boolean option that will default to false.  
+When set to true, the state of the TPM will be persisted.
+
+## Scalability
```
xpivarc, line 42: I think it's good to state here that some CSIs may not support small volumes.

design-proposals/vm-backend-storage.md
```diff
@@ -0,0 +1,73 @@
...
+Every VM will take a couple more MiBs of storage at most, which shouldn't be a concern, even on cluster running a large number of VMs.  
+For now, the only 2 candidates features for backend storage are:
+- The persistent TPM state, which takes about 6KiB when freshly created. We have yet to find its maximum size.
+- The OVMF vars, which can take up to 528KiB
+I propose finding the biggest size these can take, multiply the sum by 2 and round up to the nearest MiB to allow for future use-cases.
+
+## Update/Rollback Compatibility
+VM migration will have to be specifically handled.
```
xpivarc, line 50: Does this mean old VMs will get the storage transparently?

design-proposals/vm-backend-storage.md
jean-edouard: I agree, the problem is that we also create a k8s object called VirtualMachine, so I tried to make it as obvious as possible... Maybe just "guest VM"?

design-proposals/vm-backend-storage.md
jean-edouard: I actually don't like the symlinks idea, but that's the only way I could think of to make the files appear to be in the right place.
When there's a libvirt domain option to override the location we can just point to the actual file, but for example TPM doesn't seem to have such an option...
If you have other ideas I'm all for it, for example I've seen overlay2-mounted files, not directories, inside containers. I don't understand how that works, since you're only supposed to be able to mount directories, but maybe that mechanism could be used to our advantage?...

design-proposals/vm-backend-storage.md
jean-edouard: Right, there will be a `persistent` bool, but no explicit "enable backend storage" option.
Adding a KubeVirt option to provide the storage class is definitely a possibility.

design-proposals/vm-backend-storage.md
jean-edouard: Yes, will do!

design-proposals/vm-backend-storage.md
jean-edouard: Yes, if they are modified to use options that leverage backend storage.
Existing VMIs are (mostly) immutable and therefore won't contain any option that require storage.
I can add some language for that.

design-proposals/vm-backend-storage.md
jean-edouard: You lost me here... What is the concern about restarts?

design-proposals/vm-backend-storage.md
xpivarc: Naming is hard, I am not sure what to imagine with `guest VM`. Maybe `user`s VM`?

design-proposals/vm-backend-storage.md
xpivarc: I think Alexander explained it well https://github.com/kubevirt/community/pull/170#discussion_r841923983 . Basically, RWO doesn't guarantee that you can attach the volume to another node. Therefore the volume will be available only on the particular node and the VMI will need to be scheduled there when you stop-start the VM.

design-proposals/vm-backend-storage.md
xpivarc: I would appreciate it, thanks!

design-proposals/vm-backend-storage.md
jean-edouard: Right, but after that, Alice suggested creating new PVCs on migration, which I think is the way to go.
That way we never have to mount it in 2 places at once. I guess I did not make that obvious enough in the document, will review.

design-proposals/vm-backend-storage.md
jean-edouard: @xpivarc did you get a chance to test your idea? Is that something I could help with? Thanks!

xpivarc: My idea was to use multiple moutPaths for one volume. Would that help?

jean-edouard: Yeah, but would that require RWX? I guess it would have the undesirable side effect of showing all files in all mount points, but I still prefer that over symlinks.
We can also simplify all this by having 1 volume per feature/mountpoint and be done with it (e.g. 1 for TPM, 1 of OVMF, ....). That's probably our best bet when I think about it.

xpivarc: I think the RWX is a must for another reason https://github.com/kubevirt/community/pull/170#discussion_r848081613. 

The idea would not require RWX afaik. You could use subPaths and limit the view. I think the best would be to try it out. Unfortunately, I didn't try it yet.

design-proposals/vm-backend-storage.md
jean-edouard: I'm not sure I understand the question/concern... Once a VM is stopped, its VMI is gone, the PVC is not in use anymore and it can be used by the new VMI next start, even if it's just RWO, no?

alicefr: Certain RWO PVC once their are bound they cannot be attached to another node. This is depends on the storage provider you are using. Then, the VMs that uses that PVC can run only on the node where the PVC has been bound. However, this is a general issue for all the cases where the VM has a RWO PVC

xpivarc: Exactly as Alice pointed out. I don't understand why we should take this limitation rather than require RWX. I would argue we can have simpler logic too if we require RWX.

design-proposals/vm-backend-storage.md
```diff
@@ -0,0 +1,82 @@
+# Overview
+Today, anything that happens on the host side of VMIs (i.e. virt-launcher) is transient.  
+Any change a VMI makes outside of its own file-system is lost after a shutdown.  
+Such changes include EFI firmware settings and information stored inside the TPM.
```
vladikr, line 4: Can you please give an example of such settings?

design-proposals/vm-backend-storage.md
```diff
@@ -0,0 +1,82 @@
...
+
+For now, the only 2 candidates features for backend storage are:
+- The persistent TPM state, which takes about 6KiB when freshly created. We have yet to find its maximum size.
+- The OVMF vars, which can take up to 528KiB
+I propose finding the biggest size these can take, multiply the sum by 2 and round up to the nearest MiB to allow for future use-cases.
+The total will likely add up to 2MiB, 3MiB tops.
+
+## API Examples
+Currently, the TPM device API doesn't have any option.  
+We will add a `persistent` boolean option that will default to false.
```
iholder-redhat, line 45: I'm wondering whether the default should be false.
Isn't it a bit confusing that by default the TPM is non-functional? Perhaps it will cause workloads to fail and force the user to find this spec field.

design-proposals/vm-backend-storage.md
jean-edouard: Mostly the boot entries. When installing an operating system, an entry gets added to the boot manager with the EFI module to boot and its parameters.
Luckily, Windows and most Linux distros install their EFI module to a standard location, and it doesn't need any parameters, so they are still bootable after the entry is gone.
However, if a user wanted to compile their kernel directly as an EFI module and skip shim+grub, they would not be able to customize the command line without recompiling the kernel.

Other options include:
- Boot order
- System language
- Default display resolution
- Secureboot state/keys

Sorry for the delay!

design-proposals/vm-backend-storage.md
jean-edouard: Good point, but I'm not sure how to handle upgrades from previous versions then... Would we add "persistent: false" to all existing VM(I)s on upgrade?

iholder-redhat: Sounds good to me :)

jean-edouard: I actually disagree about defaulting to true. Enabling persistence requires setting up a RWX FS storage class and adding it to the kubevirt CR. Then the extra PVC costs storage and time, especially on migrations.
In other words, you probably don't want TPM persistence enabled unless you really need it.
Defaulting to false would also save a lot of headaches to people who already wrote yaml files for their VM(I)s and expect them to behave the same after upgrade...

iholder-redhat: Alright, I'm convinced :)
Just one last thought - maybe we should at least warn about vTPM not being persistent? My concern is that something inside the guest would assume it has a valid TPM therefore things would break after reboot. The VM owner might get confused, IOW it wouldn't be obvious why things broke. Maybe we can add another condition, or event? WDYT?

design-proposals/vm-backend-storage.md
rmohr: Subsequent boots are also significant faster when EFI vars are persisted.

design-proposals/vm-backend-storage.md
```diff
@@ -0,0 +1,81 @@
...
+- Start the VM
+- Migrate the VM if the cluster supports migration
+- Unseal the string and ensure it is correct
+- Stop and delete the VM
+- Ensure the backend storage PVC gets deleted along with the VM
+
+# Implementation Phases
+- Automatically create the PVC on VM creation
+- Implement a routine that creates the right symlinks
+- Add the right PVC handling functions to the migration code
```
vladikr, line 75: @jean-edouard is this still required?

vladikr (commented):
This proposal looks good to me.
The code is already implemented and mostly approved.
/approve
/lgtm
/hold

If there are no objections in the coming days, let's remove the hold.

rmohr:
In  a follow-up I think it may make sense to address block storage here: Like on most storage providers, filesystem PVCs and block PVCs are not streamed through the same set of infrastructure (typically filesystem is on top), distributing the PVCs across different storage types with different availability. It therefore may make sense to come up with a follow-up design to allow staying completely on block for TPM and EFI vars.

design-proposals/vm-backend-storage.md
jean-edouard: Indeed we don't, we also don't use symlinks. Updated, PTAL!

xpivarc (approved):

design-proposals/vm-backend-storage.md
jean-edouard: @rmohr do you know why boots are faster when EFI vars are persisted? Maybe thanks to the bootmgr entry that avoids browsing every EFI system partition? It still doesn't feel like that should make a perceptible difference...

design-proposals/vm-backend-storage.md
```diff
@@ -0,0 +1,81 @@
...
+## Definition of Users
+This feature is intended for end-users who wish to preserve the state of some emulated device across reboots.
+
+## User Stories
+As a VM user, I want to be able to store the encryption key of my VMs partitions inside the vTPM (example: BitLocker)
+
+## Repos
+kubevirt/kubevirt
+
+# Design
```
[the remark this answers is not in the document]
jean-edouard: Added

jean-edouard:
@xpivarc PTAL!

enp0s3 (commented):
@xpivarc @jean-edouard Hi, I was trying to get familiar with the backend-storage support and going back in time brought me to this design.
I am missing something here, lets say I have a running VM with TPM. Now I stop the VM, I upgrade kubevirt to this proposal, and I start the VM. Will it be rejected because I have nothing configured in the KV CR regarding the storage class?
Or because we decided to default it to false, I will manage to start the VM?
How can I then enable the persistent mode?

jean-edouard:
> @xpivarc @jean-edouard Hi, I was trying to get familiar with the backend-storage support and going back in time brought me to this design. I am missing something here, lets say I have a running VM with TPM. Now I stop the VM, I upgrade kubevirt to this proposal, and I start the VM. Will it be rejected because I have nothing configured in the KV CR regarding the storage class? Or because we decided to default it to false, I will manage to start the VM? How can I then enable the persistent mode?

Yes, persistence defaults to false for that reason. To enable persistence, just set persistent=true.
Note: backend-storage now works without a storage class set in the CR.

PR #172, opened 2022-04-13
Update membership-policy to specify valid OWNERS files in sponsor requirements

alromeros:
Hello everyone,

This PR aims to introduce a small update in the membership policy so the requirements for sponsors are better explained.

Best regards.

### DIFF
--- a/membership_policy.md
+++ b/membership_policy.md
@@ -38,7 +38,7 @@ Defined by: Member of the KubeVirt GitHub organization
     * Contributions according to GitHub and/or Kubernetes metrics in the past 90 days.
   * Sponsored by 2 org members. Note the following requirements for sponsors:
     * Sponsors must have close interactions with the prospective member
-    * Sponsors must be reviewers or approvers in the KubeVirt OWNERS file.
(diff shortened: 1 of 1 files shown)


jean-edouard (commented):
So if you're a reviewer/approver in any `github.com/kubevirt/` repo you can be a sponsor on all the repos? That seems a bit too permissive to me...
I think it would make more sense for reviewers/approver of a given repo to be allowed to sponsor others only on that same repo.
What do you think?

alromeros:
Hi @jean-edouard! I agree that it may be a little too permissive, however, I understand that those sponsors are just for allowing joining the organization as a member, which is unrelated to individual repositories. Correct me If I'm wrong, but in that case, I think It wouldn't make that sense to limit sponsorships to be on the same repo.

jean-edouard (commented):
@alromeros my bad, I wrongly assumed we were in the `reviewers` section.
/approve

cwilkers:
Thanks, this is a good clarification of the policy

/approve
/lgtm

PR #180, opened 2022-06-13
proposal: Introduce e2e Framework

fossedihelm:
This design proposal introduce the Framework that will support
 e2e test architecture and common operations.

There is already an ongoing discussion in https://github.com/kubevirt/kubevirt/pull/7188

Signed-off-by: fossedihelm <ffossemo@redhat.com>

### DIFF
--- a/design-proposals/e2e-framework.md
+++ b/design-proposals/e2e-framework.md
@@ -1,0 +1,92 @@
+Framework for E2E Tests
+=
+
+# Overview
+This proposal is suggesting the introduction of a `Framework` to support common operations in e2e tests.
+
+# Motivation
+
+Kubevirt has an impressive amount of e2e test, covering a large amount scenarios and features.
+
+Two paradigms can be used to make each test run on a clean environment:
+- restore a clean situation before the execution of each test.
+- make each test leave a clean situation.
+
+Currently, in kubevirt the first option is most commonly used, but specific cleanups are also
+performed after test execution. This mix can create leftovers and sometimes makes it difficult
+to understand where a problem may have occurred.
+
+It is also a symptom of a kind of non-independence of the tests from each other:
+test A does not care about cleaning the cluster, so much as it knows that there will be
+the setup of test B will do it.
+
+Last but not least, each test file must specify the cleanup and setup function on its own, and in most cases
+is something like this:
+```go
+BeforeEach(func() {
+  tests.BeforeTestCleanup()
+  
+  virtClient, err = kubecli.GetKubevirtClient()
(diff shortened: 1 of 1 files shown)


design-proposals/e2e-framework.md
```diff
@@ -0,0 +1,92 @@
...
+## User Stories
+
+As an e2e test author:
+
+I want to create a new e2e test file ensuring that there will be no leftover that may influence subsequent testing.
+
+## Repos
+Kubevirt/kubevirt
+
+## Design
```
iholder-redhat, line 64: What about separating tests to different namespaces? This would ensure that even if the environment is not cleaned properly, other tests are less likely to be affected.

It probably doesn't make sense to create a different namespace for each test (or is it? :o), but maybe a namespace per 10 tests (or some other constant)? Maybe per test-suite?

design-proposals/e2e-framework.md
```diff
@@ -0,0 +1,92 @@
...
+|----------------------|:--------------------------|:----------------------------------------------|
+| Environment setup    | SynchronizedBeforeSuite() | Once at the beginning of the whole test suite |
+| Test setup           | BeforeEach()              | Before each test execution                    |
+| Test execution       |                           |                                               |
+| Test teardown        | AfterEach()               | After each test execution                     |
+| Environment teardown | SynchronizedAfterSuite()  | Once at the end of the whole test suite       |
+
+The `Framework` supports steps 2 and 4, freeing the test files from all the common steps of preparing and
+cleaning the environment. Note that steps 2 and 4 will not disappear in the test files, but will contain
+only the logic related to that specific set of tests.
```
iholder-redhat, line 83: Maybe also explain the cleaning mechanism as well?
I really liked the "CleanableResource" idea from your original PR, maybe it can be discussed here as well

iholder-redhat (commented):
Thanks @fossedihelm!
Happy to see this, I think it's a great path forward for us.

design-proposals/e2e-framework.md
iholder-redhat: If that's possible from a performance/scalability perspective, I don't see a reason why not! :smiley:

awels: Note CDIs framework does exactly this, create a new namespace per test. We don't wait for the namespaces to be deleted before creating a new one. We delete the namespace and then move on without waiting.

fossedihelm: Thanks @awels, this could be an idea, since my tests show that deleting a namespace requires around 5 secs.

PR #181, opened 2022-06-23
Proposal:  Expose host-model migratability level of each node

Barakmor1:
Add a label to determine the host-model migratability level of each node
and use this label to configure the cluster to prefer\require node with
high host-model migratability level when scheduling vm for the first time.

Things I'm not sure about:
- Maybe we would like to apply the configuration to vms with some label
  or apply every rule to different sets of vms.
- What if some one want to **prefer** nodes and **require** nodes with some 
host-model migratability level at the same time.
Should we add affinity for the preference and the requirement?

Signed-off-by: bmordeha <bmodeha@redhat.com>

### DIFF
--- a/design-proposals/determining-host-model-migratability-level.md
+++ b/design-proposals/determining-host-model-migratability-level.md
@@ -1,0 +1,179 @@
+
+# Overview
+Add a label to determine the host-model migratability level of each node 
+and use this label to configure the cluster to prefer\require node with
+high host-model migratability level when scheduling vm for the first time.
+## Motivation
+In KubeVirt, the default CPU model is `host-model` .
+
+For host-model VMs the CPU model is being determined by the first node on which the VM 
+lands on. This means that the initial node a VM runs on significantly limit the 
+possible destinations for future migrations.
+
+If a vm with host-model CPU model is scheduled to a node with a unique `host-model-required-feature`
+in heterogeneous cluster it won't be able to migrate and this will probably **SURPRISE** the user.
+
+It isn't a trivial task for a user to calculate the set of nodes that 
+a vm with `host-model` can migrate to (for each node) especially when the cluster change 
+and nodes are being added\removed to\from the cluster.
+
+Even if a user has this information he would have to set nodeAffinity himself to prefer/require
+nodes only for the first time he schedules a vm.
+
+Also, some users\developers might want to use our testing framework in heterogeneous cluster,
+and we usually never check if a host-model migration is even possible before we migrate vms in tests.
+
+Therefore, some users might benefit from such label and configuration options.
+
+## Goals
+- Add a label to each schedulable node in the cluster with a value of
+the relative amount of nodes that a vm with `host-model` CPU model can
+migrate to when the node is the initial node of the vm.
+
+
+- Allow users to configure KubeVirtCR to add node affinity for host-model 
+migratability level to virt-launcher when scheduling a vmi for the first time.
+
+## Non Goals
+Not changing the default form in which we schedule virtual machines.
+
+## Definition of Users
+Every user that is interested of migration of vms in heterogeneous cluster.
+
+## User Stories
+- As a user with a long living heterogeneous cluster, I'm **SURPRISED** to encounter host-model migration failures and
+i don't know how to prevent the failures.
+
+
+- As a user that use KubeVirt's tests in heterogeneous cluster i don't know why the tests are flacky:
+
+    In KubeVirt-ci the cluster is homogeneous, so we aren't exposed to test failures that
+occur because of a host-model migration failures.
+
+  For instance Openshift-Virtualization project run our functional tests in heterogeneous
+clusters this issue cause flakiness, and it is hard to follow.
+
+  
+Such configuration option that prefer/require host-model migratible nodes can solve this.
+
+## Repos
+https://github.com/kubevirt/kubevirt
+
+# Design
+- List and calculate the value of the relative amount of nodes that a vm with host-model
+  CPU model can be scheduled to for each node  when adding\removing a node from the cluster.
+
+  This can be done through the `node-controller` which is 
+part of `virt-controller`. 
+
+
+- Add an optional fields to `kubevirt.spec.configuration.migrations` to determine if we want our vm's to prefer\require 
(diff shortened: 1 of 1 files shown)


design-proposals/determining-host-model-migratability-level.md
```diff
@@ -0,0 +1,123 @@
+This template is simply a guide. When making proposals, feel free to add and
+remove portions of this outline to best fit your specific case. There is no
+expectation that people strictly stick to this outline, but it is a good place
+to start.
```
iholder-redhat, line 4: leftover - can be removed

design-proposals/determining-host-model-migratability-level.md
```diff
@@ -0,0 +1,123 @@
...
+
+It isn't a trivial task for a user to calculate the set of nodes that 
+a vm with `host-model` can migrate to (for each node) especially when the cluster change 
+and nodes are being added\removed to\from the cluster.
+
+Even if a user has this information he would have to set nodeAffinity himself to prefer/require
+nodes only for the first time he schedule a vm.
+
+Therefore some users might benefit from such label and configuration options.
+
```
iholder-redhat, line 22: Maybe you can also add a motivation for testing. Today we only test homogeneous clusters (partly because of the problem you're raising here), with this label we could extend our tests into a heterogeneous environment easily.

design-proposals/determining-host-model-migratability-level.md
```diff
@@ -0,0 +1,123 @@
...
+remove portions of this outline to best fit your specific case. There is no
+expectation that people strictly stick to this outline, but it is a good place
+to start. 
+
+# Overview
+Add a label to determine the host-model migratability level of each node 
+and use this label to configure the cluster to prefer\require node with
+high host-model migratability level when scheduling vm for the first time.
+## Motivation
+If a vm with host-model cpuModel is scheduled to a node with a unique `host-model-required-feature`
```
iholder-redhat, line 11: For the ones that aren't familiar with the host-model mechanism, maybe it's worth adding some background. Maybe something like:

_In Kubevirt, the default CPU model is "host-model" <LINK TO LIBVIRT DOCUMENTATION>.
For host-model VMs the CPU model is being determined by the first node on which the VM lands on. This means that the first node a VM runs on can significantly impact the possible destination for future migrations._

design-proposals/determining-host-model-migratability-level.md
```diff
@@ -0,0 +1,123 @@
...
+and nodes are being added\removed to\from the cluster.
+
+Even if a user has this information he would have to set nodeAffinity himself to prefer/require
+nodes only for the first time he schedule a vm.
+
+Therefore some users might benefit from such label and configuration options.
+
+## Goals
+- Add a label to each schedulable node in the cluster with a value of
+the relative amount of nodes that a vm with `host-model` cpuModel can
```
iholder-redhat, line 25: I would replace `cpuModel` ->  `CPU model`.
Relevant to all of this file

design-proposals/determining-host-model-migratability-level.md
```diff
@@ -0,0 +1,123 @@
...
+affinity:
+  nodeAffinity:
+    requiredDuringSchedulingIgnoredDuringExecution:
+      nodeSelectorTerms:
+        - matchExpressions:
+          - key: kubevirt/host-model-migratability-level
+            operator: Gt
+            values:
+            - <some-numerical-number>
+```
```
iholder-redhat, line 91: This is more of an implementation detail than an API example. Please show an example of the Kubevirt CR new API here.

You can talk about the affinity implementation detail in "implementation phases" or "design"

design-proposals/determining-host-model-migratability-level.md
```diff
@@ -0,0 +1,123 @@
...
+
+
+- add fields to kubevirtCR:
+`kubevirt.spec.configuration.migrations.requireHostModelMigratabilityLevel:<some-number> `
+`kubevirt.spec.configuration.migrations.preferedHostModelMigratabilityLevel:<some-number> `
```
iholder-redhat, line 96: Maybe this can be shortened to `requireMigratabilityLevel`?

I think that most of the users don't know about host-model (which is the default). This setting would not apply only for VMs that explicitly set their CPU model to a concrete model, and these are usually advanced users that know the implications.

design-proposals/determining-host-model-migratability-level.md
```diff
@@ -0,0 +1,123 @@
...
+
+
+
+## Scalability
```
iholder-redhat, line 100: You can maybe also explain that this complexity is because the work is being done by one Pod (virt-controller) for all nodes, and the reasons it's like that (very shortly)?

design-proposals/determining-host-model-migratability-level.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Functional Testing Approach
+This will allow us to require\require `host-model` migratible nodes in the functional tests
+and make our functional tests more heterogeneous friendly.
+
+# Implementation Phases
+
+- Add the label to each node through the `node-controller`.
+
+- Add new fields to kubevirtCR and implement the desired behavior.
```
iholder-redhat, line 121: I would implement these two in a single PR

iholder-redhat (commented):
Thanks @Barakmor1! Very happy to see this!
I find this very valuable and useful.

design-proposals/determining-host-model-migratability-level.md
Barakmor1: I think that this might mislead users to think that these configuration are valid to ANY CPU Model.

design-proposals/determining-host-model-migratability-level.md
Barakmor1: Thanks for reviewing!
Done.

design-proposals/determining-host-model-migratability-level.md
Barakmor1: Done.

design-proposals/determining-host-model-migratability-level.md
Barakmor1: Done

PR #185, opened 2022-08-02
proposal: SR-IOV Interfaces Assignment Fix 

ormergi:
This is a design proposal for fixing Kubevirt's SR-IOV interfaces assignment logic, following kubevirt/kubevirt#7444.

*When reviewing use the option to `view file` from the menu on the upper right hand to view the markdown formatting.

Signed-off-by: Or Mergi <ormergi@redhat.com>

### DIFF
--- a/design-proposals/sriov-interfaces-mapping.md
+++ b/design-proposals/sriov-interfaces-mapping.md
@@ -10,7 +10,7 @@ by passing its PCI device ID from the node, through the virt-launcher pod, and t
 Each VM SR-IOV device (i.e: VF) is associated with a VM secondary network.
 
 When a VM is defined with multiple SR-IOV devices from the same [resource pool](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#configurations)
-or connected to the same NetworkAttachmentDefinition, Kubevirt cannot differentiate which VF is associated with which requested device.
+or connected to the same `NetworkAttachmentDefinition`, Kubevirt cannot differentiate which VF is associated with which requested device.
 The root cause is that the PCI information passed by the sriov-network-device-plugin (through kubelet) is not enough to directly map the VFs with their secondary networks.
 
 The outcome is a non-consistent association between the underlying VFs and the VM networks,
@@ -18,7 +18,7 @@ leaving the VM in a state where the VFs are plugged into the domain wrongly.
 <br/>
 For example:
 <br/> 
-When a SR-IOV interface is defined with a custom MAC address (e.g: `02:00:00:00:00`), a custom guest PCI address of (e.g: `0000:01:01.1`) and pointing to a NetworkAttachmentDefinition with VLAN configuration (e.g: 100).
+When a SR-IOV interface is defined with a custom MAC address (e.g: `02:00:00:00:00`), a custom guest PCI address of (e.g: `0000:01:01.1`) and pointing to a `NetworkAttachmentDefinition` with VLAN configuration (e.g: 100).
 <br/>
 It is not guaranteed that the correct VF (with VLAN 100) will be plugged into the domain on the specified PCI address (`0000:01:01.1`).
 
@@ -40,15 +40,13 @@ Fix/change third-party components (e.g: Multus, SR-IOV CNI, etc..)
 ## Existing Implementation Flow
 In order to discover and allocate VFs to Pods the [sriov-network-device-plugin](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#sr-iov-network-device-plugin) (from now on sriov-dp) and [sriov-cni](https://github.com/k8snetworkplumbingwg/sriov-cni) are being used.
 <br/>
-The sriov-dp provides an API to create resource pools of VFs according to various properties [[1]](https://github.com/openshid/sriov-network-device-plugin#configurations).
-<br/>
-Each resource pool is labeled with a resource name.
+The sriov-dp provides an API to create resource pools of VFs according to various properties [[1]](https://github.com/openshid/sriov-network-device-plugin#configurations), and each resource pool is labeled with a resource name.
 
-In the VMI spec, secondary networks point to the desired SR-IOV NetworkAttachmentDefinition object.
+In the VMI spec, secondary networks point to the desired SR-IOV `NetworkAttachmentDefinition` object.
 <br/>
-The NetworkAttachmentDefinition is set with the resource-name annotation:
+The `NetworkAttachmentDefinition` is set with the resource-name annotation:
 <br/>`k8s.v1.cni.cncf.io/resourceName=<resource name>`<br/>
-which later on passed down by Multus to the sriov-cni [[2]](https://github.com/k8snetworkplumbingwg/multus-cni/blob/a28f5cb56c79a582f5ea2b35a61b38f34b937930/examples/README.md#passing-down-device-information).
+which later on [passed down by Multus](https://github.com/k8snetworkplumbingwg/multus-cni/blob/a28f5cb56c79a582f5ea2b35a61b38f34b937930/examples/README.md#passing-down-device-information) to the sriov-cni.
 ```yaml
 kind: VirtualMachine
   ...
@@ -93,12 +91,12 @@ Given the resource name `kubevirt.io/sriov_net` the following environment variab
 As part of the VM create flow, virt-controller renders the given VMI spec and creates the corresponding virt-launcher Pod [[3]](https://github.com/kubevirt/kubevirt/blob/1a8f08103e48c5f2bb2f5826d118507ce7ec1f0c/pkg/virt-controller/services/template.go#L443) (renderLaunchManifest).
 <br/>
 For each secondary network:
-- Realize the device `resourceName` by fetching the specified NetworkAttachmentDefinition.
+- Realize the device `resourceName` by fetching the specified `NetworkAttachmentDefinition`.
 - Add the following environment variable to the compute container (spec.Env):<br/>`KUBEVIRT_RESOURCE_NAME_<network name>=<resource name>`
 
 Once virt-launcher Pod is ready, virt-handler process the VMI object and eventually trigger the virt-launcher to synchronize with the new spec (i.e: virt-handler sends gRPC call to virt-launcher [[4]](https://github.com/kubevirt/kubevirt/blob/1a8f08103e48c5f2bb2f5826d118507ce7ec1f0c/pkg/virt-launcher/virtwrap/manager.go#L822)).
 <br/>
-virt-launcher renders the VMI spec and converts it to Libvirt domain XML, and eventually passes it to Libvirt which in turn create the domain.
+virt-launcher renders the VMI spec, converts it to Libvirt domain XML and eventually passes it to Libvirt, which in turn creates the domain.
 
 As part of the VMI spec rendering, virt-launcher realizes the VFs PCI addresses that been allocated by loading `KUBEVIRT_RESOURCE_NAME` (formerly created by the virt-controller) and `PCIDEVICE` (created by the sriov-dp) environment variables.
 
@@ -154,7 +152,7 @@ KUBEVIRT_RESOURCE_NAME_sriovnet1=kubevirt.io/sriov_net
 
 ## The Problem
 The VFs PCI address information that being passed down to virt-launcher through `PCIDEVICE` (by sriov-device-plugin) and `KUBEVIRT_RESOURCE_NAME` (by Kubevirt) environment variables, is not enough to map the VFs with their secondary network.
-In the scenario where a VM has more than one network that points to a NetworkAttachmentDefinition with the same resource name, VFs might get assigned to the wrong network.
+In the scenario where a VM has more than one network that points to a `NetworkAttachmentDefinition` with the same resource name, VFs might get assigned to the wrong network.
 
 For example:
 In the scenario described in https://github.com/kubevirt/kubevirt/issues/6351,
@@ -198,7 +196,7 @@ networks:
   ...
 ```
 
-2. Two NetworkAttachmentDefinition objects pointing to the same resource:
+2. Two `NetworkAttachmentDefinition` objects pointing to the same resource:
 ```yaml
 ---
 apiVersion: k8s.cni.cncf.io/v1
@@ -292,8 +290,8 @@ When `HostDevice` object is created, the PCI address is picked as follows:
 There is no guarantee that `0000:04:03.6` is the correct PCI address for the network `sriovnet100`  and that the VF is configured with VLAN 100.
 This example shows that the current environment variable `PCIDEVICE` (given by the sriov-dp) is simply not enough to directly map between the VM’s networks and  the VFs PCI address.
 
-### Example 2: Two VM networks  connected to the same NetworkAttachmentDefinition
-VMI spec with two SR-IOV devices, each connected to different networks, that are attached to the NetworkAttachmentDefinition (different MAC addresses):
+### Example 2: Two VM networks  connected to the same `NetworkAttachmentDefinition`
+VMI spec with two SR-IOV devices, each connected to different networks, that are attached to the `NetworkAttachmentDefinition` (different MAC addresses):
 
 ```yaml
 kind: VirtualMachineInstance
@@ -321,7 +319,7 @@ spec:
   ...
 ```
 
-Single NetworkAttachmentDefinition:
+Single `NetworkAttachmentDefinition`:
 ```yaml
 ---
 apiVersion: k8s.cni.cncf.io/v1
@@ -356,7 +354,7 @@ KUBEVIRT_RESOURCE_NAME_sriovnet-vlan100=kubevirt.io/sriov_net
 ```
 
 #### Summary
-In this example we can see that both NICs points the same NetworkAttachmentDefinition, but since they have different MAC addresses, there is no way to know which PCI address holds which MAC address.
+In this example we can see that both NICs points the same `NetworkAttachmentDefinition`, but since they have different MAC addresses, there is no way to know which PCI address holds which MAC address.
 
 > **_Note_**: In order to pass-trough SR-IOV VF to a VM, the VF should be configured to use the vfio-pci driver [[1]](https://kubevirt.io/user-guide/virtual_machines/interfaces_and_networks/#sriov).
 > The vfio-pci driver is an userspace driver, when a VF is bound to it, the VF is no longer recognized by the kernel and therefore will not be presented by iplink (‘ip a’ command).
--- a/design-proposals/sriov-interfaces-mapping.md
+++ b/design-proposals/sriov-interfaces-mapping.md
@@ -1,5 +1,5 @@
 # Overview
-When creating a VM with SRI-IOV interfaces, the VFs from the same resource pool are not differentiated by KubeVirt and therefore potentially wrongly plugged to the domain. 
+When creating a VM with SR-IOV interfaces, the VFs from the same resource pool are not differentiated by KubeVirt and therefore potentially wrongly plugged to the domain. 
 </br>
 This design proposes to fix the current mapping algorithm, using information from multus annotations.
 
--- a/design-proposals/sriov-interfaces-mapping.md
+++ b/design-proposals/sriov-interfaces-mapping.md
@@ -366,18 +366,18 @@ In this example we can see that both NICs points the same NetworkAttachmentDefin
 The proposal is based on information multus-cni provides as an annotation on the virt-launcher pod. The objective is to pass down the relevant information to the virt-launcher application, so it in turn can consume it and create the correct domain configuration.
 
 In order to pass the VM network to VF PCI address mapping from the pod to the virt-launcher application, the following step are to be taken:
-- Process the `k8s.v1.cni.cncf.io/networks-status` annotation, and compose a structure that contains one-to-one mapping between a (VMI) network and the device PCI address.
+- Process the `k8s.v1.cni.cncf.io/network-status` annotation, and compose a structure that contains one-to-one mapping between a (VMI) network and the device PCI address.
 - Create an annotation with the mapping information (e.g: `kubevirt.io/network-to-pci-address`) on the virt-launcher pod. <br/> This annotation is using the downward API to make it available to the application.
 - Once the virt-launcher application requires the network-to-pci information, it can read the content. <br/>virt-launcher will use this data when creating the domain for the first time and when hotplug-ing the devices.
 
 The following sections go into more details.
 
-## Utilizing `k8s.v1.cni.cncf.io/networks-status` annotation
+## Utilizing `k8s.v1.cni.cncf.io/network-status` annotation
 This annotation includes the information required to do one-to-one mapping between a VM interface and VF PCI address:
 
 ### Example
 ```yaml
-k8s.v1.cni.cncf.io/networks-status: |-
+k8s.v1.cni.cncf.io/network-status: |-
     [
         {
             "name": "kindnet",
@@ -419,18 +419,18 @@ k8s.v1.cni.cncf.io/networks-status: |-
 > Once a Pod reaches Running state, the CNI operation is finished successfully and the annotation will present on the Pod.
 
 ### Annotations Exposure Method
-A subset of the information in the `networks-status` annotation needs to be passed down to the virt-launcher for composing the right domain configuration.
+A subset of the information in the `network-status` annotation needs to be passed down to the virt-launcher for composing the right domain configuration.
 
 There have been several methods explored, recorded in [Appendix 1: Alternative Annotations Exposure Methods](#alternative-solutions).
 
 In this section the chosen method is explained.
 
-#### Expose `networks-status` annotation to virt-launcher Pod
+#### Expose `network-status` annotation to virt-launcher Pod
 The annotation content can be exposed to virt-launcher application using Kubernetes Downward API [[1]](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/):
 
 virt-controller creates an annotation that will hold the mapping between the (VMI) SR-IOV networks and PCI addresses, e.g: `kubevirt.io/network-to-pci-address`.
 
-1. Given the `networks-status` annotation, virt-controller process the content and compose 1:1 mapping between a (VMI) network from it and the device PCI address information.
+1. Given the `network-status` annotation, virt-controller process the content and compose 1:1 mapping between a (VMI) network from it and the device PCI address information.
    Create `network-to-pci-address` annotation with the mapping information, for example:
     ```yaml
     kubevirt.io/network-to-pci-address: |-
@@ -489,15 +489,15 @@ virt-controller creates an annotation that will hold the mapping between the (VM
 
 ### New Mapping Algorithm
 Currently, virt-launcher uses the information from the `PCIDEVICE` and `KUBEVIRT_RESOURCE_NAME` environment variables,
-the proposal is to use the information from Multus `k8s.v1.cni.cncf.io/networks-status` annotation in order to map between the SR-IOV interfaces and the underlying VF PCI address.
+the proposal is to use the information from Multus `k8s.v1.cni.cncf.io/network-status` annotation in order to map between the SR-IOV interfaces and the underlying VF PCI address.
 
 #### Mapping logic
-1. Map the `interface` field in `networks-status` annotation to VMI spec.networks:
+1. Map the `interface` field in `network-status` annotation to VMI spec.networks:
 - On pod creation, a network request annotation (`k8s.v1.cni.cncf.io/networks`) is created on the virt-launcher pod [[1]](https://github.com/kubevirt/kubevirt/blob/c4b6ae63c5a7f642ab86b0755dabca3b814ecb39/pkg/virt-controller/services/template.go#L1692).
   - The annotation creates a network request per each non-default multus network.<br/>
     For each CNI entry, each name is given an indexed name `net#` saved on is set in the `interface` field, acting as an identifier.
-  - The VMI spec.networks `interface` filed equal to `interface` field the `networks-status` annotation given by multus-cni.
-- Duplicating this logic gives us a 1:1 map between the `interface` field on the `networks-status` annotations (net1, net2, etc..) and the network name.
+  - The VMI spec.networks `interface` filed equal to `interface` field the `network-status` annotation given by multus-cni.
+- Duplicating this logic gives us a 1:1 map between the `interface` field on the `network-status` annotations (net1, net2, etc..) and the network name.
 
 ##### Example
 The following VMI has 1 bridge and 2 SR-IOV secondary networks:
@@ -592,22 +592,22 @@ Since we’re only interested in the SR-IOV networks, the network-name to PCI ad
 
 ##### Backward compatibility:
 The mapping mentioned above could fail for several reasons, among them:
-- `networks-status` annotation is not present on the VMI.
-- `networks-status` annotation does not hold the PCI address information for all the SR-IOV interfaces.
-- Annotation is not valid, such a case can happen if the `networks-status` annotation is breaking the API or Multus deployed on the cluster is older than [v3.7](https://github.com/k8snetworkplumbingwg/multus-cni/releases/tag/v3.7).<br/>In any case, in the event of a mapping failure, Kubevirt will fall back to the old mapping method, in order to not break old VMIs on old deployments.
+- `network-status` annotation is not present on the VMI.
+- `network-status` annotation does not hold the PCI address information for all the SR-IOV interfaces.
+- Annotation is not valid, such a case can happen if the `network-status` annotation is breaking the API or Multus deployed on the cluster is older than [v3.7](https://github.com/k8snetworkplumbingwg/multus-cni/releases/tag/v3.7).<br/>In any case, in the event of a mapping failure, Kubevirt will fall back to the old mapping method, in order to not break old VMIs on old deployments.
 
 # Appendix 1: Alternative Annotations Exposure Methods <a id="alternative-solutions"></a>
 ## 1. virt controller exposes network to pci mapping through the VMI annotations <a id="alternative-option-1"></a>
 1. virt-controller processes the VMI (virt-launcher) pod and adds it to the VMI annotations.
 
 2. As part of virt-launcher VMI spec rendering:
-- Given the `networks-status` annotation exists on the VMI object, extract the mapping between SR-IOV networks and the VFs PCI addresses and create each HostDevice accordingly.
-- In case `networks-status` annotation doesn't exist, fall back to the legacy mapping method (based on environment variables).
+- Given the `network-status` annotation exists on the VMI object, extract the mapping between SR-IOV networks and the VFs PCI addresses and create each HostDevice accordingly.
+- In case `network-status` annotation doesn't exist, fall back to the legacy mapping method (based on environment variables).
 
 ### Backward compatibility
 |                                                       | old virt-controller                                | new virt-controller                                                                                                                                                                                                              |
 |:------------------------------------------------------|:---------------------------------------------------|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| Already existing VMs<br/>(old virt-launcher)          | 1. No regression <br/> VFs mapping might be wrong. | 2. No regression <br/><br/>VFs mapping might be wrong.<br/>virt-controller adds the pod `networks-status` annotations to VMI annotations,but virt-launcher wont perform the assignment according to it.                          |
+| Already existing VMs<br/>(old virt-launcher)          | 1. No regression <br/> VFs mapping might be wrong. | 2. No regression <br/><br/>VFs mapping might be wrong.<br/>virt-controller adds the pod `network-status` annotations to VMI annotations,but virt-launcher wont perform the assignment according to it.                           |
 | VM creation/restart/migration<br/>(new virt-launcher) | -                                                  | 4. No regression<br/>Once the VM is restarted, VMI or Pod is re-created, the new SR-IOV interfaces assignment logic will apply.<br/>In case the VM is migrated, the VM on target will have SR-IOV interfaces assigned correctly. |
 
 #### Conclusion
@@ -624,7 +624,7 @@ The mapping mentioned above could fail for several reasons, among them:
 
 ## 2. virt-controller exposes network-to-pci mapping through two VMI annotations <a id="alternative-option-2"></a>
 Following [option 1](#alternative-option-1) con (2), this solution mitigates it as follows:
-In order to ensure that the correct mapping information is passed down to virt-launcher, it is necessary to record each of the VMI virt-launcher pods `networks-status` annotation.
+In order to ensure that the correct mapping information is passed down to virt-launcher, it is necessary to record each of the VMI virt-launcher pods `network-status` annotation.
 
 1. virt-controller processes all VMI (virt-launcher) pods and for each, creates a VMI annotation that has the network-to-pci mapping.<br/>During migration, a VMI has two pods (source and target).<br/>   The pod identifier is embedded into the annotation key (e.g: pod name, UID, etc..).
    #### Example
@@ -641,7 +641,7 @@ In order to ensure that the correct mapping information is passed down to virt-l
 2. virt-launcher picks the relevant annotation and perform the mapping accordingly.
 
 #### Pros
-1. No need to synchronize the `networks-status` annotation from the pods (source and target).
+1. No need to synchronize the `network-status` annotation from the pods (source and target).
 
 #### Cons
 2. virt-controller need to maintain two annotations during migration and one on the regular flow.
@@ -651,7 +651,7 @@ As part of the virt-handler VM create/update flow, the VMI spec is processed and
 
 1. Extend virt-handler to virt-launcher command options to include the mapping information.
 
-2. As part of virt-handler VMI update flow, fetch the corresponding virt-launcher pod and read its `networks-status` annotation.<br/>If the annotation exists, extract the network-to-pci mapping and send it to virt-launcher.
+2. As part of virt-handler VMI update flow, fetch the corresponding virt-launcher pod and read its `network-status` annotation.<br/>If the annotation exists, extract the network-to-pci mapping and send it to virt-launcher.
 
 3. As part of virt-launcher VMI spec rendering:<br/> if the network-to-pci mapping information is valid, create each SR-IOV HostDevice accordingly. <br/> In case the information is invalid fall back to the legacy mapping method (based on environment variables).
 
@@ -705,13 +705,13 @@ there will be a general propose annotation that enable passing any kind of infor
 
 In order to pass the information from the pod to the virt-launcher application, the following step are to be taken:
 - Create a generic annotation for passing information from the pod manifest to the application that runs in it. This annotation is using the downward api to make it available to the application.
-- Given the `networks-status` annotation, process the content and compose structure that contains a 1:1 mapping between a (VMI) network from it and the device PCI address. This data is added to the generic annotation from earlier.
+- Given the `network-status` annotation, process the content and compose structure that contains a 1:1 mapping between a (VMI) network from it and the device PCI address. This data is added to the generic annotation from earlier.
 - Once the virt-launcher application requires this network-to-pci information, it can read the content.
 - In the context of SR-IOV, virt-launcher will use this data when creating the domain for the first time and when hotplug-ing the devices.
 
 1. virt-controller creates `virt-metadata` annotation that will hold generic information about the pod that we want to expose to the virt-launcher pod user.
 
-2. Given the `networks-status` annotation, process the content and compose 1:1 mapping between a (VMI) network from it and the device PCI address information and add it to `virt-metadata` annotation.
+2. Given the `network-status` annotation, process the content and compose 1:1 mapping between a (VMI) network from it and the device PCI address information and add it to `virt-metadata` annotation.
    For example:
 ```yaml
 kubevirt.io/virt-metadata: |-
--- a/design-proposals/sriov-interfaces-mapping.md
+++ b/design-proposals/sriov-interfaces-mapping.md
@@ -94,8 +94,7 @@ As part of the VM create flow, virt-controller renders the given VMI spec and cr
 For each secondary network:
 - Realize the device `resourceName` by fetching the specified NetworkAttachmentDefinition.
 - Read the NetworkAttachmentDefinition `k8s.v1.cni.cncf.io/resourceName` annotation.
-- Add the following environment variable to the compute container (spec.Env):
-  `KUBEVIRT_RESOURCE_NAME_<network name>=<resource name>`
+- Add the following environment variable to the compute container (spec.Env):<br/>`KUBEVIRT_RESOURCE_NAME_<network name>=<resource name>`
 
 Once virt-launcher Pod is ready, virt-handler process the VMI object and eventually trigger the virt-launcher to synchronize with the new spec (i.e: virt-handler sends gRPC call to virt-launcher [[4]](https://github.com/kubevirt/kubevirt/blob/1a8f08103e48c5f2bb2f5826d118507ce7ec1f0c/pkg/virt-launcher/virtwrap/manager.go#L822)).
 <br/>
@@ -373,7 +372,7 @@ In order to pass the VM network to VF PCI address mapping from the pod to the vi
 
 The following sections go into more details.
 
-## Utilizing “k8s.v1.cni.cncf.io/networks-status” annotation
+## Utilizing `k8s.v1.cni.cncf.io/networks-status` annotation
 This annotation includes the information required to do one-to-one mapping between a VM interface and VF PCI address:
 
 ### Example
@@ -415,8 +414,8 @@ k8s.v1.cni.cncf.io/networks-status: |-
     ]
 ```
 
-> **_Note_**: Multus provides PCI device information since version [v3.7](https://github.com/k8snetworkplumbingwg/multus-cni/releases/tag/v3.7).
-> In order to maintain backward compatibility, when the annotation is missing, virt-launcher will fall to the current mapping method based on environment variables.
+> **_Note_**: Multus provides PCI device information since version [v3.7](https://github.com/k8snetworkplumbingwg/multus-cni/releases/tag/v3.7).<br/>
+> In order to maintain backward compatibility, when the annotation is missing, virt-launcher will fall to the current mapping method based on environment variables.<br/>
 > Once a Pod reaches Running state, the CNI operation is finished successfully and the annotation will present on the Pod.
 
 ### Annotations Exposure Method
@@ -467,7 +466,7 @@ virt-controller creates an annotation that will hold the mapping between the (VM
           ...
     ```
 
-    > **_Note_**: The annotation details will be populated at `/etc/podinfo/network-to-pci-address` file inside the pod.
+    > **_Note_**: The annotation details will be populated at `/etc/podinfo/network-to-pci-address` file inside the pod.<br/>
     > Any change to the annotation will be reflected in the file.
 
 3. As part of virt-launcher VMI spec rendering, given the downward API file exists (e.g: `/etc/podinfo/network-to-pci-address`),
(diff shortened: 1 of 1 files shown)


design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,711 @@
...
+  network-to-pci-address: 
+  {
+    “sriov-network-vlan-100”: “0000:04:02.5”,
+    “sriov-network-vlan-200”: “0000:04:02.2”
+  }
+}
+```
+
+2. Mount virt-metadata annotation using Downward API as part of virt-launcher Pod creation:
+...
```
RamLavi, line 426: this lacks markdown

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,711 @@
...
+      }]
+```
+
+Since we’re only interested in the SRIOV networks, the network-name-to-pci-address mapping will be:
+- sriovnet-vlan100-secondary-mac ⇒ 0000:65:00.2
+- sriovnet-vlan100-third-mac ⇒ 0000:65:00.3
+
+> **_Note_**:
+> We are relying on the fact that the “interface” naming convention in [[2]](https://github.com/kubevirt/kubevirt/blob/c4b6ae63c5a7f642ab86b0755dabca3b814ecb39/pkg/virt-controller/services/template.go#L1692) will remain the same.
+> To ensure that, we plan cover it with e2e test in case the naming logic is changed without doing the same in the SRIOV device creation.
```
RamLavi, line 575:
```suggestion
> To ensure that, we plan move this logic to a new package that will be used for all scenarios related to this annotations. Moreover, this will be covered in e2e test in case the naming logic is changed without using the new package.
```

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,711 @@
...
+1. No need to use Downward API and mount its Volume.
+2. No need to read files and parse their content.
+3. No API changes.
+
+### Cons
+1. Require patching the VMI metadata by virt-controller VMIController [[4]](https://github.com/kubevirt/kubevirt/blob/1a8f08103e48c5f2bb2f5826d118507ce7ec1f0c/pkg/virt-controller/watch/vmi.go#L1054)  [[5]](https://github.com/kubevirt/kubevirt/blob/1a8f08103e48c5f2bb2f5826d118507ce7ec1f0c/pkg/virt-controller/watch/vmi.go#L509).
+
+2. Won't play nice with live migration.<br/>On migration there will be two sources for the networks-status annotation, the migration source and target pods, the VMI might end up with having the network-status annotation with the wrong annotation details.
+
+### Extending each VMI underlying virt-launcher Pod networks-status annotation
```
RamLavi, line 630:
```suggestion
### 2. Extending each VMI underlying virt-launcher Pod networks-status annotation
```

RamLavi: Or maybe 1.1?

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,711 @@
...
+2. virt-launcher pick the relevant details form the annotation and perform the mapping accordingly. 
+
+#### Pros
+1. The correct information required to perform the mapping is passed down even on migration. 
+ 
+#### Cons
+2. Requires virt-controller to monitor and update the annotation content according to the existence of each pod (migration source and target pods).
+
+## 2.Expose Network Status Annotation via VMI metadata by virt-handler
+As part of the virt-handler VM create / update flow, the VMI spec is processed and virt-launcher is triggered to synchronize with the new VMI spec [[1]](https://github.com/kubevirt/kubevirt/blob/1a8f08103e48c5f2bb2f5826d118507ce7ec1f0c/pkg/virt-handler/vm.go#L2639) (i.e: sends [SyncVMI](https://github.com/kubevirt/kubevirt/blob/1a8f08103e48c5f2bb2f5826d118507ce7ec1f0c/pkg/virt-launcher/virtwrap/manager.go#L822) gRPC call to virt-launcher).
```
RamLavi, line 658: maybe worth mentioning that we will need the extended `kubevirt.io/virt-launcher-networks-status` annotations for this as well

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,711 @@
...
+|:------------------------------------------------------|:---------------------------------------------------|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Already existing VMs<br/>(old virt-launcher)          | 1. No regression <br/> VFs mapping might be wrong. | 2. No regression <br/><br/>VFs mapping might be wrong.<br/>virt-controller annotates the VMI with the network status annotation from the pod,but virt-launcher will not use it.                                                  |
+| VM creation/restart/migration<br/>(new virt-launcher) | -                                                  | 4. No regression<br/>Once the VM is restarted, VMI or Pod is re-created, the new SR-IOV interfaces assignment logic will apply.<br/>In case the VM is migrated, the VM on target will have SR-IOV interfaces assigned correctly. |
+
+#### Conclusion
+- (2) virt-controller will add the network-status to the VMI, the guest will have no disruption and its VFs mapping will not change.
```
RamLavi, line 617: how about this?
```suggestion
- (related to solution section 2) virt-controller will add the network-status to the VMI, the guest will have no disruption and its VFs mapping will not change.
```

RamLavi (changes requested):
Added comments, 
Also I am missing the solution with vmi-status

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,711 @@
...
+> **_Note_**: Multus provides PCI device information since version v3.7.
+
+> **_Note_**: in order to maintain backward compatibility, when the annotation is missing, virt-launcher will fall back to naive mapping and use the legacy environment variable.
+
+> **_Note_**: Once a Pod reaches Running state, the CNI operation is finished successfully and the annotation will present on the Pod.
+
+### Annotations Exposure Method
+A subset of the information in the networks-status annotation needs to be passed down to the virt-launcher for composing the right domain configuration.
+
+There have been several methods explored, recorded in Appendix 1 - Alternative Annotations Exposure Methods .
```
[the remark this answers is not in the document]
ormergi: Done

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,711 @@
...
+virt-controller creates “virt-metadata” annotation that will hold generic information about the pod that we want to expose to the virt-launcher pod user.
+
+1. Given the networks-status annotation, process the content and compose 1:1 mapping between a (VMI) network from it and the device PCI address information and add it to virt-metadata annotation.
+For example:
+```yaml
+kubevirt.io/virt-metadata: |-
+{
+  network-to-pci-address: 
+  {
+    “sriov-network-vlan-100”: “0000:04:02.5”,
```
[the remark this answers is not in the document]
ormergi: Done

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,711 @@
...
+    “sriov-network-vlan-200”: “0000:04:02.2”
```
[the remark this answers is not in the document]
ormergi: Done

design-proposals/sriov-interfaces-mapping.md
ormergi: Done

design-proposals/sriov-interfaces-mapping.md
ormergi: Sounds good, done.

design-proposals/sriov-interfaces-mapping.md
ormergi: Isn't it clear from the numbering to which cell each bullet is related in the table?
I am asking because I think it's cleaner with only the numbering.

design-proposals/sriov-interfaces-mapping.md
ormergi: Done

design-proposals/sriov-interfaces-mapping.md
ormergi: Note that it is mentioned at line 681, do you think it should be mentioned anyway?

RamLavi (approved):
/lgtm

xpivarc:
/cc @rmohr @davidvossel @xpivarc 
Pinging you as you were involved in a discussion about this.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,714 @@
...
+| VM networks names           | sriovnet100, sriovnet200   |
+| Allocated VFs PCI addresses | 0000:04:03.6, 0000:04:03.7 |
+
+When HostDevice object is created its PCI address is picked as follows:
+networkToResource["sriovnet100"]  -------------------------> "kubevirt.io/sriov_net"
+addressesByResource["kubevirt.io/sriov_net"] ---------> "0000:04:03.6"
+
+There is no guarantee that "0000:04:03.6" is the correct PCI address for the network "sriovnet100"  and that the VF is configured with VLAN 100.
+
+This example shows that the current environment variable PCIDEVICE (given by sriov-dp) is simply not enough to directly map between the VM’s networks and  the VFs PCI address.
```
rmohr, line 277: Could you explain why this does not need a fix in sriov-dp? Do other applications in such a situation not care about knowing which device is which, if they have different configuration?

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,714 @@
...
+The mapping mentioned above could fail for several reasons, among them:
+- networks-status annotation is not present on the vmi
+- networks-status annotation does not hold the PCI address information for all the SRIOV interfaces
+- Annotation is not valid
+
+Such a case can happen if the networks-status annotation is breaking the API or Multus deployed on the cluster is older than v3.7.
+In any case, in the event of a mapping failure, we will simply use the old mapping, in order to not break old VMIs on old clusters.
+
+# Appendix 1: Alternative Annotations Exposure Methods
+## 1. virt controller exposes network to pci mapping through the VMI annotations
```
rmohr, line 593: I don't see here the established communication flow considered:

 * virt-controller writes the node mapping to  the VMI status
 * virt-handler extracts the info for the current node the VMI s running on and sends it with its [VirtualMachineOptions](https://github.com/kubevirt/kubevirt/blob/c4b6ae63c5a7f642ab86b0755dabca3b814ecb39/pkg/handler-launcher-com/cmd/v1/cmd.pb.go#L334) construct over to the domain xml converter to virt-launcher
 * works the same for a migration target virt-handler
 * we also already have a status cleanup happen after migrations, which can remove the old host entry and only keep the current one. Also virt-controller could do the cleanup, since it already provides the info in the first place.
 
 Where I would agree that would not make sense:
 
  * trying to modify the VMI status in a way in virt-handler so that virt-launcher would only see relevant info. We don't need such tricks. We have VirtualMachineOptions already.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,714 @@
...
+
+4. If the file doesn't exist, fall back to the legacy mapping method (based on environment variables).
+
+##### Backward compatibility
+- On post Kubevirt upgrade new virt-controller will treat new and old VMIs and their pod in the same manner - adding the virt-metadata annotation on the pods.
+
+- Already existing VMIs that runs old version of virt-launcher pods...
+
+##### Pros
+1. This solution's involves only virt-controller and virt-launcher.
```
rmohr, line 466: I would see this additional communication channel as a con, since it is another dimension to consider in the communication. The established gRPC interface looses integrity and one has to know about more  exceptions.

That has for instance implications when thinking about the cloud hypervisor PoC. We have a nice interface already, which we would weaken.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,714 @@
...
+2. The mapping info is passed down using the standard Kubernetes mechanism (Downward API) with no need to add additional logic.
```
rmohr, line 467: I wonder if there is a significant difference regarding to the additional logic. virt-launcher definitely is more complex.

I would kind of agree if we would not have the established path already, which includes status cleanup after migrations.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,714 @@
...
+3. Flexibility, no API change to VM or VMI objects.
```
rmohr, line 468: we already have to take care of compatibility between virt-handler and virt-controller. If we would  not already have that, and we would not need the other channel I might agree to a certain point.

I would in general not see a schemaless blob on an annotation as a plus.

rmohr (commented):
Interesting problem.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,714 @@
...
+    ]
+```
+
+> **_Note_**: Multus provides PCI device information since version v3.7.
+
+> **_Note_**: in order to maintain backward compatibility, when the annotation is missing, virt-launcher will fall back to naive mapping and use the legacy environment variable.
+
+> **_Note_**: Once a Pod reaches Running state, the CNI operation is finished successfully and the annotation will present on the Pod.
+
+### Annotations Exposure Method
```
rmohr, line 405: In general I would stay away from annotations whenever possible. Multus has to make use of them, but we should not use them if not absolutely necessary. Neither on the VMI, nor on the Pod.

design-proposals/sriov-interfaces-mapping.md
rmohr: Oh one addition: @davidvossel just reminded me that virt-launcher knows the node name already anyway, so using the Options may not even be necessary, which would make it even simpler. Just cleanup after migration would be a topic.

design-proposals/sriov-interfaces-mapping.md
ormergi: In general totaly agree, but in this case inevitable because  there is no way to get the network-to-pci mapping from an API other than from Multus network-status annotation.

Note that we use `k8s.v1.cni.cncf.io/networks` annotation in order to connect the pod to secondary networks.
The `k8s.v1.cni.cncf.io/networks-status` annotation is the result of that operation.

rmohr: Yes, if we have to integrate with third-party apps which use it there may be no other way. I am not against using existing multus annotations. I am just arguing about adding additional ones where we already have better communication pathes.

design-proposals/sriov-interfaces-mapping.md
ormergi: I added the solution you describe at option 4, could you please check and let me know if that is what you meant?

design-proposals/sriov-interfaces-mapping.md
ormergi: Device plugins are executed before the CNI plugin does, in our care the sriov-dp execute first and allocates the VFs later on the sriov-cni is executed and configures the VFs according to the network config (passed by the NetworkAttachmentDefinition).
(I guess this is why those components are independent and not a single entity.)
At the time the sriov-dp is executed it doesn't have enough information to distinguish between the VFs it allocates, it doesn't have the network config.
The NPWG device-info doc explains it well:
https://github.com/k8snetworkplumbingwg/device-info-spec#current-state--limitations

Applications that need to distinguish between such devices, should expose the networks-status annotation through the downward API. This topic is also mentioned NPWG device info doc check this out: 
https://github.com/k8snetworkplumbingwg/device-info-spec#workflow-summary (last bullet).

rmohr: so, if the status annotation from multus already contains the information on the virt-launcher pod, why would we need our own virt-metadata annotation? I probably miss something.

design-proposals/sriov-interfaces-mapping.md
ormergi: I understand your point and agree its there should be single communication channel,
note that from what we tired, piggiy-backing the current "medium" involves more complex logic than the one suggested (mounting the annotation), also it brings more complexities such as when a VM is migrated or when the migration is canceled.

Regarding cloud hypervisor PoC, I think there will be no implications because passing down the network-to-pci mapping does not involve the hypervisor, this mapping information eventually passes to Libvirt.

design-proposals/sriov-interfaces-mapping.md
ormergi: From what we tried there is quite different, other solutions than mounting the networks-status annotation
involve maintaining the the VMI status/annotation (that holds the network-to-pci mapping information).
For example: when the VM is migrated or migration is canceled, it requires synchronizing the VMI annotations/status from two different sources (source and target pods).

Note that in any case, we need to do the virt-launcher part, where the passed network-to-pci mapping is digested and the devices are assigned according to it.

design-proposals/sriov-interfaces-mapping.md
rmohr: > I understand your point and agree its there should be single communication channel,
note that from what we tired, piggiy-backing the current "medium" involves more complex logic than the one suggested (mounting the annotation), also it brings more complexities such as when a VM is migrated or when the migration is canceled.

Yes, except the entrypoints for picking or deleting an entry from a map on the status are already there for these scenarios. They just have to be extended once in a generic way (like a map based on cluster-name, where it just deletes the cluster info in question after a successful or deleted migration). That sounds much easier to me than, as an engineer having to consider that parts of the info are attached to the pod as annotation and parts are attached on the vmi status.

We then also don't have to consider things like:
 * how big will the pod get from a transfer perspective? If we hit an etcd limit, it may be the fault of the multus annotation, we are not involved
 * is some information needed for doing some potentially security sensitive task? If yes, we potentially have issues retrieving it in a trusted way from the launcher-pod-downwards-api location to virt-handler.
 
 Note that if the existing multus annotation already has the information needed to fix the sriov issue, this is a completely different discussion. Just using the annotation and extract the info from there to get it to the converter, is still a little bit weird, but a completely different discussion than introducing such a communication channel.

design-proposals/sriov-interfaces-mapping.md
ormergi: The networks-status annotation includes more information regarding the pod secondary NICs such as: NetworkAttachmentDefinition object names, the cluster CNI plugin name, pod network subnet, and basically whatever a CNI reporters according to the device-info spec.
For example: 
Consider the case where a VM is connected to the same network as the cluster nodes, that NICs status would be reflected on the networks-status annotation.
If there is a malicious entity inside the pod it could have access to that info if the annotation is mounted as is.

The idea is to extract only the relevant SR-IOV interfaces network-to-pci information and add it to the virt-metadata annotation that will be later on mounted to the pod.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+
+- No disruption to already existing VMIs workloads, though the SR-IOV mapping may not be correct.
+
+##### Pros
+1. This solution involves only virt-controller and virt-launcher.
+2. The mapping info is passed down using the standard Kubernetes mechanism (Downward API) with no need to add additional logic.
+3. Flexibility, no API change to VM or VMI objects.
+
+##### Cons
+2. Depends on Downward API feature and volume mount on the Pod.
```
rmohr, line 480: In general I just prefer to not add more mounts. I am pretty sure it is not a problem to mount to a few pods the volume where a sriov device is required, it may be a problem if we add it to all pods. If possible I would want to avoid it, but without further investigations I would have no prove that it would create an issue. If we can avoid it (for a big set of pods), we just don't have to investigate it.

ormergi:
> Interesting problem.

@rmohr thanks for the feedback it is valuable and appreciated.

design-proposals/sriov-interfaces-mapping.md
rmohr: if there is security relevant info there, then better not to mount it. :+1: I would recommend the existing info flow then through virt-handler.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+3. virt-launcher VMI extract the network-to-pci mapping from the VMI status and create each HostDevice accordingly.<br/>
+   In case there is not enough information in the VMI status (i.e: missing host PCI address) fall back to legacy mapping method. 
+
+#### Pros
+1. No need for annotations.
+
+#### Cons
+1. The status field will have to support mapping from two pods, supporting migration.
+2. It is confusing to use the VMI status as an input to the virt-launcher.
+3. The host PCI address is now exposed to the user and may be confusing.
```
rmohr, line 694: I wonder if it is more confusing than huge annotations on pods. We already track a lot of internal stuff there which users can not directly use for anything specific, like migration ports and so forth. I would not worry about that.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+2. The virt-controller updates the status fields with the mapping information.
+
+3. virt-launcher VMI extract the network-to-pci mapping from the VMI status and create each HostDevice accordingly.<br/>
+   In case there is not enough information in the VMI status (i.e: missing host PCI address) fall back to legacy mapping method. 
+
+#### Pros
+1. No need for annotations.
+
+#### Cons
+1. The status field will have to support mapping from two pods, supporting migration.
```
rmohr, line 692: Yes that is a one-time action, where luckily most entrypoints already exist since we already have to cleanup the status after migrations. Then we are in the good situation that we don't have to think about separate flows anymore.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+2. It is confusing to use the VMI status as an input to the virt-launcher.
```
rmohr, line 693: If you like you can also extract it from there and pass it via `VirtualMachineOptions` which passes in a lot of info already, which is not in the spec but needed to create correct libvirt xml files.

design-proposals/sriov-interfaces-mapping.md
rmohr: Yes looks pretty much accurate. I agree that there is likely a little bit more to code, but afterwards we have to care about much less regarding to any info flow.

design-proposals/sriov-interfaces-mapping.md
rmohr: > Regarding cloud hypervisor PoC, I think there will be no implications because passing down the network-to-pci mapping does not involve the hypervisor, this mapping information eventually passes to Libvirt.

The cloud-hyperviso PoC is pure cloud-hypervisor with its own virt-launcher, just using our gRPC protocol. Libvirt is not involved. Especially this would discussion here would probably not have any impact on it for a very long time, since we are for sure far away from adding SRIOV there, but it makes it in general more difficult to extend kubevirt if multiple info-flows exist to virt-launcher.

design-proposals/sriov-interfaces-mapping.md
RamLavi: Just to clarify, we are not exposing the network-status data, but rather a digested subset of it (network-pci map).

design-proposals/sriov-interfaces-mapping.md
RamLavi: As this option suggests to add this information to the vmi-status, I think it would be rather confusing that the user configures the vmi with guest PCI-Address, like so:
```
kind: VirtualMachine
…
spec:
  domain:
    devices:
      interfaces:
      - name: sriovnet1
        pciAddress: 0000:00:01.0
        sriov: {}
    …
  networks:
  - name: sriovnet1
    multus:
      networkName: default/sriov-network-vlan100
```
(`0000:00:01.0` being the pci-address as it will appear on the guest)

and the vmi-status would show something like this 
```
apiVersion: v1
items:
- apiVersion: kubevirt.io/v1
  kind: VirtualMachineInstance
...
  spec:
...
  status:
    guestOSInfo:
...
    interfaces:
    - infoSource: domain, guest-agent
      interfaceName: eth0
...
    - infoSource: domain, guest-agent
      interfaceName: eth1
      ipAddress: fe80::a8bb:ccff:fedd:ee01
      ipAddresses:
      - fe80::a8bb:ccff:fedd:ee01
      hostPciAddress: 0000:40:20.0
```

My point being, this pci-address is for internal use, and thus should not appear in the vmi-status. 
Also consider the complication of how to expose this while in migration.

rmohr: > My point being, this pci-address is for internal use, and thus should not appear in the vmi-status.

Yes I understand your concern. I would just not mix it. I would favor something like this:

```
status:
  nodeInfo:
     "node01":
        interfaces:
          [...]
     "node02":
        [...]
```

and just clear the whole entry from there after a migration success or failure. The annotation proposal may be even simpler, but this is simple too and it follows all conventions with the advantages mentioned.

As said, we already include a lot of info regarding to migrations and so forth which are not relevant for users either.

rmohr: Especially, if this route is given, we will never have to differentiate or change the info flow on this level not matter what comes in the future. The other approach sounds tempting now, but does not guarantee at all that this approach is not needed too at some point.

As I understand it, the proposals were mostly driven by the assumption that metadata has to be used one or the other way, since info like this can't be in the vmi status or the pod status. At least that is how I understood the feedback on the PR. I would reconsider it with the new info provided.

RamLavi:
@rmohr thank you for your valuable feedback, it is highly appreciated. :-)
If I understand your thinking, there is an already established way of communication (i.e. the grpc between virt-handler and virt-launcher), and we should not create a new one, in order to not weaken the former.

However this is not accurate - the current implementation does exactly that: 
- virt-controller writes env-variables directly to the pod
- the pod uses that to (wrongly) map the SRIOV devices
This is also the current method for vgpu and other host devices.
In that sense, we are keeping the same method.

The reason we chose to use to mount the annotation and not env-vars is because of the nature of network-status annotation - that appears only after the pod is created. Moreover, using downwardAPI via Volume means that the data can be updated on the fly and by that we can support future features of Multus interface hot plugs.

To my understanding, we are not creating a new communication path, but rather only revising the current one, with a supported feature (=downwardAPI) that allows us to reconcile the data in the future should it be needed.

rmohr:
> @rmohr thank you for your valuable feedback, it is highly appreciated. :-) If I understand your thinking, there is an already established way of communication (i.e. the grpc between virt-handler and virt-launcher), and we should not create a new one, in order to not weaken the former.
> 
> However this is not accurate - the current implementation does exactly that:
> 
> * virt-controller writes env-variables directly to the pod
> * the pod uses that to (wrongly) map the SRIOV devices
> 
> This is also the current method for vgpu and other host devices. In that sense, we are keeping the same method.

Ha, I bet you saw that one coming:

First I did not remember that. From my initial proposal read I thought we are talking about multus and gpu components writing these. Great catch. There are also some plugin-env vars defined which external device plugins can use to integrate with kubevirt. I would expect that these stay where they are (at least at the moment).

Second: It is exactly one of the things we can completely eliminate, in this case then not only not weakening the gRPC interface but strengthening it.

> 
> The reason we chose to use to mount the annotation and not env-vars is because of the nature of network-status annotation - that appears only after the pod is created. 

Yep, that one I got. It basically means for me that a pod is only then fully ready if it also has the full annotation present and we should wait for that before handing over to handler, or letting handler not procede. Otherwise virt-launcher will have to understand that is is still missing info. In this case it does not have to care. It always gets the full info.

Moreover, using downwardAPI via Volume means that the data can be updated on the fly and by that we can support future features of Multus interface hot plugs.

Yes, the gRPC channel has that of course too.

> 
> To my understanding, we are not creating a new communication path, but rather only revising the current one, with a supported feature (=downwardAPI) that allows us to reconcile the data in the future should it be needed.

Here we again come to the area of the drawback which I mentioned before. I understand that k8s supports it. We would at least need an investigation if we expect that many pods end up with the downwards api on what it means for the kubelet. We try to be as slick as possible, to leave the resources for users. The established communication channel has again the advantage, that independent of the data sensitivity it can be used for it all, as well as no infrastructure overhead in addition with the new info.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
+# Overview
+Accurately associate VMs SR-IOV interfaces with the underlying SR-IOV Virtual Functions (VF) devices (PCI) allocated by the sriov-network-device-plugin and sriov-cni.
```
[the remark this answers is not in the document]
ormergi: Note that for secondary SR-IOV networks Kubevirt VMs consume VFs with vfio-pci driver and not "network-device  mode" VFs in order to pass through them to the guest
https://kubevirt.io/user-guide/virtual_machines/interfaces_and_networks/#how-to-expose-sr-iov-vfs-to-kubevirt.

In any case, if a CNI reports the PCI address of the VF according to the NPWG docs,
it will be reflected in the networks-status annotations and the mapping will be correct (as mentioned in the doc, if the network-to-pci info is missing Kubevirt falls back to the legacy mapping method)

As for the primary network, as long as the VF is in "network device mode" Kubevirt will wire it up transparently.
Thanks for the heads up, I will add changes to reflect that it's also supported.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+2. Depends on Downward API feature and volume mount on the Pod.
+
+### New Mapping Algorithm
+Currently, virt-launcher uses the information from the `PCIDEVICE` and `KUBEVIRT_RESOURCE_NAME` environment variables,
+the proposal is to use the information from Multus `k8s.v1.cni.cncf.io/networks-status` annotation in order to map between the SR-IOV interfaces and the underlying VF PCI address.
+
+#### Mapping logic
+1. Map the `interface` field in networks-status annotation to VMI spec.networks:
+- On pod creation, a network request annotation (`k8s.v1.cni.cncf.io/networks`) is created on the virt-launcher pod [[1]](https://github.com/kubevirt/kubevirt/blob/c4b6ae63c5a7f642ab86b0755dabca3b814ecb39/pkg/virt-controller/services/template.go#L1692).
+  - The annotation creates a network request per each non-default multus network.<br/>For each CNI entry, each name is given an indexed name `net#` saved on is set in the `interface` field, acting as an identifier.
```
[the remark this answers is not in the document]
ormergi: I will change it to refer to primary network as well.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+
+## Motivation
+[KubeVirt’s SR-IOV support](https://kubevirt.io/user-guide/virtual_machines/interfaces_and_networks/#sriov) enables wiring up SR-IOV Virtual Functions (VF) to VMs
+by passing its PCI device ID from the node, through the virt-launcher pod, and then into the VM using [Libvirt](https://libvirt.org/) API.
+<br/>
+Each VM SR-IOV device (i.e: VF) is associated with a VM secondary network.
+
+When a VM is defined with multiple SR-IOV devices from the same [resource pool](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#configurations)
+or connected to the same NetworkAttachmentDefinition, Kubevirt cannot differentiate which VF is associated with which requested device.
+The root cause is that the PCI information passed by the sriov-network-device-plugin though kubelet is not enough to directly map the VFs with their secondary networks.
```
orelmisan, line 14:
```suggestion
The root cause is that the PCI information passed by the sriov-network-device-plugin through kubelet is not enough to directly map the VFs with their secondary networks.
```

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+Fix/change third-party components (e.g: Multus, SR-IOV CNI, etc..)
+
+## Existing Implementation Flow
+In order to discover and allocate VFs to Pods the [sriov-network-device-plugin](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#sr-iov-network-device-plugin) (from now on sriov-dp) and the [sriov-cni](https://github.com/k8snetworkplumbingwg/sriov-cni) are being used.
+<br/>
+The sriov-dp provides an [API to create resource pools](https://github.com/openshid/sriov-network-device-plugin#configurations) of VFs according to various properties,
+Each resource pool is labeled with a resource name.
+
+In the VMI spec, secondary networks point to the desired SR-IOV NetworkAttachmentDefinition object.
+The NetworkAttachmentDefinition is set with the Multus networks-status annotation:
```
orelmisan, line 43:
```suggestion
The NetworkAttachmentDefinition is set with the Multus `k8s.v1.cni.cncf.io/resourceName` annotation:
```

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+  annotations:
+    k8s.v1.cni.cncf.io/resourceName: kubevirt.io/sriov_net
+spec:
+  config: |-
+    {
+        "type":"sriov",
+  ...
+```
+
+As part of Pod creation flow, kubelet executes the required device plugins according to the Pod spec, in our case the sriov-dp.
```
orelmisan, line 77: Please consider giving a little background on how a Pod with SR-IOV is provisioned for the reader to better understand the problem:
1. kubelet calls the DP in order to allocate one or more VFs from a certain resource name.
2. Kubelet calls the container runtime in order to create the pod sandbox.
3. Container runtime calls Multus which does the association between network name, interface name, resource name and allocated PCI address.
4. Multus calls the SR-IOV CNI which configures the VF.
5. Multus adds the operation results in the Pod's `network-status` annotation.

The virt-launcher Pod has:
Several environment variables to map from:
1. Resource name to a list of allocated PCI addresses.
2. Network name to resource name.

Which are not enough to correctly associate VM's network and network interface with the actual VF that was previously configured and is represented as a PCI address.

ormergi:
> Here we again come to the area of the drawback which I mentioned before. I understand that k8s supports it. We would at least need an investigation if we expect that many pods end up with the downwards api on what it means for the kubelet. We try to be as slick as possible, to leave the resources for users. The established communication channel has again the advantage, that independent of the data sensitivity it can be used for it all, as well as no infrastructure overhead in addition with the new info.

@rmohr This is an interesting point, downward API scale, although this proposal is not a standard feature (more like a bug fix) I agree it should be tested like any other feature at scale.

From a few initial tests we did, the annotation mount has a insignificant memory footprint (its mount is tempfs, few KBs).
Regarding the implications on kubelet, we are not sure if there will be any, I want to believe a k8s GA v1 feature is tested at scale.
Perhaps we should ask kubevirt sig-scale wg?

rmohr:
> @rmohr This is an interesting point, downward API scale, although this proposal is not a standard feature (more like a bug fix) I agree it should be tested like any other feature at scale.
> 
> From a few initial tests we did, the annotation mount has a insignificant memory footprint (its mount is tempfs, few KBs). Regarding the implications on kubelet, we are not sure if there will be any, I want to believe a k8s GA v1 feature is tested at scale. Perhaps we should ask kubevirt sig-scale wg?

In the whole discussion there came up no need for a second standardized communication path, except a very small benefit for this specific set of info once due to having to writing a small amount of less code, but all kinds of uncertainties (which I admitted may or may not be an issue, the point is that we have to think about them). With the established case, after an initial minimal higher investment (still not sure that this is really the case if done properly, especially when you also mention later cases where you want to get info on the fly), we are all good and don't ever have to look back. It is good on all fronts. No one has to think about anything in the future.

Given that we all agree that using annotations as preferred communication path is not something which we want, and that the initial considerations why this proposal came up are not an issue for kubevirt, I would just recommend to use a temporary fix close around sriov devices or switch over to the gRPC flow.

design-proposals/sriov-interfaces-mapping.md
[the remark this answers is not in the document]
ormergi: Following the discussion at https://github.com/kubevirt/kubevirt/pull/8226#discussion_r940012523, no need to reflect the mentioned above.

design-proposals/sriov-interfaces-mapping.md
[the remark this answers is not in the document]
ormergi: For visibility, following the discussion at https://github.com/kubevirt/kubevirt/pull/8226#discussion_r940012523, no need to change the mentioned above.

design-proposals/sriov-interfaces-mapping.md
ormergi: Thanks for the information you summarized it well.
Initially, I added this background after a few iterations of reviews, we came to the conclusion that it's too much information for the reader to grasp.
I agree it will give more background and thorough understanding, but note that for the common reader it may add more noise and distraction from the gist of the problem we try to fix.

ormergi:
Changes: 
- Changed the proposed solution following the discussion about using dedicated annotation for the network-to-pci mapping.
@rmohr @RamLavi please take another look [[1]](https://github.com/kubevirt/community/pull/185/commits/29031515f3f219adf9ff5411cfc3b19ab84d65eb).
- Fixed @orelmisan review comments [[2]](https://github.com/kubevirt/community/pull/185/commits/52e45dc5537d5dfa627402ba6dfe96d12931d2b4).

design-proposals/sriov-interfaces-mapping.md
orelmisan: The main point of this explanation is that KubeVirt depends on an association that is made by Multus, and is reflected by it.
AFAIK, there is no other place to get this information from, other than the `network-status` annotation.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,761 @@
...
+A subset of the information in the `networks-status` annotation needs to be passed down to the virt-launcher for composing the right domain configuration.
+
+There have been several methods explored, recorded in [Appendix 1: Alternative Annotations Exposure Methods](#alternative-solutions).
+
+In this section the chosen method is explained.
+
+#### Expose `networks-status` annotation to virt-launcher Pod
+The annotation content can be exposed to virt-launcher application using Kubernetes Downward API [[1]](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/):
+
+virt-controller creates an annotation that will hold the mapping between the (VMI) SR-IOV networks and PCI addresses, e.g: `kubevirt.io/network-to-pci-address`.
```
RamLavi, line 432:
```suggestion
virt-controller creates an annotation that will hold the mapping between the (VMI) SR-IOV networks and PCI addresses, e.g: `kubevirt.io/network-pci-map`.
```

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,760 @@
...
+
+#### Expose `network-status` annotation to virt-launcher Pod
+The annotation content can be exposed to virt-launcher application using Kubernetes Downward API [[1]](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/):
+
+virt-controller creates an annotation that will hold the mapping between the (VMI) SR-IOV networks and PCI addresses, e.g: `kubevirt.io/network-to-pci-address`.
+
+1. Given the `network-status` annotation, virt-controller process the content and compose 1:1 mapping between a (VMI) network from it and the device PCI address information.
+   Create `network-to-pci-address` annotation with the mapping information, for example:
+    ```yaml
+    kubevirt.io/network-to-pci-address: |-
```
rmohr, line 436: This is now focused on sriov solely. So good enough for me to fix the bug at hand. As discussed. :+1: 

Please add that on follow up changes there is still the expectation that you fall in line with the VMI status and remove this information flow and move it over to virt-handler.

This may live long, since temporary solutions tend to live long, but when you start to add more data I would expect that you follow the main communicatino paths and remove this workaround.

RamLavi: >Please add that on follow up changes there is still the expectation that you fall in line with the VMI status and remove this information flow and move it over to virt-handler.

Do you mean to add it as comment to the commit or on the code?

rmohr: > Do you mean to add it as comment to the commit or on the code?

a short fixme or so would be nice too yes, but I meant in the proposal, so that we don't start future discussions at 0. Just that this is captured. Also the pros/cons sections are probably not necessarily correctly reflecting the latest discussion points, however I would not insist on reworking them. Just make it clear that eventually it has to be moved to vmi-status.

maiqueb (changes requested):
A few nits.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,760 @@
...
+In the scenario described in https://github.com/kubevirt/kubevirt/issues/6351,
+there are two `SriovNetwork` objects - each configures a different VLAN [[1]](https://github.com/kubevirt/kubevirt/issues/6351#issuecomment-918467186).
+But the VM's networks point to the wrong VFs:
+VM network `sriov-vsrx-if30` points to `default/sriov-vsrx-if30` NetworkAttachmentDefinition, which is configured with VLAN 30, but the underlying VF is configured with a different VLAN (i.e. the wrong PCI address was chosen).
+
+The issue manifests on Kubevirt as follows:
+<br/>
+The SR-IOV network PCI address is determined on virt-launcher code that converts the VMI spec to Libvirt domain.<br/>
+Libvirt domain Hostdev is represented in Kubevirt code by the HostDevice object.<br/>
+For each SR-IOV network a corresponding HostDevice object is created, the PCI address assigned as follows:
```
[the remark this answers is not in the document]
ormergi: I am not sure what you mean. inst it already a bullets list?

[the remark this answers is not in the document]
maiqueb: I mean the markdown code you're providing is [not rendering into a list](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#lists). 

Check how it renders in https://github.com/kubevirt/community/blob/4cf6bcb2328c135612e9b9b35b20b74dce8bc6a1/design-proposals/sriov-interfaces-mapping.md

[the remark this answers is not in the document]
ormergi: Done

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,760 @@
...
+As part of the VM create flow, virt-controller renders the given VMI spec and creates the corresponding virt-launcher Pod [[3]](https://github.com/kubevirt/kubevirt/blob/1a8f08103e48c5f2bb2f5826d118507ce7ec1f0c/pkg/virt-controller/services/template.go#L443) (renderLaunchManifest).
+<br/>
+For each secondary network:
+- Realize the device `resourceName` by fetching the specified NetworkAttachmentDefinition.
+- Read the NetworkAttachmentDefinition `k8s.v1.cni.cncf.io/resourceName` annotation.
+- Add the following environment variable to the compute container (spec.Env):<br/>`KUBEVIRT_RESOURCE_NAME_<network name>=<resource name>`
+
+Once virt-launcher Pod is ready, virt-handler process the VMI object and eventually trigger the virt-launcher to synchronize with the new spec (i.e: virt-handler sends gRPC call to virt-launcher [[4]](https://github.com/kubevirt/kubevirt/blob/1a8f08103e48c5f2bb2f5826d118507ce7ec1f0c/pkg/virt-launcher/virtwrap/manager.go#L822)).
+<br/>
+virt-launcher renders the VMI spec and converts it to Libvirt domain XML, and eventually passes it to Libvirt which in turn create the domain.
```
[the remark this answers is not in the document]
ormergi: Done

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,755 @@
...
+In order to discover and allocate VFs to Pods the [sriov-network-device-plugin](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#sr-iov-network-device-plugin) (from now on sriov-dp) and the [sriov-cni](https://github.com/k8snetworkplumbingwg/sriov-cni) are being used.
+<br/>
+The sriov-dp provides an [API to create resource pools](https://github.com/openshid/sriov-network-device-plugin#configurations) of VFs according to various properties,
+Each resource pool is labeled with a resource name.
+
+In the VMI spec, secondary networks point to the desired SR-IOV NetworkAttachmentDefinition object.
+<br/>
+The NetworkAttachmentDefinition is set with the resource-name annotation:
+<br/>`k8s.v1.cni.cncf.io/resourceName=<resource name>`<br/>
+which later on passed down by Multus to the sriov-cni [[1]](https://github.com/k8snetworkplumbingwg/multus-cni/blob/a28f5cb56c79a582f5ea2b35a61b38f34b937930/examples/README.md#passing-down-device-information).
```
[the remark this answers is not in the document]
ormergi: Done

design-proposals/sriov-interfaces-mapping.md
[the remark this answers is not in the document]
ormergi: Done

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+## Non-Goals
+Fix/change third-party components (e.g: Multus, SR-IOV CNI, etc..)
+
+## Existing Implementation Flow
+In order to discover and allocate VFs to Pods the [sriov-network-device-plugin](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#sr-iov-network-device-plugin) (from now on sriov-dp) and the [sriov-cni](https://github.com/k8snetworkplumbingwg/sriov-cni) are being used.
+<br/>
+The sriov-dp provides an [API to create resource pools](https://github.com/openshid/sriov-network-device-plugin#configurations) of VFs according to various properties,
+Each resource pool is labeled with a resource name.
+
+In the VMI spec, secondary networks point to the desired SR-IOV NetworkAttachmentDefinition object.
```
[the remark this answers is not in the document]
ormergi: Done

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+Each VM SR-IOV interface should be associated with the correct VF.
+
+## Non-Goals
+Fix/change third-party components (e.g: Multus, SR-IOV CNI, etc..)
+
+## Existing Implementation Flow
+In order to discover and allocate VFs to Pods the [sriov-network-device-plugin](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#sr-iov-network-device-plugin) (from now on sriov-dp) and the [sriov-cni](https://github.com/k8snetworkplumbingwg/sriov-cni) are being used.
+<br/>
+The sriov-dp provides an [API to create resource pools](https://github.com/openshid/sriov-network-device-plugin#configurations) of VFs according to various properties,
+Each resource pool is labeled with a resource name.
```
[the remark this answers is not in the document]
ormergi: Done

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+
+The outcome is a non-consistent association between the underlying VFs and the VM networks,
+leaving the VM in a state where the VFs are plugged into the domain wrongly.
+<br/>
+For example, when a SR-IOV interface is defined with a custom MAC address (e.g: 02:00:00:00:00), a custom guest PCI address of (e.g: 0000:01:01.1) and pointing to a NetworkAttachmentDefinition with VLAN configuration (e.g: 100).
+It is not guaranteed that the correct VF (with VLAN 100) will be plugged into the domain on the specified PCI address (0000:01:01.1).
+
+This issue also manifests when the VM interface boot order is set, the VM may end up booting from the wrong interface.
+
+A side effect of this issue is the VM status that may confuse its consumers, with its false advertisement, make them act incorrectly, and introduce other issues.
```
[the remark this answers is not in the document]
ormergi: Done, I rephrased it and removed the example.

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+
+When a VM is defined with multiple SR-IOV devices from the same [resource pool](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#configurations)
+or connected to the same NetworkAttachmentDefinition, Kubevirt cannot differentiate which VF is associated with which requested device.
+The root cause is that the PCI information passed by the sriov-network-device-plugin though kubelet is not enough to directly map the VFs with their secondary networks.
+
+The outcome is a non-consistent association between the underlying VFs and the VM networks,
+leaving the VM in a state where the VFs are plugged into the domain wrongly.
+<br/>
+For example, when a SR-IOV interface is defined with a custom MAC address (e.g: 02:00:00:00:00), a custom guest PCI address of (e.g: 0000:01:01.1) and pointing to a NetworkAttachmentDefinition with VLAN configuration (e.g: 100).
+It is not guaranteed that the correct VF (with VLAN 100) will be plugged into the domain on the specified PCI address (0000:01:01.1).
```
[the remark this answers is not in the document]
ormergi: Done

design-proposals/sriov-interfaces-mapping.md
```diff
@@ -0,0 +1,694 @@
...
+The proposal fixes the current solution, where VFs from the same resource pool are not differentiated by KubeVirt and therefore potentially wrongly plugged to the domain.
+
+## Motivation
+[KubeVirt’s SR-IOV support](https://kubevirt.io/user-guide/virtual_machines/interfaces_and_networks/#sriov) enables wiring up SR-IOV Virtual Functions (VF) to VMs
+by passing its PCI device ID from the node, through the virt-launcher pod, and then into the VM using [Libvirt](https://libvirt.org/) API.
+<br/>
+Each VM SR-IOV device (i.e: VF) is associated with a VM secondary network.
+
+When a VM is defined with multiple SR-IOV devices from the same [resource pool](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#configurations)
+or connected to the same NetworkAttachmentDefinition, Kubevirt cannot differentiate which VF is associated with which requested device.
```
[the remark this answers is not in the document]
ormergi: Done (all over)

design-proposals/sriov-interfaces-mapping.md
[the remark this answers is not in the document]
ormergi: Moved to goals, and rephrased the overview.

ormergi:
@rmohr hey I addressed your comment https://github.com/kubevirt/community/pull/185#discussion_r946485811 at https://github.com/kubevirt/community/pull/185/commits/f427c8670176614b906d70314e056f6a43a8a84e, could you please take another look?

@maiqueb I addressed you comments at https://github.com/kubevirt/community/pull/185/commits/ad26d3a6e0ae3cb1f25bee2f143cee0247ec0e9f and https://github.com/kubevirt/community/pull/185/commits/d2977c216c4c28437da595204e125277536e4142, please take another look.

RamLavi (approved):
/lgtm

rmohr:
/lgtm

regarding the communication flow fix for now and for the future.

orelmisan:
@ormergi Can you please answer [this](https://github.com/kubevirt/community/pull/185#discussion_r946650854) comment?
IMO an emphasis on Multus part of the problem is missing.

rmohr:
/approve

/hold

feel free to unhold for as long as nothing significant changes anymore.

design-proposals/sriov-interfaces-mapping.md
ormergi: I think the point that Kubevirt depends on Multus network-status annotation is expressed in the doc, isn't it?

Regarding the addendum, as you may already know, the flow of pod provisioning with secondary networks hides lots of nuances and various concepts such as but not only: device manager and device plugin API, CNI device info spec, and how Multus works.
Sure we can add a brief summary, but I am afraid it may leave the readers with a vague understanding (or even more confusion) since there are quite a lot of subjects to cover, and just mentioning it is not enough IMO.

As a side note, this is a proposal to fix certain functionality, it expects the reader to have a basic understanding of how Kubevirt and Multus integrate.
I tried to keep it detailed yet not overwhelming, compared to other proposals I think this proposal is pretty detailed.

maiqueb (approved):
Thank you for this detailed proposal.

I don't want to hold this off any longer than it needs, and I will not have the time to focus on it for the foreseeable future.

orelmisan (approved):
Thank you @ormergi

ormergi:
Thank you all!

/unhold

PR #189, opened 2022-09-08
Generic external monitoring

iholder101:
Human readable format: https://github.com/kubevirt/community/blob/79c24e20bf77fa709401a551cdbe18b16707a177/design-proposals/external-monitoring.md


design-proposals/external-monitoring.md
```diff
@@ -0,0 +1,173 @@
+# Overview
+
+Recently there is a lot of ongoing effort to establish a monitoring solution for the Kubevirt organization.
+This effort is very important as a good monitoring solution & design directly impacts an operator's quality.
+For example, operator framework define the [Operator Capability Levels](https://sdk.operatorframework.io/docs/overview/operator-capabilities/),
+as a metric of an operator's maturity. Looking at the current state of Kubevirt, arguably we fill the requirements
+of the first 3 levels, which are "Basic install", "Seamless Upgrades" and "Full Lifecycle" management. That said,
+we probably have some more work to do in order to become level 4 and 5, which are "Deep insights" and "Auto Pilot".
```
sradco, line 8: @iholder-redhat  KubeVirt is already in Deep Insights level and working to reach "Auto Pilot" level

design-proposals/external-monitoring.md
```diff
@@ -0,0 +1,173 @@
...
+These two levels are tightly related to our ability to perform monitoring. When we would fulfil level 4 with deep
+insights, we would carve the way to being able to auto-config/heal/scale the cluster, therefore becoming an "auto-pilot"
+operator.
```
sradco, line 11: This is incorrect. Please update based on my comment above.

design-proposals/external-monitoring.md
```diff
@@ -0,0 +1,173 @@
...
+In other words, we would have `component_virt_handler_is_unhealthy`, `component_virt_controller_is_unhealthy` etc for
+kubevirt/kubevirt, and similar alert for kubevirt/CDI, kubevirt/HCO, etc. This not only causes a burden of implementing
+these alerts separately in every repo, but also demands alignment between every different repo. Developers of every
+repository can differ on their opinions, leading to a huge effort of aligning between them all. Eventually, this leads
+to a situation in which it's very difficult to promote changes and new ideas that require a great amount of discussions
+and alignment.
+
+Once we have something like `kubevirt/monitoring` with a completely decoupled and autonomous from the operator, the
+need for alignment would be reduced significantly. There will still be a need to agree upon which data the operator
+exposes and how, but the monitoring logic itself will be decided in one single place.
```
sradco, line 38: @iholder-redhat  Will this fit also small operators? Won't ir cause to much of an overhead?

design-proposals/external-monitoring.md
```diff
@@ -0,0 +1,173 @@
...
+expert on migration flows aren't familiar with monitoring areas and vice-versa. This often leads to a very large development
+effort in order to merge new PRs.
+
+### Enhanced security: monitor publicly available data only
+
+Today we don't have a clear definition of which data is private and internal, and which data is public.
+
+Let's take an example: the amount of times that virt-controller had reconciled a VM. Currently, this is a completely
+internal data that isn't exposed. On the other hand, monitoring code can report it as metrics to Prometheus. So, is this
+public or private data?
```
sradco, line 62: @iholder-redhat I'm not sure what you mean here. A metric is exposed if it can benefit either the operator for automation, the developers for regression tests and other purposes and for users that need to understand what is the state of their env and if they have issues they get alerts that are based on metrics.
There is no "private/internal" vs. "public" data. We add metrics if they are needed and bring value to any of the stakeholders.

design-proposals/external-monitoring.md
iholder-redhat: IIUC deep insights means that we have many metrics and alerts that provide insightful details of the cluster.
I thought this is a work in progress. Do you think we're already there?

design-proposals/external-monitoring.md
iholder-redhat: Yes.
I think the opposite is correct. The fact that the monitoring code resides within the operator's code enforces the monitoring developers to dive deep into the operator's code, go through tough code reviews, convince the maintainers of their design, etc. This enormous overhead can be minimized by simply learning the objects and their fields, developing quick with fast and simply code reviews, without a need to convince the repo maintainers / reviewers.

As this proposal points out, this is just one of the reasons to work that way, alongside keep private data safe, having better design and simpler code at the operator level, etc.

design-proposals/external-monitoring.md
iholder-redhat: Sure, I'll change that.

design-proposals/external-monitoring.md
iholder-redhat: Let me give you an example.
Let's say that we have an operator that runs a certain CRD. For the example's sake, let's say it's a Human CRD. Now let's say that this Human object has the following fields as part of its status: **age** and **date of birth**.

Now let's say that we want to monitor the Human's location. With how things currently work, we'd probably place monitoring logic in the Human controller which notifies Prometheus about each location change for this Human. In other words, the monitoring code now exposes the human's location to the world, although this location **is not part of Human's status**. So is the location supposed to be exposed, or a hidden implementation detail?

---------

This is very similar to what's happening today in Kubevirt.
As an example - do we expose the number of reconciliations our controller have done? Only to Prometheus. The same goes for VM's dirty rate, for example. This design is problematic. Every piece of data that we're dealing with needs to either be completely open and reachable to the outside world, or kept secretly as an implementation detail.

Following this design will enforce us to think about which data is exposed or not and would eventually help to secure the privacy of our objects. With the current design it's highly probable that we'd accidentally expose something we're not supposed to.

rmohr (commented):
Thanks itamar, a few points inline and a general overall response:

I am not sure I get what you mean with "operator code". Can it be that the differentiation between metrics collection and alerting is not clearly defined in that proposal?

Is it understood that no alert rule needs to be added to KubeVirt? Only metrics which can't be gathered from e.g. looking at a VMI object need to.

It is for instance completely valid to just add a metric, which different people want to use differently for alerting.

In the k8s ecosystem we are already in the pretty fortunate sitatuion that everyone agreen on prometheus metrics and scraping, expsure, ... is therefore already standardized. What to make out of these metrics is not. It can make sense to have a small set of alerting rules in the repo (deployed by the operator), but I think that some of the lately pushed alert rules should not be added to kubevirt and I also put holds on some of those. 

I also want to highlight that not every metric needs to be added to kubevirt necessarily. Like if I insist on wanting to have a timestamp metric (does normally not really sense from a monitoring perspective), or you need it for a very specific reason, you can just look at the VMIs and scrape/push it with your own  exporter. It is not a good generic case possibly.

iholder101:
> Thanks itamar, a few points inline and a general overall response:
> 
> I am not sure I get what you mean with "operator code". Can it be that the differentiation between metrics collection and alerting is not clearly defined in that proposal?

Yes, you're right, sorry for that.

> I also want to highlight that not every metric needs to be added to kubevirt necessarily. Like if I insist on wanting to have a timestamp metric (does normally not really sense from a monitoring perspective), or you need it for a very specific reason, you can just look at the VMIs and scrape/push it with your own exporter. It is not a good generic case possibly.

You're touching at the heart of what this proposal is aiming to solve.

We both agree that some metrics belong to Kubevirt core and some don't. Saying that, I think it's fair to assume that a lot of Kubevirt users would want to add their custom metrics based on their specific needs. Conceptually, the optimal solution would be a complete separation between Kubevirt code and custom monitoring code.

Taking this thought another step forward - most of the custom metrics are based on an already known and visible data. For example, let's say that I want to have a custom metrics reporting advanced statistics about migration based on the VMI migration configuration, my custom labels, etc (this might be a bad example) - all of the data is already exposed the visible on the cluster though inspecting the VMI manifests. Therefore, at least for most of my metrics, all I need (theoretically) is to declarative configure which VMI fields I want to track, which calculation to do with them, what exactly to expose, etc.

Something like the following comes to my mind:
```yaml
variables:
    - name: migrationMode
      fromKind: VirtualMachineInstance
      path: status.MigrationConfiguration.MigrationMode
    - name: lastMigrationTimestamp
      fromKind: VirtualMachineInstance
      path: status.MigrationConfiguration.MigrationTimestamp
      filter:
        labels:
            "vm-I-care-about": "true"
metrics:
    - name: my-metric
      type: bool
      promQL: 'if $migrationMode == "pending" && time.Now() - $lastMigrationTimestamp > 123'
```

--------------

Saying all the above, this is a very premature proposal, so don't take any of this too seriously by now. I understand that I need to think this through, and that's exactly what's being done behind the scenes alongside many other contributors from the monitoring area.

I'm converting this to draft for the time being. Discussions will remain in different channels. Good chance this will be worked on and re-opened in the near future.

PR #193, opened 2022-10-18
Proposal: Temporary VNC token

akrejcir:
This design describes how a bound service account token can be used to access VNC of a specific VM.


design-proposals/bound-token-for-vnc.md
```diff
@@ -0,0 +1,77 @@
...
+[tokens](https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/1205-bound-service-account-tokens)
+bound to a service account. They are mainly used in services running inside the cluster to authenticate
+between each other. 
+
+This document describes how to create a token that can be used to authenticate with the API server
+to only access VNC endpoint of a singe VM.
+
+## Motivation
+
+An administrator may want to give VNC access to a client application without giving it permissions to access
```
lyarwood, line 13: So instead of granting a specific user access to the `/vnc` subresource of a given `VirtualMachineInstance` we instead provide the user access to the `/vnc-token` subresource of the same `VirtualMachineInstance` to generate a token they can then use to access the former for a specific amount of time?

I assume a RBAC role per mapping isn't an option here because users couldn't use it to unlock access to their specific `VirtualMachineInstance` for another user?

design-proposals/bound-token-for-vnc.md
```diff
@@ -0,0 +1,77 @@
...
+
+- A VM owner wants to give access to VNC to a third party application without giving it permissions to modify the VM.
+
+## Repos
+
+- https://github.com/kubevirt/kubevirt
+
+# Design
+
+An API subresource under `virtualmachineinstance` , `/vnc-token` is used to get a token for VNC.
```
lyarwood, line 33: `vnc/token` given we already have `vnc/screenshot`?

design-proposals/bound-token-for-vnc.md
```diff
@@ -0,0 +1,77 @@
+# Overview
+
+Kubernetes API server can create 
+[tokens](https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/1205-bound-service-account-tokens)
```
lyarwood, line 4: Any better docs available?

design-proposals/bound-token-for-vnc.md
akrejcir: I'm not sure I understand your question. A user can have access to the `/vnc` subresource. The token will be used to give access to `/vnc` for non-registered users. When an API request with the token is made, it will be authenticated as a specific ServiceAccount, which only has permissions to access `/vnc`.

lyarwood: Apologies I mixed client with an actual registered user of the cluster, you do go on to highlight that again below.

design-proposals/bound-token-for-vnc.md
lyarwood: ACK understood.

design-proposals/bound-token-for-vnc.md
```diff
@@ -0,0 +1,77 @@
...
+## API Examples
+
+Getting the VNC token:
+```
+GET /apis/subresources.kubevirt.io/v1/namespaces/${VMI_NAMESPACE}/virtualmachineinstances/${VMI_NAME}/vnc-token
+```
+
+Parameters:
+- `duration` - Time duration while the token is valid.
+
```
lyarwood, line 48: Can you include an example response here?

design-proposals/bound-token-for-vnc.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Scalability
+
+Creating a token requires an API call to the kubernetes API server.
```
lyarwood, line 52: It's also going to create a `ServiceAccount` etc per successful request right?

design-proposals/bound-token-for-vnc.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Update/Rollback Compatibility
+
+### Update
+A new version of `virt-controller` will create and watch `ServiceAccounts`, `Roles` and `RoleBindings` for each VMI
```
lyarwood, line 57: Why pre-create these resources, couldn't we do it per request?

lyarwood: Thanks but that's still a pre-created `ServiceAccount` per` VirtualMachineInstance` right? Why not only create the `ServiceAccount` with the first token request?

design-proposals/bound-token-for-vnc.md
akrejcir: Done

dominikholler:
@jobbler @rmohr are you interested in reviewing this design proposal?

design-proposals/bound-token-for-vnc.md
lyarwood: Apologies as I'm likely missing something here but couldn't the reconcile logic be applied by the controller once the resource had been created by the api? Why does this only work if created by the controller? 

I don't want to bike shed on this point it just seems odd to pre-create `ServiceAccounts` that given the use case aren't going to be used by the majority of `VirtualMachineInstances`.

design-proposals/bound-token-for-vnc.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+Parameters:
+- `duration` - Time duration while the token is valid.
+
+Response definition:
+```go
+type VncTokenResponse struct {
+	Token string `json:"token"`
+	ExpirationTimestamp metav1.Time `json:"expirationTimestamp"`
```
rmohr, line 53: I suppose we would return a temporary service account token with `expirtionSeconds` set, so that even if our controllers are down, we don't end up in a situation where the token may still be valid?

akrejcir: Yes, that is how it will be implemented. The token will have `expirationSeconds` and `boundObjectRef` set.

The `expirationTimestamp` in this response object is only for information, so that user doesn't need to decode the token.

rmohr: Great. Maybe mention that in the doc. I think it is a key design-prerequisite, so that it is ok to go this route :)

Please also mention cleanup.

design-proposals/bound-token-for-vnc.md
rmohr: There is one big concern which I have here.

Just grabbing our `virt-controller` cluster-role:

```yaml
rules:
- apiGroups:
  - policy
  resources:
  - poddisruptionbudgets
  verbs:
  - get
  - list
  - watch
  - delete
  - create
  - patch
- apiGroups:
  - ""
  resources:
  - pods
  - configmaps
  - endpoints
  verbs:
  - get
  - list
  - watch
  - delete
  - update
  - create
- apiGroups:
  - ""
  resources:
  - events
  verbs:
  - update
  - create
  - patch
- apiGroups:
  - ""
  resources:
  - pods/finalizers
  verbs:
  - update
- apiGroups:
  - ""
  resources:
  - pods/eviction
  verbs:
  - create
- apiGroups:
  - ""
  resources:
  - pods/status
  verbs:
  - patch
- apiGroups:
  - ""
  resources:
  - nodes
  verbs:
  - get
  - list
  - watch
  - update
  - patch
- apiGroups:
  - apps
  resources:
  - daemonsets
  verbs:
  - list
- apiGroups:
  - apps
  resources:
  - controllerrevisions
  verbs:
  - watch
  - list
  - create
  - delete
- apiGroups:
  - ""
  resources:
  - persistentvolumeclaims
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - delete
  - patch
- apiGroups:
  - snapshot.kubevirt.io
  resources:
  - '*'
  verbs:
  - '*'
- apiGroups:
  - pool.kubevirt.io
  resources:
  - virtualmachinepools
  - virtualmachinepools/finalizers
  verbs:
  - watch
  - list
  - create
  - delete
  - update
  - patch
  - get
- apiGroups:
  - kubevirt.io
  resources:
  - '*'
  verbs:
  - '*'
- apiGroups:
  - subresources.kubevirt.io
  resources:
  - virtualmachineinstances/addvolume
  - virtualmachineinstances/removevolume
  - virtualmachineinstances/freeze
  - virtualmachineinstances/unfreeze
  - virtualmachineinstances/softreboot
  verbs:
  - update
- apiGroups:
  - cdi.kubevirt.io
  resources:
  - '*'
  verbs:
  - '*'
- apiGroups:
  - k8s.cni.cncf.io
  resources:
  - network-attachment-definitions
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - apiextensions.k8s.io
  resources:
  - customresourcedefinitions
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - authorization.k8s.io
  resources:
  - subjectaccessreviews
  verbs:
  - create
- apiGroups:
  - snapshot.storage.k8s.io
  resources:
  - volumesnapshotclasses
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - snapshot.storage.k8s.io
  resources:
  - volumesnapshots
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - delete
- apiGroups:
  - storage.k8s.io
  resources:
  - storageclasses
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - flavor.kubevirt.io
  resources:
  - virtualmachineflavors
  - virtualmachineclusterflavors
  - virtualmachinepreferences
  - virtualmachineclusterpreferences
  verbs:
  - list
  - watch
- apiGroups:
  - migrations.kubevirt.io
  resources:
  - migrationpolicies
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - ""
  resources:
  - namespaces
  verbs:
  - get
```

you can see that we don't need access in any form to ServiceAccounts`, `Role` or `RoleBinding`. What we basically allow the virt-controller SA now, is giving any service account in any namespace the same amount of permissions which it has itself.

I would as an alternative consider the following design:

 * e.g. maintainer temporary in-memory tokens in virt-handler, which can be created and retrieved over `vnc-token`
 * add a vnc endpoint which can be accesses by a SA, hand out these sa credentials  to any user
 * in the vnc endpoint, when reaching virt-handler, validate that the token is valid

This would not require more privileges. Maybe other options which don't require privilege increase, exist too.

akrejcir: What do you think about this idea:
- VMI will have a special annotation (for example `vnc-token.kubevirt.io: true`) that will signal to `virt-controller` that it should create and watch the `ServiceAccount` and other resources for this VMI. This annotation will not be there by default.
- On request to `/vnc-token` endpoint, `virt-api` will add this annotation, then wait until the `ServiceAccount` is created and create a token.

rmohr: Hm, not sure that helps, since virt-controller would still need permissions to service accounts and so forth. With the possible alternaitve, e.g. virt-operator can create this permanent service account with basically no permissions. Just to avoid DOS on a possible token-based endpoint. If DOS is not a concern (like apiserver would only be by accident public accessible, and there still the token would prevent active misuse), we may not even need that.

rmohr: What I try to see, is if there is a way to provide this feature without increasing virt-controller privileges.

akrejcir: I have removed my previous comment, because I forgot to refresh this page before I posted it and it referred to Lee's comment.

@rmohr, how can we pass the token used for authentication with k8s API into `virt-handler` to check if it is correct?
Can you maybe point me to some place in the code, or explain how could I implement it?

akrejcir: Or how about this idea. The token has `audiences` field which could contain the UID of the VMI that it can access.

https://github.com/kubevirt/kubevirt/blob/692d3801a89413a417e1d7553e469f4010b6779a/vendor/k8s.io/api/authentication/v1/types.go#L159

Then `virt-api` on receiving `/vnc` request, could check if the UID from the token is equal to the UID of the VMI. This means that we don't need to create and watch `ServiceAccounts` for every VMI. There would be a single `ServiceAccount` with only access to `/vnc` and from it all tokens would be created.

Now I need to figure out how to access the token audiences from `virt-api`.

rmohr: @akrejcir just pushed an experiment I did a while ago to https://github.com/rmohr/kubevirt/pull/19. It roughly implements what I mean. Let me know what you think.

akrejcir: I see, thanks.

I couldn't find a way to access the bound ServiceAccount token that is used for authentication from `virt-api`, so my idea in the previous comment is not possible.

So I think we can use your idea, to generate random tokens in `virt-handler`. I will change the proposal.

rmohr: So, you don't just have to take the proposal. There are some points to clarify:

 * for convenience it would be best, if this extra REST endpoint is reachable without auth just with the token, is that ok from a security perspective?
 * if the VM migrates, a new token would have to be generated. So sometimes you may have the issue that you ask for a token which is valid for 30 seconds, but it is already invalid after 1 second.
 
 I would therefore list it as an alternative implementation and we can discuss both, instead of just replacing your proposal.
The limitations above may not be much of an issue but they really depend on the actual usage-pattern which we want to provide.

Like, you may want to hand out tokens which are valid for one week. That would not be possible. If you want to just hand out via a user-portal temporary access tokens which are valid for 1 minute, so that retrieval and usage can function properly (e.g. what oVirt does with the qemu tokens), then we are very likely good.

design-proposals/bound-token-for-vnc.md
```diff
@@ -0,0 +1,110 @@
...
+
+An administrator may want to give VNC access to a client application without giving it permissions to access
+anything else in the cluster. The VNC client could be pointed to VMI `/vnc` subresource and use
+the token to authenticate. 
+
+## Goals
+
+- A temporary token can be created to access VNC.
+- It can only be used with one specific VM.
+- It is valid only for a specific duration.
```
jobbler, line 21: Will there be a way to specify no expiration?

For example, if I am an admin and I create VMs for users to use as desktops, I would not want to keep recreating the tokens.
Or if a user needs to run other applications inside the VM for their normal job. Such as they have a linux desktop on their laptop for daily use but sometimes need to run an app inside windows.

I suspect a method to allow the user to renew their token would be a potential security issue.

design-proposals/bound-token-for-vnc.md
```diff
@@ -0,0 +1,110 @@
...
+
+## User Stories
+
+- A VM owner wants to give access to VNC to a third party application without giving it permissions to modify the VM.
```
jobbler, line 25: I assume this is meant to not allow permissions to modify the VM definition inside OpenShift, not the contents of the OS inside the VM?

design-proposals/bound-token-for-vnc.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Functional Testing Approach
+
+These cases should be tested:
+- Token can be created.
+- Token can be used to access VNC.
+- Token expires after a set time.
+- Token cannot be used to access VNC for a different VMI.
+- Token cannot be used to access any other resource.
+
```
geetikakay, line 105: Do we support multiple parallel sessions using same token?

design-proposals/bound-token-for-vnc.md
```diff
@@ -0,0 +1,110 @@
...
+### Rollback
+The new resources will have `ownerReferences` pointing to the VMI, so even on rollback, they will be garbage collected
+when VMI is deleted.
+
+## Functional Testing Approach
+
+These cases should be tested:
+- Token can be created.
+- Token can be used to access VNC.
+- Token expires after a set time.
```
geetikakay, line 102: if we create bulk tokens is there a chance of tokens id's getting repeated?

design-proposals/bound-token-for-vnc.md
```diff
@@ -0,0 +1,110 @@
...
+- Token cannot be used to access VNC for a different VMI.
+- Token cannot be used to access any other resource.
+
+# Implementation Phases
+
+1. Add functionality to `virt-controller` to create `ServiceAccount`, `Role` and `RoleBinding` for every VMI.
```
geetikakay, line 108: if two vmi are in same namespace, then also they need separate sa

design-proposals/bound-token-for-vnc.md
rmohr: > For example, if I am an admin and I create VMs for users to use as desktops, I would not want to keep recreating the tokens.

VNC over websockets in KubeVirt is explicitly not recommended for desktop usage. Just to make that clear, while it is great if you can use it that way for your use cases. :) 

The vga adapter has for instance a very small hardcoded amount of memory and since this  goes over the apiserver it may have unintended side effects if this is used by a bigger group of engineers as real desktop over this VNC session.

> For example, if I am an admin and I create VMs for users to use as desktops, I would not want to keep recreating the tokens.

I am not sure if this proposal is covering that use-case. @akrejcir ? I would expect that for permanent access, you can already right now create a ServiceAccount, Role and RoleBinding and maintain it like all other RBAC permissions for your cluster. The VNC token idea may sound appealing initially because you don't have to create these objects, bit in practice it probably means for admins more hassle to manage permanent RBAC access of some features without direct yaml applying.

What do you think?

design-proposals/bound-token-for-vnc.md
akrejcir: Correct. I will clarify it.

design-proposals/bound-token-for-vnc.md
akrejcir: Maybe, if we try to create more tokens for the same VMI with exactly the same expiration time. It will depend on the implementation.

design-proposals/bound-token-for-vnc.md
akrejcir: Correct, each SA would only have permissions to access one VMI.
But maybe we will implement a different design, I will update this document with alternatives today.

design-proposals/bound-token-for-vnc.md
rmohr: Just to clarify KubeVirt only allows one active session per VM. So you can have one session to VM1 and one session to VM2, but not two to VM1 at the same time.

design-proposals/bound-token-for-vnc.md
akrejcir: I have pushed a new change with three design alternatives.

design-proposals/bound-token-for-vnc.md
akrejcir: Done.

design-proposals/bound-token-for-vnc.md
geetikakay: @akrejcir please do share how you plan to implement. it will be nice to know and based on that testing can be done.

dominikholler:
@akrejcir Would it be useful to add a link to https://github.com/kubevirt/kubevirt/issues/8683 in the design document? This way it would be possible to trace from to design to the PR.

akrejcir:
I've updated the proposal.

We don't need to create a new VNC endpoint (`/vnc-access`) to use the token. The existing `/vnc` can accept the token and use it for authentication. If the token parameter is not present, the endpoint will use username and group for authentication.

design-proposals/bound-token-for-vnc.md
akrejcir: We need to discuss which implementation we will use.

design-proposals/bound-token-for-vnc.md
akrejcir: Yes, it does not cover the use-case of permanent tokens. They would need a different implementation.
I think for now we can keep this design only for temporary tokens, and permanent access can be created using RBAC, as you've mentioned.

design-proposals/token-for-vnc.md
```diff
@@ -0,0 +1,158 @@
...
+### Drawbacks
+- Token will not work when a VM is migrated, because the other `virt-handler` does not have the token in memory.
+- If `virt-handler` is down, the token will be invalidated.
+
+## API Examples
+
+### Getting the VNC token
+```
+GET /apis/subresources.kubevirt.io/v1/namespaces/${VMI_NAMESPACE}/virtualmachineinstances/${VMI_NAME}/vnc-token
+```
```
dominikholler, line 93: What permissions are required to trigger this call?

rmohr (commented):
Checking all the implementations and pros and cons  (and as you could see I tried to solve this myself already), I am really not sure if we have a critical mass to solve this within core kubevirt.

When debating with myself, at some point I always end at the point where I think that "core" KubeVirt should not weaken the k8s RBAC model, and something like this may have to live outside of "core" kubevirt.

rmohr:
> If the VNC token API would be implemented in a new service, it would probably need to forward the VNC connection to `virt-api` which would then forward it to the VM. Wouldn't that be too much overhead?

I am not even sure if that traffic should go through the apiserver, if accessing VNC is the only request. It may be more a service-ip service in-between? But it would be possible for an external service inside the cluster to directly talk to virt-handler if useing the virt-api certificate secret.

dominikholler:
> I am not even sure if that traffic should go through the apiserver, if accessing VNC is the only request. It may be more a service-ip service in-between? 
> But it would be possible for a service inside the cluster to directly talk to virt-handler if useing the virt-api certificate secret.

Even the users expected to use this feature are probable quite skilled, I would try to route the traffic trough the API server to avoid any additional configuration which could break the feature.

> We are really stretching the boundaries of the k8s security model here a little bit (and how apiserver access is supposed to be used) :/

I agree, but the api server trouble-free accessible. Do you have any other ideas which does not require any assumptions about networking?

rmohr:
> > I am not even sure if that traffic should go through the apiserver, if accessing VNC is the only request. It may be more a service-ip service in-between?
> > But it would be possible for a service inside the cluster to directly talk to virt-handler if useing the virt-api certificate secret.
> 
> Even the users expected to use this feature are probable quite skilled, I would try to route the traffic trough the API server to avoid any additional configuration which could break the feature.

If you have a customer which needs, that I would not necessarily suggest that the customer does this themselves. I am just not sure if we have a broader use-case which justifies this bending as part of KubeVirt. Kubernetes is pretty modular and this is something which could be perfectly added on top.

> 
> > We are really stretching the boundaries of the k8s security model here a little bit (and how apiserver access is supposed to be used) :/
> 
> I agree, but the api server trouble-free accessible. Do you have any other ideas which does not require any assumptions about networking?

If it is not inside KubeVirt you have a lot of possibilities on how to solve this and how to route this:

 * apiserver
 * loadbalancer
 * your add-on service can use serviceaccount tokens, or any other form of token, this feature would then not in general elevate virt-controller privileges for a niece (?) use-case
 
I would tend to not integrate it into KubeVirt and build it on top (that is also why I did not propose my solution). But I would 
 1. love to hear opinions of other maintainers as well as a broader understanding of how needed this is across KubeVirt adopters. 
 2. love to understand the scope better. The "I need permanent access" case is anyway already solved with ServiceAccounts how they are today, also admins can easily give time-boxed access (days, weeks, hours) with TTL on a ServiceAccount. So the only remaining case seems to be ticketing (getting a token, so that I can get as kind of an oauth token as legitimiation to establish the connection actually through the really desired channel).

Thoughs?

fabiand:
loadbalancers, and service accounts, don't solve the issue by itself.
But I agree that we can solve it external to kubevirt.

akrejcir:
I've updated the file to describe a VNC proxy service, external from kubevirt.
It does not describe how to expose the service.  On OpenShift, we can use `Route`.

Should I close this PR, when this will not be in core kubevirt?

rmohr:
> I've updated the file to describe a VNC proxy service, external from kubevirt. It does not describe how to expose the service. On OpenShift, we can use `Route`.
> 
> Should I close this PR, when this will not be in core kubevirt?

If you think we had a sufficient discussion around this and you think it does not belong into core kubevirt, I would say yes.

PR #195, opened 2022-10-21
Adding demo video and transcript for VM export

aburdenthehand:
This demo was provided by @awels and I've included a transcript for it based on his notes.

Signed-off-by: Andrew Burden <aburden@redhat.com>

### DIFF
--- a/demos/VM_Export/demoscript_Oct-VM-export.txt
+++ b/demos/VM_Export/demoscript_Oct-VM-export.txt
@@ -21,8 +21,9 @@ What does this mean?
 
 All PVCs, data volumes, and memorydump volumes on the VM will be exported.
 (No other volume types will be exported)
-Also, when you create a VM export, a dedicated export proxy server is created which can then be exposed to the outside world with an Ingress/Route or NodePort.
-The dedicated export server is created so that this data transfer does not involve and potentially overload the Kubernetes API server: the proxy connects to the export server and the user connects to the proxy. 
+Also, when you create a VM export, a dedicated export server is created. 
+The export server interacts with a proxy, which is created when you first enable the VM export feature gate, and the proxy interacts with an Ingress/Route or NodePort.
+This eliminates the Kubernetes API from the process.
 This server has its own certificates, which we will see later in the VM export object.
 
 <0:38> 
--- a/demos/VM_Export/demoscript_Oct-VM-export.txt
+++ b/demos/VM_Export/demoscript_Oct-VM-export.txt
@@ -1,0 +1,81 @@
+VM Export
+<Intro>
+In this demo we will be exporting a VM disk, and having a quick look at the behaviour of the process. 
+
+Why? We might use VM Export so that it can be imported into another cluster, platform, or system. 
+
+<Start>
+We start with a KubeVirt cluster. 
+This cluster has the virtctl client which includes additional commands to interact with VMs. 
+And we have a stopped Fedora 36 VM that we want to export.
+On the left we have our command pane. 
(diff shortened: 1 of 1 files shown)


jean-edouard:
Transcript is perfect, thank you for this!
Just a suggestion for these demos, if the transcript was a `README.md`, github would show it nicely under the video link, just a thought.
/lgtm

demos/VM_Export/demoscript_Oct-VM-export.txt
```diff
@@ -0,0 +1,81 @@
...
+<0:15>
+To export the VM disk, we use virtctl with the 'vmexport create' option, and specify the name of the VM.
+And we see the VM export has been created and is pending as the pod is created.
+When the pod is running, we can see that the VM export is now ready.
+
+What does this mean?
+
+All PVCs, data volumes, and memorydump volumes on the VM will be exported.
+(No other volume types will be exported)
+Also, when you create a VM export, a dedicated export proxy server is created which can then be exposed to the outside world with an Ingress/Route or NodePort.
```
awels, line 24: Small detail `Also, when you create a VM export, a dedicated export proxy server is created which can then be exposed to the outside world with an Ingress/Route or NodePort.` is not technically incorrect, but the proxy is created when you enable the feature gate. The export server is created when you create a VM export, the proxy then talks to the export server and the route/ingress talks to the proxy. This eliminates the k8s API from the process.

awels (commented):
looks good, one minor technical detail.

demos/VM_Export/demoscript_Oct-VM-export.txt
aburdenthehand: Thanks @awels - does this now accurately explain it?

rmohr:
@aburdenthehand I am fine with adding it here, but wonder if the youtube channel (possibly also with links on kubevirt.io) may not reach a broader audience instead.

rmohr:
Actually I am not sure I am fine with that. After this PR gets merged I have to download 40+MB on a git clone :/

aburdenthehand:
@rmohr This material is more intended as a demo library for KubeVirt folks to then use/promote at events/meetups. They require some form of narration. 
You raise a good point that I had not considered. AFAIK you would only have to download it once, as the video would not change with time, but I was intending to have a variety of demo videos available here. Collectively I suppose this could increase to a couple hundred MB. And over time these might need to be updated or replaced as KubeVirt and the features evolve.
If that size is a burden on the repo then I need to rethink this.

rmohr:
> You raise a good point that I had not considered. AFAIK you would only have to download it once, as the video would not change with time, but I was intending to have a variety of demo videos available here. Collectively I suppose this could increase to a couple hundred MB. And over time these might need to be updated or replaced as KubeVirt and the features evolve.
> If that size is a burden on the repo then I need to rethink this.

Also hosting that on the CI GCP store may be an option, but a non-official youtube channel may be fine too? If think it would be better to not have such big files here in the repo. What do others think?

rmohr:
/hold

let's figure out a way to only add links to the definitely valuable videos.

jobbler:
@rmohr @aburdenthehand I am partial to hosting the videos on Youtube. It seems to be a place that a lot of people search. I think this information is valuable to an admin wanting to understand how things work. It makes it easier for them to figure out how to use the cluster to its full potential and also troubleshoot when things go wrong.  Even if this and future videos are meant to be as a demo library, I see a lot of value to cluster admins and think the wider exposure would be good.

awels:
Did we ever put this stuff in a youtube channel?

PR #198, opened 2022-11-13
Proposal: Clone from existing snapshot instead of new one for each clone

akalenyu:
Smart cloning today creates a VolumeSnapshot for any new clone, following PoC, we identified some storages prefer 1:N.

Signed-off-by: Alex Kalenyuk <akalenyu@redhat.com>

### DIFF
--- a/design-proposals/clone-from-existing-snapshot.md
+++ b/design-proposals/clone-from-existing-snapshot.md
@@ -1,0 +1,92 @@
+# Overview
+Today's smart cloning methods use a 1:1 cloning method which creates a temporary snapshot of the source PVC for each intended clone.  
+Some storage systems (possibly ceph) are designed to scale better with a 1:N method where a single snapshot of the source PVC is used to create multiple clones.  
+
+This design presents a possible way of implementing a flow where a snapshot is created for a golden image and then subsequent CDI smart clones use that snapshot as the PVC DataSource.
+
+## Motivation
+Following a [PoC](https://github.com/kubevirt/containerized-data-importer/pull/2355), we have identified a performance improvement that reinforces the above claim for ceph,  
+and would like to productize that PoC so other storage systems could capitalize on this.
+
+## Goals
+* StorageProfile cloneStrategy will be added to tip off CDI about which storage can capitalize on this type of cloning
+* (?) DataVolume will provide an API to opt-in/out of cloning from a cached snapshot
+
+## Non Goals
+* Invalidation of source DataVolumes' cached snapshots as these are defined to be immutable by [design](https://github.com/kubevirt/community/blob/main/design-proposals/golden-image-delivery-and-update-pipeline.md?plain=1#L11)
+
+## Definition of Users
+This feature is intended for cluster admins who wish to provide boot sources for their users to create VMs from,  
+in the most suiting manner for their storage.
+
+## User Stories
+* As a KubeVirt user, I want to store a base image as a snapshot and clone from the snapshot instead of a pvc because this scales better on the storage systems used.
+
+## Repos
+* **containerized-data-importer**: DataVolume, Snapshot, StorageProfile controllers
+
+# Design
+Today we already have a mechanism in the form storage profiles that detects the preferred cloning strategy (`copy/snapshot/csi-clone`).  
+The proposition is to extend this mechanism with a new `cached-snapshot` strategy.  
+Upon dealing with target DataVolumes that request to clone a certain existing volume, we would calculate whether we should use the new cloning strategy.  
+If that is the case, we will lazily decide if we should create a snapshot of the source.
+
+## Snapshot of source does not exist
+We create a snapshot and continue the smart clone flow as usual.
+
+## Snapshot of source exists
+Proceed to create a PVC out of this snapshot.
+
+## Controller details
+Today the datavolume controller passes the responsibility of creating a snapshot to the smart clone controller by creating a VolumeSnapshot object.  
+Since we won't be creating a snapshot object at all times, a change is needed.  
+The proposed change is to have the smart clone controller watch certain DataVolume fields (annotation/API) and act on their appearance.
+
+## DataVolume level API (?)
+No final decision whether we want DataVolume API so would appreciate any input!
+
+We could opt for a seamless implementation via StorageProfiles,
+and not introduce DataVolume APIs.
+
+However, if we do,
+```yaml
+kind: DataVolume
+metadata:
+  name: clone-from-snap-target
+spec:
+  source:
+    snapshot:
+      namespace: snap-ns
+      name: snap-name
+```
+Or directly introduce with sourceRef
+```yaml
+kind: DataVolume
+metadata:
+  name: clone-from-snap-target
+spec:
+  sourceRef:
+    kind: DataSource
+    name: snap-source
+```
+
+
+## API Examples
+(tangible API examples used for discussion)
+
+## Scalability
+During PoC testing we discovered an error that stops progress after 700~ clones regarding JWT tokens
+```bash
+error verifying token: square/go-jose/jwt: validation failed, token is expired (exp)
+```
+
+## Update/Rollback Compatibility
+Clone operations on new installs will opt-in to this type of cloning if we know for fact it scales better on the storage provisioner  
+(Following PoC, we know this holds for ceph rbd).
+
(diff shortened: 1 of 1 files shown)


design-proposals/clone-from-existing-snapshot.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+
+## API Examples
+(tangible API examples used for discussion)
+
+## Scalability
+During PoC testing we discovered an error that stops progress after 700~ clones regarding JWT tokens
+```bash
+error verifying token: square/go-jose/jwt: validation failed, token is expired (exp)
```
mhenriks, line 80: bugzilla for this?

design-proposals/clone-from-existing-snapshot.md
```diff
@@ -0,0 +1,92 @@
...
+## Snapshot of source exists
+Proceed to create a PVC out of this snapshot.
+
+## Controller details
+Today the datavolume controller passes the responsibility of creating a snapshot to the smart clone controller by creating a VolumeSnapshot object.  
+Since we won't be creating a snapshot object at all times, a change is needed.  
+The proposed change is to have the smart clone controller watch certain DataVolume fields (annotation/API) and act on their appearance.
+
+## DataVolume level API (?)
+No final decision whether we want DataVolume API so would appreciate any input!
```
mhenriks, line 46: My preference would be to start with updating DataVolume/DataSource to support snapshot source.  Then MAYBE add caching for PVC source later.

Snapshot is a better source of truth than PVC because snapshots are immutable.  My feeling is that they should be promoted as THE way to organize golden images.

design-proposals/clone-from-existing-snapshot.md
```diff
@@ -0,0 +1,92 @@
...
+      name: snap-name
+```
+Or directly introduce with sourceRef
+```yaml
+kind: DataVolume
+metadata:
+  name: clone-from-snap-target
+spec:
+  sourceRef:
+    kind: DataSource
```
mhenriks, line 69: `DataSource` should be updated to refer to snapshot source

design-proposals/clone-from-existing-snapshot.md
mhenriks: May need to update our token parameters for explicit snapshot source.  I believe "PersistentVolumeClaim" is baked in there somewhere.

May also have to reconsider the implicit rbac check.  Currently check that user can create pods.  For VolumeSnapshot source, should probably be "create pvcs and pods". 

For explicit rbac check, there is "datavolumes/source." maybe that is good enough?  Or should consider different sources (pvc/snapshot).

design-proposals/clone-from-existing-snapshot.md
```diff
@@ -0,0 +1,92 @@
...
+(tangible API examples used for discussion)
+
+## Scalability
+During PoC testing we discovered an error that stops progress after 700~ clones regarding JWT tokens
+```bash
+error verifying token: square/go-jose/jwt: validation failed, token is expired (exp)
+```
+
+## Update/Rollback Compatibility
+Clone operations on new installs will opt-in to this type of cloning if we know for fact it scales better on the storage provisioner
```
mhenriks, line 84: Yeah caching should be opt-in

mhenriks (commented):
Perhaps this is a controversial opinion but I am in favor of an explicit snapshot source.  And golden images should organized based on snapshots since they are immutable.

design-proposals/clone-from-existing-snapshot.md
akalenyu: > bugzilla for this?

So this only happens with the PoC (unmerged code), as with CDI today we cannot get so many clones due to ceph flattening issues.

Are you suggesting that this could be a hidden problem in today's CDI? just well hidden since we couldn't successfully spin up 700~?

design-proposals/clone-from-existing-snapshot.md
akalenyu: > Then MAYBE add caching for PVC source later.

Not sure I got that


> Snapshot is a better source of truth than PVC because snapshots are immutable. My feeling is that they should be promoted as THE way to organize golden images.
This idea makes sense to me but it would be a pretty major change wrt golden images;
What if certain storages don't perform better this way? or don't support snapshots at all?

akalenyu:
> Perhaps this is a controversial opinion but I am in favor of an explicit snapshot source. And golden images should organized based on snapshots since they are immutable.

Yeah not controversial at all, I think it's the natural thing that comes to mind.
Just a different approach compared to the seamless smart clone using the storageprofile preferred strategy.
Correct me if I got it wrong, but going forward with this approach, there will be no auto snapshot creation, so
- Users are expected to create their own
- We might have DataImportCron's create some golden snapshots for convenience

/cc @aglitke 
would appreciate your input here as well

design-proposals/clone-from-existing-snapshot.md
akalenyu: You mean extended to support something other than PVCs?

akalenyu:
/cc @salanki
IIUC some CoreWeave folks were interested in this, so feel free to share input!

design-proposals/clone-from-existing-snapshot.md
mhenriks: >Are you suggesting that this could be a hidden problem in today's CDI? just well hidden since we couldn't successfully spin up 700~?

My comments regarding the token above are related to security, not scalability.  Don't want to be able to reuse a token created for PVC source with snapshot.

design-proposals/clone-from-existing-snapshot.md
mhenriks: > Not sure I got that, sorry

Isn't this design mostly about caching a snapshot of a PVC which will be used for DataVolume cloning?  I.e `cached-snapshot` cloning strategy?

Basically, I don't think we should be in the caching biz unless there is a really good reason.  Because if you are in the caching biz, then you are in the cache invalidation biz.  And that is not fun!  How would we detect/handle the case when the PVC is updated?

> What if certain storages don't perform better this way? or don't support snapshots at all?

DataImportCron will have to be smart enough to handle both cases and create the appropriate resources.  The change shouldn't matter to consumers as they just care about the DataSource.

> WDYT About ReadOnly PVCs btw? We agreed on immutable golden images in the design but never implemented
anything to reflect that. I guess users could specify this access mode themselves, or we could do that since we opinionate them

ReadOnly PVCs are interesting, DataImportCron can take advantage of them when there's no snapshot support, but I don't see any advantages for them otherwise.  And they obviously have the scale disadvantage that was the motivation for this design

design-proposals/clone-from-existing-snapshot.md
mhenriks: Yeah something like that

design-proposals/clone-from-existing-snapshot.md
akalenyu: > Isn't this design mostly about caching a snapshot of a PVC which will be used for DataVolume cloning? I.e `cached-snapshot` cloning strategy?
> 
> Basically, I don't think we should be in the caching biz unless there is a really good reason. Because if you are in the caching biz, then you are in the cache invalidation biz. And that is not fun! How would we detect/handle the case when the PVC is updated?

Yeah, it is, just not sure what `caching for PVC source` is or looks like.

I was thinking of not being in the caching biz, maybe we could change the wording of the clone strategy.
Basically, just create new DVs if you want new snapshot sources, which seemed to fit nicely with the design of DataImportCron sources (immutable, but nothing enforces this)

arnongilboa: > Basically, just create new DVs if you want new snapshot sources, which seemed to fit nicely with the design of DataImportCron sources (immutable, but nothing enforces this)

Right. Adding snapshot source to DataSource and DataVolume seems like the cleanest way to go.

design-proposals/clone-from-existing-snapshot.md
akalenyu: > IIRC long term tokens are only created for host assisted clones ATM

Yeah this subthread is starting to make much more sense now that I am getting properly acquainted with this side of CDI
So yeah long term token are only for host assisted so probably went over 5 min when we were doing 700~ clones stress test for the PoC

PR #203, opened 2023-01-16
Proposal:Add ability to find best match for CPU Model dynamically

Barakmor1:
Calculates the newest CPU Model that is supported in by enough nodes that
respect vmi's `nodeAffinity` and `nodeSelectors`.
POC: https://github.com/kubevirt/kubevirt/pull/9060
Signed-off-by: bmordeha <bmodeha@redhat.com>

### DIFF
--- a/design-proposals/best-match-cpu-model.md
+++ b/design-proposals/best-match-cpu-model.md
@@ -1,0 +1,165 @@
+
+# Overview
+Add ability to find best match for CPU Model dynamically after a vmi is scheduled to a node.
+
+The story is about calculating non-obsolete CPU Model 
+that is supported by enough nodes ( * ) that respect
+VMI's nodeAffinity and nodeSelectors.
+
+( * ) Number that is high enough for the VMI's migratability
+      but not too high to avoid using obsolete CPU Model that 
+      is more likely to be widely supported.
+
+    
+This will allow users to choose CPU Model configuration in
+`KubevirtCR` or in`vmi.spec.domain.CPU.model` that will result
+dynamic calculation of CPU Model that is relatively
+new and migratable in the cluster.
+
+That can be useful when the user's cluster is heterogeneous.
+
+## Motivation
+In KubeVirt, the default CPU model is `host-model` 
+which is highly coupled to the initial node of a VMI.
+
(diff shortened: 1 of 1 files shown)


jean-edouard (approved):

design-proposals/best-match-cpu-model.md
```diff
@@ -0,0 +1,165 @@
...
+- When using the CPU Model matcher a VMI's CPU Model will be calculated **AFTER** virt-launcher is
+scheduled to a node.
+
+**Calculation** : First we will get all the potential nodes that respect the VMI's node affinity, 
+node selectors and has the same vendor as the initial node.  
+Then for each CPU Model that has newer launch year then `MinimalLaunchYear` that is supported 
+on the initial node we will calculate how many potential nodes support it.  
+We will define a `ThresholdSupport` and we will choose the CPU Model that is
+supported by at least `ThresholdSupport` of the cluster's node with 
+the newest launch year.
```
iholder101, line 97: What if multiple CPU models have the same rank, i.e. supported by the same amount of nodes? Would we take the oldest model?

iholder101 (commented):
Thanks @Barakmor1! Awesome to see this, I like this approach a lot!

host-model is indeed problematic w.r.t migrations in an heterogeneous cluster. I think such an approach is much needed.

A few notes:
1) IIUC choosing a named model over host-model is slightly less performant and might cause small penalties for the guest. Perhaps it's worth mentioning that.
2) Considering the point above - do we want this to be the default model? Or should we stay with host-model as a default?
3) IIUC, as long as the guest doesn't use CPU features that its named model doesn't support there is no different w.r.t. performance penalty between newer and older CPU models (@vladikr: I'd appreciate if you keep me honest here). If that's true, I think we should give precedence to older models as they have a higher probability of being supported by other (perhaps dynamically added) nodes.

aburdenthehand:
Hello @Barakmor1
What's the state of this proposal, and do you need anything to get it merged?

Barakmor1:
@aburdenthehand Hey,
 I'll be tackling the Proposal shortly. Apologies for the hold-up; I've been engaged with an exciting project named @kubevirt/application-aware-quota  which has taken up most of my time.

PR #212, opened 2023-03-02
proposal: Live resize vCPU and Memory

kvaps:
related issues:

- https://github.com/kubernetes/kubernetes/pull/102884
- https://github.com/kubevirt/kubevirt/issues/398
- https://github.com/kubevirt/kubevirt/issues/8787

PR #216, opened 2023-04-11
CPU Hotplug design proposal

jean-edouard:
This design proposal is an alternative to https://github.com/kubevirt/community/pull/212
The key differences are:
- This proposal focuses on CPU hotplug, keeping in mind that the API will most likely be extended later for memory hotplug
- The implementation doesn't use the new Kubernetes vertical scaling API, as it not yet ready for production (currently in alpha stage). Instead, a live migration is used to scale the CPU resources up/down. However in the future, we definitely want to switch to using the vertical scaling API
- There's already a WIP/POC implementation of the design: https://github.com/kubevirt/kubevirt/pull/9401

### DIFF
--- a/design-proposals/cpu-hotplug.md
+++ b/design-proposals/cpu-hotplug.md
@@ -1,0 +1,164 @@
+# Overview
+This proposal is about adding CPU hotplug support to KubeVirt.
+CPU hotplug will allow users to dynamically add/remove CPU resources to/from running VMs.
+While this design is focused on CPU hotplug, it aims to provide a generic API design for resource hotplug that will be reusable for things like memory.
+
+## Motivation
+CPU hotplug is now a common virtualization feature, but KubeVirt lacks support for it.
+Live vertical scaling of VM resources in general can be very useful in a variety of scenarios.
+
+## Goals
+- Hotplug and hot-unplug of CPU resources from VMs should be achievable by editing the VM object itself (no sub-resources)
+- Underlying technology used for hotplug should be independent of the VM API, meaning we should be able to swap out how hotplug is technically performed in the future without impacting API compatibility
+- Implementation should be achievable today, with Kubernetes APIs that are at least in beta. Unfortunately, at the time of writing, the Kubernetes vertical pod scaling API is only alpha
+
+## Non Goals
+- Revising existing hotplug mechanisms like Volume hotplug
+- Dynamically applying affinity rules to VMIs
+
+## Definition of Users
+A user with (at a minimum) create, update, and patch access for VM objects within a namespace.
+
+## User Stories
+As a VM user, I would like to dynamically increase or decrease the number of CPUs used by a VM without requiring the VM to restart.
+
+## Repos
+kubevirt/kubevirt
+
+# Design
+In VM objects, the existing CPU `sockets` field will now be dynamic, provided that it has been declared as such.
+Declaring the field as dynamic will be done by adding a `cpu` entry to a new section under the VM spec, called `liveUpdateFeatures`.
+Under the `cpu` entry, it will be possible to define a maximum number of sockets. That number is needed by LibVirt and will default to 4 times the initial number of sockets if not set. That default value will be configurable in the KubeVirt CR.
+On VM startup, that value will translate to a new `maxSockets` entry under `spec.domain.cpu` in the VMI object. However, all VMI CPU fields will stay immutable, and CPU sockets will only be adjustable on the VM object.
+
+Increasing the number of CPU sockets will not only add CPUs to the guest but also increase the CPU resources available to the virt-launcher pod, via a migration (more on that in the Implementation section).
+The way LibVirt handles CPU hotplug is by exposing the maximum number of CPUs to the guest and turning off the unused CPUs. In this document, when we refer to "adding" CPUs to the guest, we're really just turning them on. This is why we need to define a maximum number of sockets in advance.
+
+## Drawbacks / Limitations
+- Since CPU hot-(un)plug involves a live migration, VMs will have to be `LiveMigratable` to enable CPU hotplug
+- Each disabled vCPU, so `(maxSockets - sockets) * cores * thread`, consumes 8MiB of overhead memory
+- This feature will be incompatible with CPU requests/limits, at least initially, since those values need to dynamically scale according to the current number of enabled CPUs
+
+## API Examples
+
+### VM Spec API
+
+```
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  labels:
+    kubevirt.io/vm: vm-cirros
+  name: vm-cirros
+spec:
+  running: false
+  liveUpdateFeatures:
+    cpu
+      maximumSockets: 4
+  template:
+    metadata:
+      labels:
+        kubevirt.io/vm: vm-cirros
+    spec:
+      domain:
+        cpu:
+          sockets: 2
+          cores: 4
+          threads: 2
+```
+In this example, the user will be able to double the guest CPU resources by increasing the value of `sockets`.
+
+### VMI Spec API
+
+The VM definition above would translate to the following VMI:
+
+
+```
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  labels:
+    special: vm-cirros
+  name: vm-cirros
+spec:
+  domain:
+    cpu:
+      maxSockets: 4
+      sockets: 2
+      cores: 4
+      threads: 2
+```
+
+### Status API
+
+The VMI status must indicate the current number of CPU sockets. The value needs to be propagated back to the VM status to allows the user to have an indication of when their changes applied to the running VMI.
+
+The user should be able to inspect the VM.Status value below to determine the state of their changes. This is similar in concept to someone inspecting a Deployments status to determine the the number of pods that are available.
+
+```
+// The number of sockets the vm has currently active.
+vmi.Status.Resources.Sockets.Ready
+vm.Status.Resources.Sockets.Ready
+```
+
+Conditions on the VM and VMI should indicate the status of the pending CPU hotplug actions. When errors occur, the conditions should propagate the errors to the VM.Status.Conditions field so users can receive feedback into the status of their declared state.
+
+```
+// VirtualMachineCPUReady indicates that the desired amount of cpu is active on the vmi
+VirtualMachineCPUReady VirtualMachineConditionType = "CPUReady"
+```
+
+When the desired state is equal to the ready state, then these conditions will have `Status=True`, otherwise `Status=False` with a human readable reason and message indicating whether the hotplug change is pending or encountering errors.
+
+## Extensibility
+We will be able to use the new `liveUpdateFeatures` section to manage things like memory hotplug in the future.
+
+## Update/Rollback Compatibility
+This is an entirely new feature, so updates from older versions will not be a problem.
+
+## Functional Testing Approach
+Functional tests will simply:
+- Create and start a VM with `cpu` set as one of the `liveUpdateFeatures`
+- Modify the number of CPU sockets for the VM
+- Ensure the guest sees the changed number of CPUs
+- Ensure the CPU resources on the virt-launcher pod were adjusted accordingly
+
+# Implementation Phases
+
+The first two phases will happen together in the same pull request (PR), since the first phase is mostly useless without the second one.
+In the unlikely case that hot-unplugging CPUs is more complicated than anticipated, hotplug-only could be submitted as a first PR, then hot-unplug later on as a second PR.
+The third phase will happen later as a separate PR. It should be able to transparently replace the migration mechanism without API changes.
+
+## Phase 1: Creating the KubeVirt API and linking it to the LibVirt CPU hotplug API
+## Phase 2: Triggering a live migration when the number of sockets changes, to adjust the virt-launcher CPU resources
+
+The VMI [workload update controller](https://github.com/kubevirt/kubevirt/blob/ce8aab7874d2e1586787e3e2a17306b7edca1b8a/pkg/virt-controller/watch/workload-updater/workload-updater.go#L476) is extended to live migrate VMIs to satisfy updates to a VMI resource request/limits field.
+
+Sequence of events:
+- User updates CPU sockets on a vm.spec.template.spec.domain.cpu
+- VM controller determines the request/limits required and writes the necessary changes to the active vmi.Spec
+- VMI workload update controller detects the mismatch between the resource request/limits on the pod vs vmi.Spec and live migrates the VMI to satisfy vmi.spec
+- Migration controller constructs a new target pod with the desired CPU request/limits
+- VMI live migrates
+- virt-handler triggers the hotplug/hot-unplug LibVirt action after the live migration completes
+- VM status is updated to reflect that the pending hotplug/hot-unplug actions have completed
+
+## Phase 3 (future): Using the [Kubernetes vertical scaling API](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources ) once it reaches a more mature state
+
+The inplace vertical pod feature allows us to dynamically change the VMI pods resource request/limits. This feature will allow us to avoid live migration, however the issue for us is timing. This feature is alpha in Kubernetes 1.27. We need it to be at least in beta to integrate it without risking stability and API reliability.
+
+Sequence of events
+- User updates CPU sockets on a vm.spec.template.spec.domain.cpu
+- VM controller writes these change to the active vmi.Spec
+- VMI controller updates the active pod spec to reflect the request/limit changes
+- Pod changes are applied
+- virt-handler triggers the hotplug/hot-unplug action after the inplace update completes.
+- VM status is updated to reflect that the pending hotplug/hot-unplug actions have completed
+
+# References
+
+- LibVirt Domain XML CPU hotplug
+https://libvirt.org/formatdomain.html#cpu-allocation
+- Kubernetes inplace pod updates
+https://github.com/kubernetes/kubernetes/pull/102884
+https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources
--- a/design-proposals/cpu-hotplug.md
+++ b/design-proposals/cpu-hotplug.md
@@ -30,12 +30,10 @@ In VM objects, the existing CPU `sockets` field will now be dynamic, provided th
 Declaring the field as dynamic will be done by adding a `cpu` entry to a new section under the VM spec, called `liveUpdateFeatures`. Please note that prior to defining this field the hotplug action would not take place i.e. changing the number of sockets will be staged until further reboot.
 Under the `cpu` entry, it will be possible to define a maximum number of sockets. That number is needed by LibVirt and will default to 4 times the initial number of sockets if not set. That default value will be configurable in the KubeVirt CR.
 On VM startup, that value will translate to a new `maxSockets` entry under `spec.domain.cpu` in the VMI object. However, all VMI CPU fields will stay immutable, and CPU sockets will only be adjustable on the VM object.
-Once the hoplug action begins, subsequent changes to the number of sockets would be rejected until the hotplug process completes.
-During the hotplug action a notifitcation will be created upon successful update of the CPU resources on the pod level. It will indicate that now it safe to proceed to the hotplug
+Once the hoplug action begins, subsequent changes to the number of sockets would be rejected until the hotplug process completes. CPU Resource update on the pod level must occur before the actual hotplug action on the Libvirt level, hence, during the hotplug action a notifitcation will be created upon successful update of the CPU resources on the pod level. It will indicate that now it safe to proceed to the hotplug
 action on the Libvirt/QEMU level. 
 
-Increasing the number of CPU sockets will not only add CPUs to the guest but also increase the CPU resources available to the virt-launcher pod, via a migration (more on that in the Implementation section). CPU Resource update on the pod level must occur before the actual hotplug action on the Libvirt level. 
-Therefore, a notifitcation will be created upon successful update of the CPU resources on the pod level. It will indicate that its safe to proceed with updating the libvirt domain with the new topology.
+Increasing the number of CPU sockets will not only add CPUs to the guest but also increase the CPU resources available to the virt-launcher pod, via a migration (more on that in the Implementation section).
 
 The way LibVirt handles CPU hotplug is by exposing the maximum number of CPUs to the guest and turning off the unused CPUs. In this document, when we refer to "adding" CPUs to the guest, we're really just turning them on. This is why we need to define a maximum number of sockets in advance.
 
--- a/design-proposals/cpu-hotplug.md
+++ b/design-proposals/cpu-hotplug.md
@@ -91,24 +91,23 @@ spec:
 
 ### Status API
 
-The VMI status must indicate the current number of CPU sockets. The value needs to be propagated back to the VM status to allows the user to have an indication of when their changes applied to the running VMI.
+The VMI status must indicate whether CPU hotplug is taking place. As the hotplug process begins the current state of `spec.domain.cpu` needs to be propagated back to the VM/VMI `status.currentCPUTopology` to allow the user to have an indication of the hotplug processing.
 
-The user should be able to inspect the VM.Status value below to determine the state of their changes. This is similar in concept to someone inspecting a Deployments status to determine the the number of pods that are available.
+During the hotplug process, the current CPU topology state will be reflected in VM/VMI `Status.currentCPUTopology` while the desired CPU topology state will be in `spec.Domain.CPU`. When the hotplug process completes the `Status.currentCPUTopology` will be cleared.
 
 ```
-// The number of sockets the vm has currently active.
-vmi.Status.Resources.Sockets.Ready
-vm.Status.Resources.Sockets.Ready
+	// CPU allow specified the detailed CPU topology inside the vmi.
+		CurrentCPUTopology *CPU `json:"currentCPUTopology,omitempty"`
 ```
 
 Conditions on the VM and VMI should indicate the status of the pending CPU hotplug actions. When errors occur, the conditions should propagate the errors to the VM.Status.Conditions field so users can receive feedback into the status of their declared state.
 
 ```
-// VirtualMachineCPUReady indicates that the desired amount of cpu is active on the vmi
-VirtualMachineCPUReady VirtualMachineConditionType = "CPUReady"
+	// Indicates that the VMI is in progress of Hot vCPU Plug/UnPlug
+	VirtualMachineInstanceVCPUChange VirtualMachineInstanceConditionType = "HotVCPUChange"
 ```
 
-When the desired state is equal to the ready state, then these conditions will have `Status=True`, otherwise `Status=False` with a human readable reason and message indicating whether the hotplug change is pending or encountering errors.
+Upon succesful hotplug action the condition will be removed. Otherwise, a human readable reason and message indicating whether the hotplug change is pending or encountering errors.
 
 ## Extensibility
 We will be able to use the new `liveUpdateFeatures` section to manage things like memory hotplug in the future.
--- a/design-proposals/cpu-hotplug.md
+++ b/design-proposals/cpu-hotplug.md
@@ -27,15 +27,21 @@ kubevirt/kubevirt
 
 # Design
 In VM objects, the existing CPU `sockets` field will now be dynamic, provided that it has been declared as such.
-Declaring the field as dynamic will be done by adding a `cpu` entry to a new section under the VM spec, called `liveUpdateFeatures`.
+Declaring the field as dynamic will be done by adding a `cpu` entry to a new section under the VM spec, called `liveUpdateFeatures`. Please note that prior to defining this field the hotplug action would not take place i.e. changing the number of sockets will be staged until further reboot.
 Under the `cpu` entry, it will be possible to define a maximum number of sockets. That number is needed by LibVirt and will default to 4 times the initial number of sockets if not set. That default value will be configurable in the KubeVirt CR.
 On VM startup, that value will translate to a new `maxSockets` entry under `spec.domain.cpu` in the VMI object. However, all VMI CPU fields will stay immutable, and CPU sockets will only be adjustable on the VM object.
+Once the hoplug action begins, subsequent changes to the number of sockets would be rejected until the hotplug process completes.
+During the hotplug action a notifitcation will be created upon successful update of the CPU resources on the pod level. It will indicate that now it safe to proceed to the hotplug
+action on the Libvirt/QEMU level. 
 
-Increasing the number of CPU sockets will not only add CPUs to the guest but also increase the CPU resources available to the virt-launcher pod, via a migration (more on that in the Implementation section).
+Increasing the number of CPU sockets will not only add CPUs to the guest but also increase the CPU resources available to the virt-launcher pod, via a migration (more on that in the Implementation section). CPU Resource update on the pod level must occur before the actual hotplug action on the Libvirt level. 
+Therefore, a notifitcation will be created upon successful update of the CPU resources on the pod level. It will indicate that its safe to proceed with updating the libvirt domain with the new topology.
+
 The way LibVirt handles CPU hotplug is by exposing the maximum number of CPUs to the guest and turning off the unused CPUs. In this document, when we refer to "adding" CPUs to the guest, we're really just turning them on. This is why we need to define a maximum number of sockets in advance.
 
 ## Drawbacks / Limitations
 - Since CPU hot-(un)plug involves a live migration, VMs will have to be `LiveMigratable` to enable CPU hotplug
+- Workload update method should be defined explicitly as `LiveMigrate` under the Kubevirt configuration.
 - Each disabled vCPU, so `(maxSockets - sockets) * cores * thread`, consumes 8MiB of overhead memory
 - This feature will be incompatible with CPU requests/limits, at least initially, since those values need to dynamically scale according to the current number of enabled CPUs
 
@@ -113,7 +119,8 @@ Upon succesful hotplug action the condition will be removed. Otherwise, a human 
 We will be able to use the new `liveUpdateFeatures` section to manage things like memory hotplug in the future.
 
 ## Update/Rollback Compatibility
-This is an entirely new feature, so updates from older versions will not be a problem.
+The feautre doesn't intorduce any breaking changes so existing functionality of versions prior to this feature should not be impacted upon update/rollback.
+However if existing workloads would like to opt-in for this feature after an update, the VMI should be re-created since the addition of new fields to the VMI API.
 
 ## Functional Testing Approach
 Functional tests will simply:
@@ -121,6 +128,7 @@ Functional tests will simply:
 - Modify the number of CPU sockets for the VM
 - Ensure the guest sees the changed number of CPUs
 - Ensure the CPU resources on the virt-launcher pod were adjusted accordingly
+- Ensure interoperability with non-hotplug related  live-migration occasions (workload-update, evacuation, user-triggered)
 
 # Implementation Phases
 
@@ -136,10 +144,13 @@ The VMI [workload update controller](https://github.com/kubevirt/kubevirt/blob/c
 Sequence of events:
 - User updates CPU sockets on a vm.spec.template.spec.domain.cpu
 - VM controller determines the request/limits required and writes the necessary changes to the active vmi.Spec
+- VM controller writes the required request/limits in the vmi.Status
 - VMI workload update controller detects the mismatch between the resource request/limits on the pod vs vmi.Spec and live migrates the VMI to satisfy vmi.spec
 - Migration controller constructs a new target pod with the desired CPU request/limits
-- VMI live migrates
-- virt-handler triggers the hotplug/hot-unplug LibVirt action after the live migration completes
+- Migration controller identfies a match between the required request/limits on the vmi.Status and those of the VMI target pod
+- Migration controller marks the VMI as safe to hotplug.
+- VMI live migrates.
+- virt-handler identifies the mark and triggers the hotplug/hot-unplug LibVirt action after the live migration completes
 - VM status is updated to reflect that the pending hotplug/hot-unplug actions have completed
 
 ## Phase 3 (future): Using the [Kubernetes vertical scaling API](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources ) once it reaches a more mature state
@@ -149,10 +160,13 @@ The inplace vertical pod feature allows us to dynamically change the VMI pods re
 Sequence of events
 - User updates CPU sockets on a vm.spec.template.spec.domain.cpu
 - VM controller writes these change to the active vmi.Spec
-- VMI controller updates the active pod spec to reflect the request/limit changes
-- Pod changes are applied
-- virt-handler triggers the hotplug/hot-unplug action after the inplace update completes.
-- VM status is updated to reflect that the pending hotplug/hot-unplug actions have completed
+- VM controller writes the required request/limits in the vmi.Status
+- VMI controller updates the active pod spec to reflect the request/limit changes.
+- Pod changes are applied.
+- VMI controller identfies a match between the required request/limits on the vmi.Status and those of the VMI pod
+- VMI controller marks the VMI as safe to hotplug.
+- virt-handler identifies the mark and triggers the hotplug/hot-unplug action.
+- VM status is updated to reflect that the pending hotplug/hot-unplug actions have completed.
 
 # References
 


design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,146 @@
...
+Under the `cpu` entry, it will be possible to define a maximum number of sockets. That number is needed by LibVirt and will default to 4 times the initial number of sockets if not set.
+
+Increasing the number of CPU sockets will not only add CPUs to the guest but also increase the CPU resources available to the virt-launcher pod, via a migration (more on that in the Implementation section).
+The way LibVirt handles CPU hotplug is by exposing the maximum number of CPUs to the guest and turning off the unused CPUs. In this document, when we refer to "adding" CPUs to the guest, we're really just turning them on. This is why we need to define a maximum number of sockets in advance.
+
+## Drawbacks / Limitations
+- Since CPU hot-(un)plug involves a live migration, VMs will have to be `LiveMigratable` to enable CPU hotplug
+- Each disabled vCPU, so `(maxSockets - sockets) * cores * thread`, consumes 8MiB of overhead memory
+- This feature will be incompatible with CPU requests/limits, at least initially, since those values need to dynamically scale according to the current number of enabled CPUs
```
vladikr, line 39: Won't the target pod be created with the updated topology and hence increase the requested overhead?

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,146 @@
...
+
+```
+// The declared number of sockets the vm is converging on
+vmi.Status.Resources.Sockets.Desired
+vm.Status.Resources.Sockets.Desired
+
+// The number of sockets the vm has currently active.
+vmi.Status.Resources.Sockets.Ready
+vm.Status.Resources.Sockets.Ready
+```
```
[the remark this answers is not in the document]
jean-edouard: I think that makes sense. @davidvossel WDYT?

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,146 @@
...
+Sequence of events:
+- User updates CPU sockets on a vm.spec.template.spec.domain.cpu
+- VM controller determines the request/limits required and writes the necessary changes to the active vmi.Spec
+- VMI workload update controller detects the mismatch between the resource request/limits on the pod vs vmi.Spec and live migrates the VMI to satisfy vmi.spec
+- Migration controller constructs a new target pod with the desired CPU request/limits
+- VMI live migrates
+- virt-handler triggers the hotplug/hot-unplug LibVirt action after the live migration completes
+- VM status is updated to reflect that the pending hotplug/hot-unplug actions have completed
+
+## Phase 3 (future): Using the [KubeVirt vertical scaling API](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources ) once it reaches a more mature state
```
kvaps, line 128:
```suggestion
## Phase 3 (future): Using the [Kubernetes vertical scaling API](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources ) once it reaches a more mature state
```

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,146 @@
...
+kind: VirtualMachine
+metadata:
+  labels:
+    kubevirt.io/vm: vm-cirros
+  name: vm-cirros
+spec:
+  running: false
+  liveUpdateFeatures:
+  - cpu
+      maximumSockets: 4
```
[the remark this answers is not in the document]
kvaps: This approach is not intended to live-update VMI resources directly?
Thus to hotplug CPU you always have to have the above VM resource, am I right?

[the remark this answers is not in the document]
jean-edouard: > Is that list with maps? I think it's better to make map of maps

Right, this has to be a map of maps, will fix.

> However, personally I don't like to have parameters like maximumSockets outside of domain section

Since `maximumSockets` is an attribute of the CPU live update feature, and since it doesn't make sense when the feature is not enabled, I thought having is there made the most sense. We could move it to `domain` and ensure the live update feature is enabled in a webhook, but I'm not sure it adds any value...

> This approach is not intended to live-update VMI resources directly?
> Thus to hotplug CPU you always have to have the above VM resource, am I right?

Live updates features would indeed require VMs and not just VMIs. Technically, final users of KubeVirt should only ever work with VMs, VMIs being just intermediate objects that they should not be concerned with. In reality though, VMIs do appear to be quite popular!
That being said, having the `liveUpdateFeatures` and dynamic `sockets` in boths VMs and VMIs feels like a bit of a headache to reconcile...

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,146 @@
...
+The third phase will happen later as a separate PR. It should be able to transparently replace the migration mechanism without API changes.
+
+## Phase 1: Creating the KubeVirt API and linking it to the LibVirt CPU hotplug API
+## Phase 2: Triggering a live migration when the number of sockets changes, to adjust the virt-launcher CPU resources
+
+The VMI [workload update controller](https://github.com/kubevirt/kubevirt/blob/ce8aab7874d2e1586787e3e2a17306b7edca1b8a/pkg/virt-controller/watch/workload-updater/workload-updater.go#L476) is extended to live migrate VMIs to satisfy updates to a VMI resource request/limits field.
+
+Sequence of events:
+- User updates CPU sockets on a vm.spec.template.spec.domain.cpu
+- VM controller determines the request/limits required and writes the necessary changes to the active vmi.Spec
```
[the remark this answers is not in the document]
jean-edouard: I probably need to rework this sentence. The idea is to disallow user specified CPU requests/limits when the CPU live update feature is enabled, because we do need to dynamically change the CPU request values. If we didn't do that, the virt-launcher pod won't get more CPU resources, and increasing the number of CPU sockets will have little to no effect...

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,164 @@
...
+
+The VMI status must indicate the current number of CPU sockets. The value needs to be propagated back to the VM status to allows the user to have an indication of when their changes applied to the running VMI.
+
+The user should be able to inspect the VM.Status value below to determine the state of their changes. This is similar in concept to someone inspecting a Deployments status to determine the the number of pods that are available.
+
+```
+// The number of sockets the vm has currently active.
+vmi.Status.Resources.Sockets.Ready
+vm.Status.Resources.Sockets.Ready
+```
```
vladikr, line 102: Would it make sense to simply keep the `CurrentCPUTopology` as a whole in the status? I know that only sockets change, but that would show the actual current topology and no one would be confused.
I just had something like that in mind: ```vmiCopy.Status.CurrentCPUTopology = vmi.Spec.Domain.CPU```

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,164 @@
...
+
+Conditions on the VM and VMI should indicate the status of the pending CPU hotplug actions. When errors occur, the conditions should propagate the errors to the VM.Status.Conditions field so users can receive feedback into the status of their declared state.
+
+```
+// VirtualMachineCPUReady indicates that the desired amount of cpu is active on the vmi
+VirtualMachineCPUReady VirtualMachineConditionType = "CPUReady"
+```
+
+When the desired state is equal to the ready state, then these conditions will have `Status=True`, otherwise `Status=False` with a human readable reason and message indicating whether the hotplug change is pending or encountering errors.
```
vladikr, line 111: I would expect the condition to be present during the whole period of the hot plug/unplug and would it would be deleted when the operation is done.

This would also serve as an indication to the controller that the CPU change operation is ongoing (and there is no need to create a new migration). As in https://github.com/kubevirt/kubevirt/pull/9401/files#diff-05d3e75077f11cdd56ce409e1e87c9abcdd09fe103540cb25f920d0db319cdd0R212

What do you think?

enp0s3:
/assign @acardace 

Hi. I've updated the status API, can you PTAL?

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,163 @@
...
+    cpu:
+      maxSockets: 4
+      sockets: 2
+      cores: 4
+      threads: 2
+```
+
+### Status API
+
+The VMI status must indicate whether CPU hotplug is taking place. As the hotplug process begins the current state of `spec.domain.cpu` needs to be propagated back to the VM/VMI `status.currentCPUTopology` to allow the user to have an indication of the hotplug processing.
```
acardace, line 94: What about `status.pendingLiveUpdates.currentCPUTopology` ? I think it would be better to embed this in a `pendingLiveUpdates` struct so that we can plug other hotplug features status info in there. I'm mainly thinking about memory hotplug too.

vladikr: @acardace , this is actually the current topology as the guest currently sees it, the `desired` state is reflected in the spec. The same pattern is used by ResourceQuota, for example - when the spec is changed the actual state of the quota is listed in the status and only changes then the controller actually start using the changes requested in the spec.

enp0s3: @acardace Hi! wrapping these statuses under common subject would introduce a technical debt, since after successful hotplug action we are removing the `currentCPUTopology`. Now assume we have it wrapped, we will have to take care about removing the wrapper as well. Perhaps its better to leave it flat to ease the maintenance, WDYT?

acardace: You guys are right, let's keep it like this.

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,163 @@
...
+As a VM user, I would like to dynamically increase or decrease the number of CPUs used by a VM without requiring the VM to restart.
+
+## Repos
+kubevirt/kubevirt
+
+# Design
+In VM objects, the existing CPU `sockets` field will now be dynamic, provided that it has been declared as such.
+Declaring the field as dynamic will be done by adding a `cpu` entry to a new section under the VM spec, called `liveUpdateFeatures`.
+Under the `cpu` entry, it will be possible to define a maximum number of sockets. That number is needed by LibVirt and will default to 4 times the initial number of sockets if not set. That default value will be configurable in the KubeVirt CR.
+On VM startup, that value will translate to a new `maxSockets` entry under `spec.domain.cpu` in the VMI object. However, all VMI CPU fields will stay immutable, and CPU sockets will only be adjustable on the VM object.
```
enp0s3, line 32: @vladikr @jean-edouard Hi. I have a concern about setting a default for `spec.liveUpdateFeatures.CPU` because assuming that memory hotplug will live here too, and the user wants to configure memory hotplug only, then we will set defaults for CPU, and then the VM update admitter will prevent the user from configuring the CPU resources, although the user wanted only memory hotplug. WDYT?

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,163 @@
...
+
+## User Stories
+As a VM user, I would like to dynamically increase or decrease the number of CPUs used by a VM without requiring the VM to restart.
+
+## Repos
+kubevirt/kubevirt
+
+# Design
+In VM objects, the existing CPU `sockets` field will now be dynamic, provided that it has been declared as such.
+Declaring the field as dynamic will be done by adding a `cpu` entry to a new section under the VM spec, called `liveUpdateFeatures`.
```
EdDev, line 30: For the network interface hotplug, I suggested to use the existing domain resources requests (or/and limits) with a dedicated resource name. The semantics of "requests" is for assuring reservation and "limits" for placing an upper limit.

Is there a reason not to re-use the "map" of resources for the same?

EdDev (commented):
Hotplug seems to be a "hot" topic these days and several SIG/s are working on it at different domains.

It may be useful to align and sync the suggested API/s for the existing resources with a look to the future as well.

design-proposals/cpu-hotplug.md
jean-edouard: `liveUpdateFeatures` is, at its core, the list of features that can be live-updated. If it's empty, then CPU/mem/other will **not** be live-updateable.
In the case of `cpu: {}`, CPUs are now live-updatable, and maxSockets gets set to its default value.
That's how I see it at least.

vladikr: I'd prefer each feature to be explicitly listed. The list of supported features can change, but the VM spec may remain the same...

design-proposals/cpu-hotplug.md
vladikr: I see the map of `liveUpdateFeatures` to be inconsistent with ` resources requests`, in the future `liveUpdateFeatures` may contain not just resources thinks like nodeSelector, etc...

EdDev: Well, I do not understand the grouping then.
Subject domains are distributed in the spec, but this one seems to try and group "live-update" and de facto duplicates other subject domains trees.

The parts which are about reservation or limitation of resources, can re-use the `Domain.Resources` structure. When the scope changes and requires a more complex coordination, new things could be created.

Are the existing resources discussed in this proposal not a fit for the domain resource semantics?
Is there an example of something that is not a fit? I want to understand better the `resources` limitation in this regard.

vladikr: Normally, changes made to the VMI template will not take effect immediately, but only on a creation of a new VMI.
The new `vm.spec.liveUpdateFeatures` should tell which changes to the vmi template will propagate immediately.
Having this I'm mind, I don't see a clear connection to the resources struct.   

> Are the existing resources discussed in this proposal not a fit for the domain resource semantics?
> Is there an example of something that is not a fit? I want to understand better the resources limitation in this regard.

This is not part of this proposal, however, it was discussed to add something like `Affinity` will be added in the future to dynamically control the node affinity of the VMI. 
Additionally, the Memory struct which will be added in the future, could preallocate the hot-plugged memory, so it could have something like a `MaxQuantity` field in it.

I don't see a reason why we need to force the liveUpdateFeatures to use the resource struct.

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,163 @@
...
+Under the `cpu` entry, it will be possible to define a maximum number of sockets. That number is needed by LibVirt and will default to 4 times the initial number of sockets if not set. That default value will be configurable in the KubeVirt CR.
+On VM startup, that value will translate to a new `maxSockets` entry under `spec.domain.cpu` in the VMI object. However, all VMI CPU fields will stay immutable, and CPU sockets will only be adjustable on the VM object.
+
+Increasing the number of CPU sockets will not only add CPUs to the guest but also increase the CPU resources available to the virt-launcher pod, via a migration (more on that in the Implementation section).
```
xpivarc, line 34: Do we want to mention that the sockets will be identical? Meaning you can hot plug more cpus than one could think?

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,163 @@
...
+## User Stories
+As a VM user, I would like to dynamically increase or decrease the number of CPUs used by a VM without requiring the VM to restart.
+
+## Repos
+kubevirt/kubevirt
+
+# Design
+In VM objects, the existing CPU `sockets` field will now be dynamic, provided that it has been declared as such.
+Declaring the field as dynamic will be done by adding a `cpu` entry to a new section under the VM spec, called `liveUpdateFeatures`.
+Under the `cpu` entry, it will be possible to define a maximum number of sockets. That number is needed by LibVirt and will default to 4 times the initial number of sockets if not set. That default value will be configurable in the KubeVirt CR.
```
xpivarc, line 31: Is this field mutable?

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,163 @@
...
+```
+	// CPU allow specified the detailed CPU topology inside the vmi.
+		CurrentCPUTopology *CPU `json:"currentCPUTopology,omitempty"`
+```
+
+Conditions on the VM and VMI should indicate the status of the pending CPU hotplug actions. When errors occur, the conditions should propagate the errors to the VM.Status.Conditions field so users can receive feedback into the status of their declared state.
+
+```
+	// Indicates that the VMI is in progress of Hot vCPU Plug/UnPlug
+	VirtualMachineInstanceVCPUChange VirtualMachineInstanceConditionType = "HotVCPUChange"
```
xpivarc, line 107: I would suggest to be more specific, referring to plug or unplug.

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+Upon succesful hotplug action the condition will be removed. Otherwise, a human readable reason and message indicating whether the hotplug change is pending or encountering errors.
+
+## Extensibility
+We will be able to use the new `liveUpdateFeatures` section to manage things like memory hotplug in the future.
+
+## Update/Rollback Compatibility
+This is an entirely new feature, so updates from older versions will not be a problem.
```
xpivarc, line 116: What about VMs created in x-1 version? Are they able to leverage this feature?

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Functional Testing Approach
```
xpivarc, line 118: Are there plans to test following:
1. Unplug
2. Hot plug/unplug during hot plug/unplug
3. Interaction with max inflight migrations?
?

design-proposals/cpu-hotplug.md
```diff
@@ -0,0 +1,163 @@
...
+
+The VMI [workload update controller](https://github.com/kubevirt/kubevirt/blob/ce8aab7874d2e1586787e3e2a17306b7edca1b8a/pkg/virt-controller/watch/workload-updater/workload-updater.go#L476) is extended to live migrate VMIs to satisfy updates to a VMI resource request/limits field.
+
+Sequence of events:
+- User updates CPU sockets on a vm.spec.template.spec.domain.cpu
+- VM controller determines the request/limits required and writes the necessary changes to the active vmi.Spec
+- VMI workload update controller detects the mismatch between the resource request/limits on the pod vs vmi.Spec and live migrates the VMI to satisfy vmi.spec
+- Migration controller constructs a new target pod with the desired CPU request/limits
+- VMI live migrates
+- virt-handler triggers the hotplug/hot-unplug LibVirt action after the live migration completes
```
xpivarc, line 142: I completely  miss an information which will be used by handler to determine that the migration happened.

xpivarc (commented):
I think this is converging but I still miss few things to call this complete.

One thing which I am missing is a description of behavior of what does happen if I change the the number of sockets while a hot plug/unplug is in progress. More over it would be good to describe what and when needs to be validated. Last but not least the design mostly speaks about "happy" path but what does happen if none of the nodes have enough cpus?

design-proposals/cpu-hotplug.md
vladikr: Do you mean that the total number of plugged vCPUs will be (number of changed sockets) * cores * threads?
I'm not sure I understand the question...

design-proposals/cpu-hotplug.md
vladikr: Everything in kubevirt CR is mutable, no?

design-proposals/cpu-hotplug.md
vladikr: Why would it matter? the main point is that the number of vCPUs in changing. Whether this is plug or unplug one could understand by looking at the vmi spec (desired state) and the currentCPUTopology (current state)

design-proposals/cpu-hotplug.md
vladikr: Not without a re-creation of the VMI

design-proposals/cpu-hotplug.md
vladikr: Hmm.. this should take place in finalizeMigration() - this assumes that the migration is over.
Here virt-handler would check if vmi.Status.CurrentCPUTopology exists and will call `SyncVirtualMachineCPUs`

vladikr: but this seems like an implementation detail to me.

design-proposals/cpu-hotplug.md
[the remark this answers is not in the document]
xpivarc: Yes, that is exactly what I mean. I just find it better to be written down.

design-proposals/cpu-hotplug.md
xpivarc: Yes, is my understanding correct?

design-proposals/cpu-hotplug.md
xpivarc: I thought this refers mainly to `VM.spec.liveUpdateFeatures.cpu`. KV is of course mutable and I don't see a problem there.

design-proposals/cpu-hotplug.md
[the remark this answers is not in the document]
vladikr: @enp0s3 could you please add this line?

design-proposals/cpu-hotplug.md
vladikr: Right, changing the number of maxCpus in the `liveUpdateFeatures.cpu` struct shouldn't have any effect on the running VMI. It should only propagate to the vmi spec on re-creation.

design-proposals/cpu-hotplug.md
xpivarc: Fair enough

design-proposals/cpu-hotplug.md
xpivarc: I would write it down

design-proposals/cpu-hotplug.md
vladikr: Yes. That's correct. I guess that's what topology means. I'd prefer this to be explained in detail in the user guide.
Perhaps you can suggest a better sentence here?

design-proposals/cpu-hotplug.md
xpivarc: I agree that it is implementation detail but at the same time one of which matters imho. Why? Well as you said in this case you would like to implement it in `finalizeMigration` but what about the version where we use in-place update? 

Also `SyncVirtualMachineCPUs` seems like new API and it is not documented. Why do we need it and why can't we do this in `SyncVMI` (existing) call?

vladikr: > what about the version where we use in-place update?

such a version doesn't exist today. However, `SyncVirtualMachineCPUs` was introduced specifically to address the vCPU change which can happen with or without migration.
I honestly don't see how we can mix this functionality with a regular syncVMI flow. 
`SyncVirtualMachineCPUs` should happen at a specific time when we know that vCPUs can be manipulated. virt-handler should orchestrate this step and it shouldn't happen on ever syncVMI call.

design-proposals/cpu-hotplug.md
EdDev: > Normally, changes made to the VMI template will not take effect immediately, but only on a creation of a new VMI. The new `vm.spec.liveUpdateFeatures` should tell which changes to the vmi template will propagate immediately. Having this in mind, I don't see a clear connection to the resources struct.

Indeed, changes to the VM spec template are currently never passed on to the underlying VMI.
Storage hotplug have and not network hotplug are using a command queue to propagate such changes.

Is this proposal is suggesting a new way to propagate live changes?

> I don't see why we need to force the liveUpdateFeatures to use the resource struct.

That was not my intention.
The way storage & network hotplug are propagated from a VM to a VMI is through a queue on the VM status.
Users can only use the virt-api to update live changes, not directly on the VM. I.e. changes on the VM directly are never "live".

The usage of the resource generic structure is to pass in various key-value that represent resources of the domain, relevant for both immutable and mutable resources. The ones that are live updatable, are processed by the VMI, the others are left for the next boot. If there are existing fields for the resource in other places (e.g. disks, volumes, interfaces, cpu count), no need to add to the resource key/value anything.

> > Are the existing resources discussed in this proposal not a fit for the domain resource semantics?
> > Is there an example of something that is not a fit? I want to understand better the resources limitation in this regard.
> 
> This is not part of this proposal, however, it was discussed to add something like `Affinity`, which will be added in the future to control the node affinity of the VMI dynamically. Additionally, the Memory struct which will be added in the future, could preallocate the hot-plugged memory, so it could have something like a `MaxQuantity` field in it.
> 
> I don't see why we need to force the liveUpdateFeatures to use the resource struct.

design-proposals/cpu-hotplug.md
xpivarc: Maybe it is enough to add "Today's approach, hotplugging sockets, semantically means that more than one vCPU can be added, by adding one socket, depending on the topology. This can be expressed by formula (number of changed sockets) * cores * threads "

design-proposals/cpu-hotplug.md
acardace: I honestly like this proposal as is, the introduction of the `liveUpdateFeatures` field makes it straightforward to understand which resources are live updatable in a running VM.

To me, it just makes it easier to better understand and use the hotplug capabilities of the VM.

design-proposals/cpu-hotplug.md
xpivarc: >such a version doesn't exist today. However, SyncVirtualMachineCPUs was introduced specifically to address the vCPU change which can happen with or without migration.

From goals section "Underlying technology used for hotplug should be independent of the VM API, meaning we should be able to swap out how hotplug is technically performed in the future without impacting API compatibility"

I am trying to see if current approach/API is sufficient for seamless change of way we do this.

Therefore let me re-iterate once again. How do we signal the hotplug in case of different way of implementing this?

While writing this I also don't understand what is an expecting behavior in this case:
1. VM is being migrated or is going to be migrated
2. Users changes the number of sockets
3. Migration is completed

What happens in step 3? Will we explode?

xpivarc: @enp0s3 Consistency and it should be rather other way around. What is benefit of new API and not keeping it consistent?

xpivarc: @acardace 

>I agree with @vladikr on this, we need a new API to make this easier to work with when the vertical scaling API will land.

I am probably missing something. How is this making it easier? Also please document the API requirement here.

>SyncVMI is not a good place to put hotplug logic otherwise you will see continuous notification of a device being hotplugged in the guest, it's definitely better to hotplug when requested with a different code path.

Why would we have continuous notifications? Would simple `if` be sufficient?
`if hasDifferentCPU {
    do the magic and log, otherwise no notification
}`

>it's definitely better to hotplug when requested with a different code path.

What are the benefits?

acardace: > I am probably missing something. How is this making it easier? Also please document the API requirement here.

I mean internal API, between libvirt and kubevirt.

Requirement is it should hotplug CPUs :)

> 
> > SyncVMI is not a good place to put hotplug logic otherwise you will see continuous notification of a device being hotplugged in the guest, it's definitely better to hotplug when requested with a different code path.
> 
> Why would we have continuous notifications? Would simple `if` be sufficient? `if hasDifferentCPU { do the magic and log, otherwise no notification }`

Yes, of course, you can do it by adding fields in the vmi.Status or check what the current domain looks like, to check whether you already hotplugged CPUs or whether a hotplug is in progress but SyncVMI() creates a new api.Domain anew every time and I'd rather leave its functionality and semantics as is.

I think it's simpler and more maintainable/readable if we use another internal API to do live hotplug which is a one-time live operation.

> 
> > it's definitely better to hotplug when requested with a different code path.
> 
> What are the benefits?

For me simplicity.

acardace: > > such a version doesn't exist today. However, SyncVirtualMachineCPUs was introduced specifically to address the vCPU change which can happen with or without migration.
> 
> From goals section "Underlying technology used for hotplug should be independent of the VM API, meaning we should be able to swap out how hotplug is technically performed in the future without impacting API compatibility"
> 
> I am trying to see if current approach/API is sufficient for seamless change of way we do this.
> 
> Therefore let me re-iterate once again. How do we signal the hotplug in case of different way of implementing this?
> 
> While writing this I also don't understand what is an expecting behavior in this case:
> 
> 1. VM is being migrated or is going to be migrated
> 2. Users changes the number of sockets
> 3. Migration is completed
> 
> What happens in step 3? Will we explode?

After the VM has been migrated virt-controller will notice that the CPUTopology in the vmi.Status is different than the desired one and will hotplug I guess?

acardace: > > I mean internal API, between libvirt and kubevirt.
> 
> I don't understand the answer. We never strived to have 1:1 mapping to libvirt nor do I think it is easier.

I didn't say that.

> 
> > Yes, of course, you can do it by adding fields in the vmi.Status or check what the current domain looks like, to check whether you already hotplugged CPUs or whether a hotplug is in progress but SyncVMI() creates a new api.Domain anew every time and I'd rather leave its functionality and semantics as is.
> 
> I believe the semantics doesn't change.

I don't agree.

> 
> > I think it's simpler and more maintainable/readable if we use another internal API to do live hotplug which is a one-time live operation.
> 
> My experience is contradicting this. The less we expand this internal API the better as we reuse what we already have. Also now we are leaking logic which is responsible for what should happen to virt-handler. My perception is that we mostly kept this in virt-launcher.

I don't know maybe you're right about all of this, it's also hard to discuss this back and forth in GH like this, what's most important is that I think we're diverging from the objective, this is a design PR, not an implementation one.

What I mean is that I don't really mind where it lands (SyncVMI vs separate API) as long as the implementation is sound, but we should focus on the overall design here which I think it's outlined pretty clearly.

> 
> > After the VM has been migrated virt-controller will notice that the CPUTopology in the vmi.Status is different than the desired one and will hotplug I guess?
> 
> Sure but migrated Pod will have the same resources, meaning not adjusted to what was asked, and still I believe you want to trigger the hot un/plug. Leaving the domain updated but Pod is not. Or even worse making race behavior.

Then it's going to migrate again, but I'd stop here since I feel like we're going into a rabbit hole and these are implementation details anyway, the desired outcome is clear, hotplug has to happen if requested.

vladikr: > > I mean internal API, between libvirt and kubevirt.
> 
> I don't understand the answer. We never strived to have 1:1 mapping to libvirt nor do I think it is easier.
> 
> > Yes, of course, you can do it by adding fields in the vmi.Status or check what the current domain looks like, to check whether you already hotplugged CPUs or whether a hotplug is in progress but SyncVMI() creates a new api.Domain anew every time and I'd rather leave its functionality and semantics as is.
> 
> I believe the semantics doesn't change.
> 
> > I think it's simpler and more maintainable/readable if we use another internal API to do live hotplug which is a one-time live operation.
> 
> My experience is contradicting this. The less we expand this internal API the better as we reuse what we already have. Also now we are leaking logic which is responsible for what should happen to virt-handler. My perception is that we mostly kept this in virt-launcher.
> 
> > After the VM has been migrated virt-controller will notice that the CPUTopology in the vmi.Status is different than the desired one and will hotplug I guess?
> 
> Sure but migrated Pod will have the same resources, meaning not adjusted to what was asked, and still I believe you want to trigger the hot un/plug. Leaving the domain updated but Pod is not. Or even worse making race behavior.

@xpivarc It will not. The target pod will have adjusted resources. 
When we create the target pod, it is being created from a template that uses the updated the vmi spec, thus it will correctly calculate and update the requested resources.

xpivarc:
@jean-edouard @enp0s3 
We have discussed this out-of-band with @vladikr & @acardace. We are missing one important high-level design point. The point is signaling. How is virt-handler or virt-launcher going to recognize that the Pod env was updated and it is safe to hot-un/plug resources? 
I think this is the last part that needs to be described before we move forward.

The above should be implementation independent. Meaning it should work with migration base and in-place pod update implementation as well without the need to modify API.

Also, we need to make sure that once we make a signal the VMI will not change the resources.

design-proposals/cpu-hotplug.md
EdDev: This suggestion is not straightforward to me, because it introduces two places for the same information to exist. At least, this is not a simple solution, considering that users just want to request changes to a running VM.
But the real interesting feedback is from the feature stakeholders, which should confirm if this fits their needs or not.

But my main concern is about not having a common consistent method to do live updates.
IMO this needs to be discussed in a cross SIG forum and once agreed upon on the API, to convert all to the same method.

ormergi: >I honestly like this proposal as is, the introduction of the liveUpdateFeatures field makes it straightforward to understand which resources are live updatable in a running VM.

It seems that it does more than that, it enables setting an upper limit for CPU sockets inside the guest.
Per my understanding this can be achieved through `resource.limits` or `spec.domain.cpu`.

As for today you have two places to set the CPU resource, adding another place make the API more cumbersome and worse experience IMO.

design-proposals/cpu-hotplug.md
dankenigsberg:
```suggestion
- This feature makes CPU requests/limits second-class entries, as they would be automatically recalculated by KubeVirt based on the number of sockets.
```
+1 for a rephrase ^^^

enp0s3:
@acardace @xpivarc @vladikr Hi. I've updated the doc, PTAL. Thanks!

dankenigsberg (changes requested):

fabiand:
I'm not able to respond to @dankenigsberg outdated comment.
If we agree that the current cookie-cut approach is different to live-updates, then we need a way to control this.
Current users of VMs can rely on the functionality that updates to sockets are not permitted.
Thus enabling this feature by default would be a surprise to them.

EdDev:
> I'm not able to respond to @dankenigsberg outdated comment. If we agree that the current cookie-cut approach is different to live-updates, then we need a way to control this. Current users of VMs can rely on the functionality that updates to sockets are not permitted. Thus enabling this feature by default would be a surprise to them.

You can comment in its thread [here](https://github.com/kubevirt/community/pull/216#discussion_r1204313159).
This is GitHub trying to provide two things that do not play well together: Show new comments coming in and keeping old threads alive.

design-proposals/cpu-hotplug.md
enp0s3: @dankenigsberg You stated that as a user you want the CPU hotplug to happen automatically, w/o additional API fields like `liveUpdateFeatures`, but how will it fit with this drawback? and if the suggestion is to gate CPU hotplug in KV CR, what about multi-tenant envs?

dankenigsberg: Hmm. I did not think about this issue. I cannot imagine a user changing the number of cores but wanting it to take effect only in the far future. However I can image someone being surprised by by the request/limits suddenly recalculated.

I think that the way to solve this is to have KubeVirt recalculate the requests/limits of the virt-launcher pod and only if requests/limits were not specified. The VM spec is the input to KubeVirt, controlled by humans. It would be a violation of layering if KubeVirt is to edit its own spec.

I don't think that it is important to block hot-plug on the cluster level and certainly not on the tenant level. VM owners can decide to use it only when they need it. I think that `liveUpdateFeatures` introduces complexity that is not really needed by users; it is an expensive commitment that clutters the API and forces us to add many tests. If we believe that we must cater to users who do not want hot-plug, a cluster-wide feature gate would be less expensive.

dankenigsberg:
> I'm not able to respond to @dankenigsberg outdated comment. If we agree that the current cookie-cut approach is different to live-updates, then we need a way to control this. Current users of VMs can rely on the functionality that updates to sockets are not permitted. Thus enabling this feature by default would be a surprise to them.

I find the costs (in API complexity, testing) much higher than the chance that someone relies on the current behavior. If we must cater for such users, a cheaper option is a cluster-wide feature gate.

vladikr:
> > I'm not able to respond to @dankenigsberg outdated comment. If we agree that the current cookie-cut approach is different to live-updates, then we need a way to control this. Current users of VMs can rely on the functionality that updates to sockets are not permitted. Thus enabling this feature by default would be a surprise to them.
> 
> I find the costs (in API complexity, testing) much higher than the chance that someone relies on the current behavior. If we must cater for such users, a cheaper option is a cluster-wide feature gate.

This proposal lays a foundation for other future "live-updateable" features such as memory hotplug, live changing node affinity, etc
We won't be able to get away with a feature gate for each of these additional features. 
For example, for a memory hotplug (depending on the driver) we will need to specify the amount of memory for the hot plug. This would need to be immutable.
It may also be tricky for `Affinity` changes.

I think that it's nice to gather all of these under one struct:
```
 liveUpdateFeatures:
      - CPU: 
             MaximumSockets: 8 (optional)
      - Memory:
             MaximumQuantity: 10Gi (optional)
      - Affinity: {}
```

* Network and Storage can also be added in the future

We can of course manage without the `liveUpdateFeatures` struct, but then we will lose both the predictability, 
should the changes be staged or live propagated and structure as in some cases we would still need to set some fields (e.g. `MaxSockets` / `MaxQuantity` / Maybe other filed)  on the immutable part of the VM spec:
```
kind: VirtualMachine
...
spec:                                                                                                                                            
  MaximumSockets: 8
  MaximumQuantity: 10Gi
  Template:
    ...
```

dankenigsberg:
> > I find the costs (in API complexity, testing) much higher than the chance that someone relies on the current behavior. If we must cater for such users, a cheaper option is a cluster-wide feature gate.
> 
> This proposal lays a foundation for other future "live-updateable" features such as memory hotplug, live changing node affinity, etc We won't be able to get away with a feature gate for each of these additional features. For example, for a memory hotplug (depending on the driver) we will need to specify the amount of memory for the hot plug. This would need to be immutable. It may also be tricky for `Affinity` changes.
> 
> I think that it's nice to gather all of these under one struct:
> 
> ```
>  liveUpdateFeatures:
>       - CPU: 
>              MaximumSockets: 8 (optional)
>       - Memory:
>              MaximumQuantity: 10Gi (optional)
>       - Affinity: {}
> ```
> 
> * Network and Storage can also be added in the future
> 
> We can of course manage without the `liveUpdateFeatures` struct, but then we will lose both the predictability, should the changes be staged or live propagated

I understand the motivation. I just don't believe that it is worth the effort. Who are these people that want to modify the VM resource now, but *want* it to not to take effect until the next stop/start cycle? Cannot they just wait for the VM to stop and only then modify it?

> and structure as in some cases we would still need to set some fields (e.g. `MaxSockets` / `MaxQuantity` / Maybe other filed) 

Here I believe that we are putting too much burden on the users and too early. I would prefer KubeVirt to have an opinion about the reasonable max values. I don't think that a typical user has the time and knowledge to make an informed decision on this. If I am mistaken and somebody asks for this knob, it can be added in the future, but until then we keep the API simpler.


> on the immutable part of the VM spec:
> 
> ```
> kind: VirtualMachine
> ...
> spec:                                                                                                                                            
>   MaximumSockets: 8
>   MaximumQuantity: 10Gi
>   Template:
>     ...
> ```

vladikr:
> > > I find the costs (in API complexity, testing) much higher than the chance that someone relies on the current behavior. If we must cater for such users, a cheaper option is a cluster-wide feature gate.
> > 
> > 
> > This proposal lays a foundation for other future "live-updateable" features such as memory hotplug, live changing node affinity, etc We won't be able to get away with a feature gate for each of these additional features. For example, for a memory hotplug (depending on the driver) we will need to specify the amount of memory for the hot plug. This would need to be immutable. It may also be tricky for `Affinity` changes.
> > I think that it's nice to gather all of these under one struct:
> > ```
> >  liveUpdateFeatures:
> >       - CPU: 
> >              MaximumSockets: 8 (optional)
> >       - Memory:
> >              MaximumQuantity: 10Gi (optional)
> >       - Affinity: {}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * Network and Storage can also be added in the future
> > 
> > We can of course manage without the `liveUpdateFeatures` struct, but then we will lose both the predictability, should the changes be staged or live propagated
> 
> I understand the motivation. I just don't believe that it is worth the effort. Who are these people that want to modify the VM resource now, but _want_ it to not to take effect until the next stop/start cycle? Cannot they just wait for the VM to stop and only then modify it?
> 
> > and structure as in some cases we would still need to set some fields (e.g. `MaxSockets` / `MaxQuantity` / Maybe other filed)
> 
> Here I believe that we are putting too much burden on the users and too early. I would prefer KubeVirt to have an opinion about the reasonable max values. I don't think that a typical user has the time and knowledge to make an informed decision on this. If I am mistaken and somebody asks for this knob, it can be added in the future, but until then we keep the API simpler.

Do we really want *all* VMs to be hot-pluggable by default, without a way to tune this per VM?

At least, Regarding CPU, guest OS users would see a weird topology with some CPUs being offline. 
For the Memory hot plug, afaiu, we don't need an upper memory limit with virtio-mem, but we do need it for the balloon driver...  


> 
> > on the immutable part of the VM spec:
> > ```
> > kind: VirtualMachine
> > ...
> > spec:                                                                                                                                            
> >   MaximumSockets: 8
> >   MaximumQuantity: 10Gi
> >   Template:
> >     ...
> > ```

acardace:
> At least, Regarding CPU, guest OS users would see a weird topology with some CPUs being offline. For the Memory hot plug, afaiu, we don't need an upper memory limit with virtio-mem, but we do need it for the balloon driver...

@vladikr by the way also memory hotplug with virtio-mem needs a `MaxMemory` field so the design is fitting for that also.

dankenigsberg:
answering several comments at once:

> Do we really want /all/ VMs to be hot-pluggable by default, without a way to tune this per VM?

I think we do. Or more precisely, I think most users would like KubeVirt to make the call (e.g. if the user asked for a machineType that does not support it. I don't know if the example is realistic). Users need hot-plug only very late in the game, when they have an important workload (which cannot be stopped), eating all their CPU/memory. With the merged proposal, we're telling them - sure, you could hot-plug, but only if you had thought about it in the first place. It defies the purpose.

> by the way also memory hotplug with virtio-mem needs a MaxMemory field so the design is fitting for that also.
Most users would not want a new knob for MaxMemory. They have no way to predict it, they don't know the costs of any particular value. They'd prefer KubeVirt to take the call and set it to, say, twice the start-up memory in the guest.

> I'm not sure it's valuable to argue about "who" the user is who needs this behavior. If we change this behavior without a clear way for the user to express intent, then we've broken api behavior

It is very important to imagine whose behavior is being broken and how important it is to them. Otherwise we would never fix any bug. Hyrum is correct, but the conclusion of his law is *not* "never change a line of code without exposing a new knob to the user". It is "be cognizant of your users, think of realistic repercussions of your changes, take reasonable risks and be prepared to make somebody on the Internet unhappy for a while". Or in another words, https://xkcd.com/1172/

PR #218, opened 2023-04-27
vHostUser interface Design Proposal

nvinnakota10:
This is a design proposal for the vhostuser interface to support UserSpace networking feature by the kubevirt VM.

The document provides information on the Feature/Interface proposed, why it is needed, how it can enable fast packet processing when dpdk is available.


design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,78 @@
+
+# Overview
+- User Space networking is a feature that involves packet path from the VM to the data plane running in userspace like OVS-DPDK etc, by-passing the kernel.
+- With VhostUser interface, the kubevirt VM created can utilize the dpdk virtio functionality and avoid using kernel path for the traffic when DPDK is involved, allowing usage of fast datapath and improve performance.
```
alicefr, line 4: I understand that here the main focus is dpdk, but it will be useful to see if it can apply also to other vhost user backend

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,78 @@
...
+## Non Goals
+- The user space networking will require the host to a DPDK dataplane.
+- The kubevirtci creates kernel mode dataplane hosts for testing the VMs. It needs to be updated to support DPDK.
+
+## Definition of Users
+- Any user who has a (k8s) cluster with a dpdk dataplane can use the interface/feature.
+- The feature will be supported only if a valid CNI is being used.
+
+## User Stories
+- As a user/admin, I want to use user space networking feature if I have DPDK supported VM and the host has a DPDK dataplane.
```
alicefr, line 25: Where will be dpdk dataplane running and who will manage it?

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,78 @@
...
+- The design involves introducing vHostUser interface type and required parameters to support it. 
+
+- Creating the VMs will follow the same process with a few changes highlighted as below:
+    1. A client posts a new VM definition to the K8s API Server.  
+    2. The K8s API Server validates the input and creates a VM custom resource definition (CRD) object.  
+    3. The virt-controller observes the creation of the new VM object and creates a corresponding virt-launcher pod.
+    4. **The virt-controller will add two new volumes to the virt-launcher pod.**\
+        a. **EmptyDir volume named (shared-dir) is used to share the vhostuser socket file with the virt-launcher pod and dpdk dataplane.**\
+        b. **DownwardAPI volume named (podinfo) is used to share vhostuser socket file name with the virt-launcher via pod annotations.**
+    5. Scheduling of the launcher pod will be handled by kubernetes.
```
alicefr, line 44: This isn't clear to me. What are the criteria for scheduling a VM with dpdk? Will you use a device plugin and expose the resource? These details should be clarify if you post the API changes

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,78 @@
...
+    6. **The CNI will mount the shared-dir i.e., EmptyDir volume with the mount-path as "/var/lib/cni/usrcni/".**
+    7. **The CNI will update the virt-launcher pod annotations with vhostsocket-file name and details.**
+    8. The virt-controller observes the creation of the virt-launcher pod for the VM and updates the nodeName field in the VM object with the nodeName retrieved from the virt-launcher pod.
```
alicefr, line 47: I don't think it is necessary to list all the steps that don't diverge from the regular flow. For example , the 8 step isn't very clear to me

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,78 @@
...
+    9. The virt-handler observes that a VM got assigned to host where it is running on.
+   10. The virt-handler will provide the VM Specification and signals the virt-launcher via IPC(unix domain socket) to create the VM in the virt-launcher pod.
+   11. The virt-launcher using a libvirtd instance creates the domain.
+   12. **The virt-launcher reads the DownwardAPI volume and retrieves the vhostsocket-file name specified in pod annotations and uses it to create a unix socket in the VM, while launching the VM using libvirtd.**
+   13. **The virt-launcher is modified to skip establishing the networking between VM and the virt-launcher pod using bridge(Refer in Kubevirt Networking section in https://kubevirt.io/2018/KubeVirt-Network-Deep-Dive.html).**
```
alicefr, line 52: How would you do that? Do we need a new network mode?

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,78 @@
...
+   14. **Instead of using the bridge through the launcher pod, the vHostUser interface of the VM will be directly connected to the DPDK datplane using the vhost socket**
+
+- With the above process a kubevirt VM with a vHostUser interface can be acheived. All the steps highlighted will be the changes made as part of the design.
+
+## Scalability
+- There should not be any scalability issues, as the feature is to create an interface when useful.
```
alicefr, line 58: What about migration with a VM using dpdk?

design-proposals/vhostuser-interface.md
nvinnakota10: Removing the steps that will be common for creating a VM and does not change for the VM with vHostUser interface to avoid confusion.

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,78 @@
...
+
+- kubevirt/kubevirt
+   - Introduction of vhostuser type interface.
+- kubevirt/kubevirtci
+   - Enable usage of dpdk on the setups created by kubevirtci.
+
+# Design
+- The design involves introducing vHostUser interface type and required parameters to support it. 
+
+- Creating the VMs will follow the same process with a few changes highlighted as below:
```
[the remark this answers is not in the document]
nvinnakota10: Added API examples for reference.

design-proposals/vhostuser-interface.md
nvinnakota10: The dpdk data plane runs in UserSpace. The user can configure OVS with dpdk or use an appropriate multus CNI to manage it.

design-proposals/vhostuser-interface.md
nvinnakota10: The new interface will be using the vhost socket created by the shared-dir mount. This will allow the VM with a dpdk interface to connect to the DPDK data plane directly bypassing the linux.

design-proposals/vhostuser-interface.md
nvinnakota10: The changes made are generic, so it can also be used by VPP (vector packet processing) or any other host user backend.

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,96 @@
...
+          - name: vhost-user-vn-blue
+            vhostuser: {}
+          useVirtioTransitional: true
+      networks:
+      - name: vhost-user-vn-blue
+        multus:
+          networkName: vn-blue
+
+```
+
```
alicefr, line 68: nit: extra lines not needed

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,96 @@
...
+# Design
+- The design involves introducing vHostUser interface type and required parameters to support it. An EmptyDIR volume (shared-dir) and DownwardAPI (podinfo) volume are mounted on the virtlauncher pod to have the support for new interface.
+- These mounts will allow the virt-launcher pod to create a VM with an additional interface which can be reached using the vhostSocket mentioned in the DownwardAPI.
+
+- Creating the VMs will follow the same process with a few changes highlighted as below:
+    1. **Once the VM spec virt-controller will add two new volumes to the virt-launcher pod.**\
+        a. **EmptyDir volume named (shared-dir) is used to share the vhostuser socket file with the virt-launcher pod and dpdk dataplane.**\
+        b. **DownwardAPI volume named (podinfo) is used to share vhostuser socket file name with the virt-launcher via pod annotations.**
+    2. **The CNI will mount the shared-dir i.e., EmptyDir volume with the mount-path as "/var/lib/cni/usrcni/".**
```
alicefr, line 42: Is this mount transparent to k8s? If not, we can end up in a clean-up hell. Let's imagine that the CNI plugin is not responding and doesn't perform the unmount, what will happen if we want to delete the virt-launcher pod?

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,96 @@
...
+          interfaces:
+          - name: vhost-user-vn-blue
+            vhostuser: {}
+          useVirtioTransitional: true
+      networks:
+      - name: vhost-user-vn-blue
+        multus:
+          networkName: vn-blue
+
+```
```
alicefr, line 67: That's great thanks! Would you mind also highlighting the changes on the virt-launcher pod, like additional volumes, network interfaces, and annotations?

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,96 @@
...
+## Goals
+- The kubevirt VM when created with DPDK features, should be able to use fast datpath if the host supports the DPDK dataplane allowing fast packet processing to improve performance of the traffic.
+
+## Non Goals
+- The user space networking will require the host to a DPDK dataplane.
+- The kubevirtci creates kernel mode dataplane hosts for testing the VMs. It needs to be updated to support DPDK.
+
+## Definition of Users
+- Any user who has a (k8s) cluster with a dpdk dataplane can use the interface/feature.
+- The feature will be supported only if a valid CNI is being used.
```
maiqueb, line 22: please list a few examples of `valid CNI` .

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,96 @@
...
+   - Enable usage of dpdk on the setups created by kubevirtci.
+
+# Design
+- The design involves introducing vHostUser interface type and required parameters to support it. An EmptyDIR volume (shared-dir) and DownwardAPI (podinfo) volume are mounted on the virtlauncher pod to have the support for new interface.
+- These mounts will allow the virt-launcher pod to create a VM with an additional interface which can be reached using the vhostSocket mentioned in the DownwardAPI.
+
+- Creating the VMs will follow the same process with a few changes highlighted as below:
+    1. **Once the VM spec virt-controller will add two new volumes to the virt-launcher pod.**\
+        a. **EmptyDir volume named (shared-dir) is used to share the vhostuser socket file with the virt-launcher pod and dpdk dataplane.**\
+        b. **DownwardAPI volume named (podinfo) is used to share vhostuser socket file name with the virt-launcher via pod annotations.**
```
xpivarc, line 41: Is this required? What information will be provided this way?

The problem with `DownwardAPI` is that it is async and we can't really know when the information will be available in the context of `virt-launcher`. Therefore I would rather suggest that the `virt-controller` will proxy this information to `vmi` if we really need it. Then we can send a `sync` command at the right time.

xpivarc (commented):
Hi @nvinnakota10 ,
Would it be possible to refer to cni providing this functionality and include how it works for Pods?

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Repos
+
+- kubevirt/kubevirt
+   - Introduction of vhostuser type interface.
+- kubevirt/kubevirtci
+   - Enable usage of dpdk on the setups created by kubevirtci.
+
+# Design
+- The design involves introducing vHostUser interface type and required parameters to support it. An EmptyDIR volume (shared-dir) and DownwardAPI (podinfo) volume are mounted on the virtlauncher pod to have the support for new interface.
```
maiqueb, line 35: IIUC this will need to update virt-controller's code to update how it templates the launcher pods, right ? 

Why couldn't we just re-use the kubevirt / kubevirt-private directories that already exist in the launcher pods ? Do the "valid" CNI plugins require access to this path ?

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,96 @@
...
+- As a user/admin, I want to use user space networking feature if I have DPDK supported VM and the host has a DPDK dataplane.
+
+## Repos
+
+- kubevirt/kubevirt
+   - Introduction of vhostuser type interface.
+- kubevirt/kubevirtci
+   - Enable usage of dpdk on the setups created by kubevirtci.
+
+# Design
```
maiqueb, line 34: could you provide a sequence diagram with this info ? 

It would be easier to follow the text with a visual representation of what's / when's happening.

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,96 @@
...
+```yaml
+   devices:
+          interfaces:
+          - name: vhost-user-vn-blue
+            vhostuser: {}
+          useVirtioTransitional: true
+      networks:
+      - name: vhost-user-vn-blue
+        multus:
+          networkName: vn-blue
```
maiqueb, line 65: Would you please provide a functional example of a NAD ? 

While that's besides the kubevirt umbrella, it is still valuable to check what parameters it would have. 

If you think that's distracting here, please add it to an `annex` section in he end of the doc.

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,96 @@
...
+
+Since, KubeVirt will always explicitly define the pod interface name for multus-cni. It will be computed from the VMI spec interface name, to allow multiple connections to the same multus provided network.
+
+The vHostUser Interface will be defined in the VM spec as  shown below:
+
+```yaml
+   devices:
+          interfaces:
+          - name: vhost-user-vn-blue
+            vhostuser: {}
```
maiqueb, line 60: this implies adding yet another binding mechanism.

We've tried not to do that in the past (just a heads up ...) . 

We are actually trying to decouple this from KubeVirt, thus allowing end users to provide (and use ...) their bindings without moving the maintenance burden to KubeVirt.

maiqueb (commented):
Thanks for your effort in driving / helping move forward this by providing a design doc.

I didn't review in depth; I hope I can find the time to provide a	more comprehensive review.

design-proposals/vhostuser-interface.md
nvinnakota10: The DownwardAPI will provide the annotations containing the information on the socket file name and its location, which can be used by the virtualization-launcher pod while creating the VM using libvirt.

design-proposals/vhostuser-interface.md
```diff
@@ -0,0 +1,96 @@
...
+- kubevirt/kubevirtci
+   - Enable usage of dpdk on the setups created by kubevirtci.
+
+# Design
+- The design involves introducing vHostUser interface type and required parameters to support it. An EmptyDIR volume (shared-dir) and DownwardAPI (podinfo) volume are mounted on the virtlauncher pod to have the support for new interface.
+- These mounts will allow the virt-launcher pod to create a VM with an additional interface which can be reached using the vhostSocket mentioned in the DownwardAPI.
+
+- Creating the VMs will follow the same process with a few changes highlighted as below:
+    1. **Once the VM spec virt-controller will add two new volumes to the virt-launcher pod.**\
+        a. **EmptyDir volume named (shared-dir) is used to share the vhostuser socket file with the virt-launcher pod and dpdk dataplane.**\
```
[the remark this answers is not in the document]
nvinnakota10: The purpose of the empty dir is to act as a socket directory in the Compute container and allow qemu to create a vhostuser socket in the directory /var/lib/kubelet/pods/<podID>/volumes/kubernetes.io~empty-dir/shared-dir, which will be mounted by the CNI in the respective kubelet. The host path's mount can vary based on where the CNI wants to mount.

Moreover, to have a better transparency with k8s, emptyDir is more advantageous, as the clean up will be done by k8s, i.e., if we delete the pod, the empty directory will also be deleted.

ref: https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

design-proposals/vhostuser-interface.md
nvinnakota10: Updated the document, created a new folder to have both the proposal and sequence diagram.

design-proposals/vhostuser-interface.md
nvinnakota10: Since, k8s deletes the emptydir once the pod is deleted, the deletion of virtualization-launcher pod would not cause any issues.

design-proposals/vhostuser-interface.md
nvinnakota10: The mentioned interface and appropriate multus network details in the VM spec should suffice. Since the volumes and annotations will be handled through virt-controller, they can be skipped in the VM definition.

design-proposals/vhostuser-interface.md
nvinnakota10: One of the few available CNI that can be used in this case is UserSpace CNI by intel. Update the document to specify the same.

design-proposals/vhostuser-interface.md
nvinnakota10: Added the sequence diagram with the information explained in the Design.

design-proposals/vhostuser-interface.md
nvinnakota10: @maiqueb , Can you please give a few pointers on how this works. Thanks.

design-proposals/vhostuser-interface.md
nvinnakota10: Provided a simple functional example that can be used for NAD. The NAD definition uses UserSpace CNI (by intel) in the config, since it is used as cni with ovs-dpdk.

design-proposals/vhostuser-interface.md
nvinnakota10: Virt-controller's code won't be modified much. Similar to existing volumes and volume mounts as in pkg/virt-controller/services/rendervolumes.go, two new APIs will be used for the mentioned volumes. Since this is just a re-use of the existing functionality, I followed the same approach.

nvinnakota10:
> @nvinnakota10 I'm still missing the scheduling part or is it part of CNI? My question is will be a VMI requesting this network interface automatically scheduled on nodes where the CNI plugin is installed or does the virt-launcher pod requires additional information?

The scheduling part is done by the k8s. In the VM spec, the user can specify custom labels and use them as selectors for the scheduling part. Moreover, user can specify the huge pages required and memory requests. This will allow user to label the nodes that can support and k8s to schedule accordingly based on requirements.

design-proposals/vhostuser-interface.md
maiqueb: Keep in mind CNI deletes can be missed / fail to cleanup all resources on the pod.

design-proposals/vhostuser-interface.md
maiqueb: Sure. 

1. update the [KubeVirt API](https://github.com/kubevirt/kubevirt/blob/24a67ecc90a44cc6109465bec109f216d1af2676/staging/src/kubevirt.io/api/core/v1/schema.go#L1237) with this new interface type.
2. AFAIU, nothing needs to be done for the phase#1 (virt-handler performs some privileged operations on behalf of the launcher). Everything you need should be done by userspace-cni, right ? 
3. the launcher domain manager **must** [generate the interface domXML](https://github.com/kubevirt/kubevirt/blob/24a67ecc90a44cc6109465bec109f216d1af2676/pkg/virt-launcher/virtwrap/converter/network.go#L90) snippet for your particular interface. More stuff done [here](https://github.com/kubevirt/kubevirt/blob/8c4fa60175c11257b955553a0328b44e86a044d4/pkg/network/domainspec/generators.go#L255) (example for macvtap).

maiqueb: But as I've said in [this comment](https://github.com/kubevirt/community/pull/218#discussion_r1183788198), we're currently investigating how to allow users to extend the network binding options without requiring the core KubeVirt maintainers to maintain said bindings... We've recently discouraged users from posting new binding types ...

FYI .

@EdDev can provide more information; I just shipped around a few half baked ideas on how this could be done, but there's nothing concrete yet.

nvinnakota10: @maiqueb, I believe I am not using a binding mechanism for the vhostuser interface, as I am not doing what exactly has been done for interfaces like bridge, masquerade, PASST.

I am using this PR https://github.com/kubevirt/kubevirt/pull/3208 as a reference and most of the changes are from it, I don't see any call following a binding mechanism.

Please let me know if you think this is still not clear.

design-proposals/vhostuser-interface.md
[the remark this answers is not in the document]
alicefr: This isn't always true, see the comments in https://github.com/kubevirt/community/pull/218#discussion_r1183351913

design-proposals/vhostuser-interface.md
alicefr: I see 2 way for solving this:
 1. Specify the socket path in KubeVirt CR (as a new field) and we let virt-handler mount the socket into virt-launcher. This can follow what we have for container and hotplugged disk
 2. Introduce a new device plugin that mounts the socket. This is at least transparent to k8s and it could avoid the DownwardAPI as you can pass the socket path as an environmental variable.

I used the second approach for the SCSI reservation https://github.com/kubevirt/kubevirt/pull/9177. You could implemented as an external plugin and then we just need to teach virt-launcher to use the new environmental variable

design-proposals/vhostuser-interface.md
maiqueb: Hm, you may be right, seems we also don't need to generate the interface domXML snippet (on the launcher pod):
https://github.com/kubevirt/kubevirt/pull/3208/files#diff-030aa8b7ff9bf4a511d1d5641c1dec04205a944137672be3a691f0b1b4068753R93

design-proposals/vhostuser-interface.md
maiqueb: Given these 2 choices, I'd go for the latter (device plugin) since it's native to kubernetes.

alicefr: The downside with the device plugin approach is that is implying exposing a resource on the node.  You might or might not expose it in this way. That was one of the reason I was asking about the scheduling [here](https://github.com/kubevirt/community/pull/218#issuecomment-1532602246)

nvinnakota10: @alicefr , @maiqueb,

The main reason I am using EmptyDir and DownwardAPI in my implementation is that they allow transparency between CNI and the kubemanager.

I am not aware of Device Plugin type, will I be able to achieve the transparency between CNI and the kubemanager as they need to use the same name i.e., they should have the same name for the vhostSocket file.

design-proposals/vhostuser-interface.md
EdDev: While pod networking configuration is not needed by virt-handler, the domain configuration is surely needed [1].

I am thinking that we can generalize these types of bindings. I.e noop bindings which do not touch the pod networking.
But we are still left with the logic of how to define the domain spec (libvirt domain config), including NUMA, hugepages, etc.

[1] https://github.com/kubevirt/kubevirt/pull/3208/files#diff-bc5b4c898f931f6ffeea71321777a71452dcb2e521946f77d28a72a885154a53

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,143 @@
...
+   - Enable usage of dpdk on the setups created by kubevirtci.
+
+# Design
+- The design involves introducing vHostUser interface type and required parameters to support it. An EmptyDIR volume (shared-dir) and DownwardAPI (podinfo) volume are mounted on the virtlauncher pod to have the support for new interface.
+- These mounts will allow the virt-launcher pod to create a VM with an additional interface which can be reached using the vhostSocket mentioned in the DownwardAPI.
+
+- Creating the VMs will follow the same process with a few changes highlighted as below:
+    1. **Once the VM spec virt-controller will add two new volumes to the virt-launcher pod.**\
+        a. **EmptyDir volume named (shared-dir) "/var/lib/cni/usrcni/" is used to share the vhostuser socket file with the virt-launcher pod and dpdk dataplane. This socket file acts as an UNIX socket between the VM interface and the dataplane running in the UserSpace.**\
+        b. **DownwardAPI volume named (podinfo) is used to share vhostuser socket file name with the virt-launcher via pod annotations.**
```
EdDev, line 41: If the `EmptyDir` mounts the actual socket to a specific path, why do we need the DownwardAPI ?
What information needs to arrive to the virt-launcher filesystem and why it cannot be passed through a field in the VMI (e.g. the in the interface details).

EdDev (commented):
Sorry for the late feedback on this.
There is a lot to digest here and it will probably take me some time to learn all the details, sorry in advance.

As @maiqueb mentioned earlier, we are trying to find a elegant way to decouple as much as possible the different network interface types to ease and distribute the maintenance effort.
This type seems not to require the first step which configures the pod networking, only the second step where there is a need to define the domain configuration to consume the endpoint (it this case a socket).

In the meantime, I think the discussion on this proposals can continue as usual.

design-proposals/vhostuser-interface.md
xpivarc: Do you want to ever support migration? If so the (even without it) it would be required/preferable to not use the `DownwardAPI` rather us `virt-controller` which can propagate the annotations from `virt-launcher`  to `vmi`.

EdDev: > Do you want to ever support migration? If so the (even without it) it would be required/preferable to not use the `DownwardAPI` rather us `virt-controller` which can propagate the annotations from `virt-launcher` to `vmi`.

It is already in use for SR-IOV, by itself it is not a blocker for migration.
But if there are other reasonable options, I agree it would be preferable.

I mainly want to understand why it is needed in more details ([ref](https://github.com/kubevirt/community/pull/218#discussion_r1189408556)).

design-proposals/vhostuser-interface.md
nvinnakota10: @maiqueb, sorry for the confusion. I am actually using the binding interface method. Like I mentioned I made a few changes, and forgot I was using the interface binding method for vhostuser interface.

https://github.com/kubevirt/kubevirt/compare/release-0.59...Juniper:kubevirt:release-0.59

stefanha:
Please use the term vhost-user-net to refer to the network device since there are other vhost-user devices too (blk, fs, gpio, etc). KubeVirt may support other vhost-user devices in the future, so it helps to be clear what is specific vhost-user-net and what is generic vhost-user infrastructure that other devices will also use.

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,143 @@
...
+
+- As a result, when we have a host with DPDK dataplane running in Userspace, the traffic from kubevirt VM will go through the kernel and reach the dataplane running in userspace. This longer path slows down the traffic between the application and the dataplane when both are running in userspace.
+
+- In order to improve the traffic performance and make use of the dpdk, we need to use a different path that can by-pass the kernel and reach the dataplane from the user space directly. This so called fast path will run within the user space and will allow fast packet processing.
+
+## Goals
+- The kubevirt VM when created with DPDK features, should be able to use fast datpath if the host supports the DPDK dataplane allowing fast packet processing to improve performance of the traffic.
+
+## Non Goals
+- The user space networking will require the host to a DPDK dataplane.
```
stefanha, line 17: Can you rephrase this? I don't understand this sentence.

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,143 @@
...
+- The kubevirtci creates kernel mode dataplane hosts for testing the VMs. It needs to be updated to support DPDK.
```
stefanha, line 18: This is listed under "Non Goals", so do you mean testing with DPDK is not a goal? There will be no tests?

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Definition of Users
+- Any user who has a (k8s) cluster with a dpdk dataplane can use the interface/feature.
+- The feature will be supported only if a multus userspace CNI is being used, for instance userspace cni from intel.
+
+## User Stories
+- As a user/admin, I want to use user space networking feature if I have DPDK supported VM and the host has a DPDK dataplane and userspace cni with multus support.
```
stefanha, line 25: What does "DPDK supported VM" mean?

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,143 @@
...
+- These mounts will allow the virt-launcher pod to create a VM with an additional interface which can be reached using the vhostSocket mentioned in the DownwardAPI.
+
+- Creating the VMs will follow the same process with a few changes highlighted as below:
+    1. **Once the VM spec virt-controller will add two new volumes to the virt-launcher pod.**\
+        a. **EmptyDir volume named (shared-dir) "/var/lib/cni/usrcni/" is used to share the vhostuser socket file with the virt-launcher pod and dpdk dataplane. This socket file acts as an UNIX socket between the VM interface and the dataplane running in the UserSpace.**\
+        b. **DownwardAPI volume named (podinfo) is used to share vhostuser socket file name with the virt-launcher via pod annotations.**
+    2. **The CNI will mount the shared-dir i.e., EmptyDir volume's host path /var/lib/kubelet/pods/<podID>/volumes/kubernetes.io~empty-dir/shared-dir . This will be deleted, to delete the vHostUser Interface while deleteing the VM and/or virt-launcher pod.**
```
stefanha, line 42: It is unclear to me who deletes what and when. Are you saying KubeVirt deletes the shared-dir and any files inside it when the VM terminates?

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,143 @@
...
+    3. **The CNI will update the virt-launcher pod annotations with vhostsocket-file name and details.**
+    4. **The virt-launcher reads the DownwardAPI volume and retrieves the vhostsocket-file name specified in pod annotations and uses it to create a unix socket in the VM, while launching the VM using libvirtd.**
+    4. **The virt-launcher is modified to skip establishing the networking between VM and the virt-launcher pod using bridge(Refer in Kubevirt Networking section in https://kubevirt.io/2018/KubeVirt-Network-Deep-Dive.html).**
+    5. **Instead of using the bridge through the launcher pod, the vHostUser interface of the VM will be directly connected to the DPDK datplane using the vhost socket**
```
stefanha, line 46: It sounds like `<source type='unix' path='/tmp/vhost1.sock' mode='server'/>` will be used so QEMU creates the UNIX domain socket and then something must trigger DPDK to connect to this socket?

Which component sends a command to DPDK to establish the UNIX domain socket connection that QEMU is waiting for so the VM can start?

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,143 @@
...
+- VM migration works the same way, as the only change is having a new interace, which only requires a dataplane of the host in userpsace.
+
+## Update/Rollback Compatibility
+- Should have no impact.
+
+## Functional Testing Approach
+Functional test can:
+
+- Create VM with vhostUser interface
+- Create VM with multiple interfaces
```
stefanha, line 79: Do you mean mixed vhost-user-net and traditional virtio-net devices or multiple vhost-user-net devices?

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,143 @@
...
+
+- Creating the VMs will follow the same process with a few changes highlighted as below:
+    1. **Once the VM spec virt-controller will add two new volumes to the virt-launcher pod.**\
+        a. **EmptyDir volume named (shared-dir) "/var/lib/cni/usrcni/" is used to share the vhostuser socket file with the virt-launcher pod and dpdk dataplane. This socket file acts as an UNIX socket between the VM interface and the dataplane running in the UserSpace.**\
+        b. **DownwardAPI volume named (podinfo) is used to share vhostuser socket file name with the virt-launcher via pod annotations.**
+    2. **The CNI will mount the shared-dir i.e., EmptyDir volume's host path /var/lib/kubelet/pods/<podID>/volumes/kubernetes.io~empty-dir/shared-dir . This will be deleted, to delete the vHostUser Interface while deleteing the VM and/or virt-launcher pod.**
+    3. **The CNI will update the virt-launcher pod annotations with vhostsocket-file name and details.**
+    4. **The virt-launcher reads the DownwardAPI volume and retrieves the vhostsocket-file name specified in pod annotations and uses it to create a unix socket in the VM, while launching the VM using libvirtd.**
```
stefanha, line 44: Can you add section that specifies the DownwardAPI annotations?

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: I will change the line. I will be testing VMs with different interfaces types like bridge etc.

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,143 @@
...
+- kubevirt/kubevirtci
+   - Enable usage of dpdk on the setups created by kubevirtci.
+
+# Design
+- The design involves introducing vHostUser interface type and required parameters to support it. An EmptyDIR volume (shared-dir) and DownwardAPI (podinfo) volume are mounted on the virtlauncher pod to have the support for new interface.
+- These mounts will allow the virt-launcher pod to create a VM with an additional interface which can be reached using the vhostSocket mentioned in the DownwardAPI.
+
+- Creating the VMs will follow the same process with a few changes highlighted as below:
+    1. **Once the VM spec virt-controller will add two new volumes to the virt-launcher pod.**\
+        a. **EmptyDir volume named (shared-dir) "/var/lib/cni/usrcni/" is used to share the vhostuser socket file with the virt-launcher pod and dpdk dataplane. This socket file acts as an UNIX socket between the VM interface and the dataplane running in the UserSpace.**\
```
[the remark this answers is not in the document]
nvinnakota10: The proposed changes can help us with the below scenarios:

1. VM with multiple vhostUser Interfaces
2. Multiple VMs with one or more vHostUser Interfaces present on the same compute.

The changes proposed will allow user to create any number of interfaces, as long as the interface allocation is within the node limits i.e, permitted by the node.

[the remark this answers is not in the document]
stefanha: When a VM has multiple vhost-user-net UNIX domain sockets, do they all go in the same shared-dir? How are the socket names communicated (the text mentions a singular socket file but doesn't explain how it works when there are many socket files).

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: By "DPDK supported VM" I mean that it should have at least one interface binded to DPDK and should allow running DPDK applications on the VM.

design-proposals/vhostuser/vhostuser-interface.md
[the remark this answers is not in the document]
nvinnakota10: Yes, they all go to the same shared-dir. This shared directory can be linked to a custom directory in the host, by the CNI. CNI can then give a unique name to each socket created this way. 

Naming strategy might vary from CNI to CNI to make sure the sockets created are unique.

In a generic case, We can make the CNI to link the shared-dir to a custom path/pod-uuid/socket-name

The user should make sure to know how the CNI handles adding a virtio interfaces for dpdk/vpp and make necessary changes to have the sockets stored are unique in name.

Hope that answers what you are looking for.

sbrivio-rh:
By the way, I'm not sure this is relevant, but we're working to add [vhost-user support](https://gitlab.com/lvivier/passt/-/tree/vhost-user) to [passt](https://passt.top), and KubeVirt could very reasonably use it for improved throughput once it's ready.

The current QEMU and passt command lines are something on the lines of:

```
./passt --vhost-user -s /tmp/passt_1.socket

qemu-system-x86_64 ... -m 4G \
             -object memory-backend-memfd,id=memfd0,share=on,size=4G \
             -numa node,memdev=memfd0 \
             -chardev socket,id=chr0,path=/tmp/passt_1.socket \
             -netdev vhost-user,id=netdev0,chardev=chr0 \
             -device virtio-net,netdev=netdev0 \
             ...
```

of course, the aim is different from DPDK: passt aims at unprivileged isolation rather than direct access, and it lives in the pod itself (no need to pierce through namespaces to get that to the UNIX socket), so I guess it should have no influence on this proposal, but I thought I'd mention it anyway.

design-proposals/vhostuser/vhostuser-interface.md
stefanha: My understanding is this feature only relates to a vhost-user-net backend (like DPDK) running on the host. The VM sees a virtio-net device and there is no need to run DPDK inside the guest. The only requirement is that the VM has a virtio-net guest driver.

I think "DPDK supported VM" part can be removed from the user story.

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: I have added an example on how DownwardAPI looks.

design-proposals/vhostuser/vhostuser-interface.md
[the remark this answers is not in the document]
nvinnakota10: I have rephrased implementation part with more details and explanation.

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: I have updated the design implementation with more details.

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: We have a test plan to bring up a VM with vHostUser interface and another interface type like Bridge.

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: Updated the design proposal.

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: Updated the design proposal explaining the requirements of the VM like having dpdk virtio interfaces to use these features.

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: @stefanha , I didn't get your question.

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: @EdDev , Also the CNI/KubeManager will be creating the socket files while creating the virtio interfaces. So, to have the consistency, we expect that the CNI uses the socket name reperesenting the poduid+intfname+id. This info will not be available from the VM YAML spec.

DownwardAPI will help us maintain consistency as the CNI creates the sockets in a custom path, which will be shared to the launcher pod. This socket name will be passed to qemu.

alicefr:
> By the way, I'm not sure this is relevant, but we're working to add [vhost-user support](https://gitlab.com/lvivier/passt/-/tree/vhost-user) to [passt](https://passt.top), and KubeVirt could very reasonably use it for improved throughput once it's ready.
> 
> The current QEMU and passt command lines are something on the lines of:
> 
> ```
> ./passt --vhost-user -s /tmp/passt_1.socket
> 
> qemu-system-x86_64 ... -m 4G \
>              -object memory-backend-memfd,id=memfd0,share=on,size=4G \
>              -numa node,memdev=memfd0 \
>              -chardev socket,id=chr0,path=/tmp/passt_1.socket \
>              -netdev vhost-user,id=netdev0,chardev=chr0 \
>              -device virtio-net,netdev=netdev0 \
>              ...
> ```
> 
> of course, the aim is different from DPDK: passt aims at unprivileged isolation rather than direct access, and it lives in the pod itself (no need to pierce through namespaces to get that to the UNIX socket), so I guess it should have no influence on this proposal, but I thought I'd mention it anyway.

@sbrivio-rh the setup for passt is for sure different, but I think the KubeVirt API extension for vhost-user-net interface can help for passt too. I'm referring to the discussion in https://github.com/kubevirt/community/pull/218#discussion_r1183788198

stefanha:
> Hi @nvinnakota10 , Would it be possible to refer to cni providing this functionality and include how it works for Pods?

:+1: 
Yes, please. It's worth defining the interface for regular k8s Pods (without KubeVirt in the picture). For example, someone may want to run a DPDK application in a regular Pod attached to a DPDK software switch provided by CNI using vhost-user.

design-proposals/vhostuser/vhostuser-interface.md
stefanha: I don't understand why this point is listed under "Non Goals". Non goals are things that are out of scope. For example, supporting PCI network card device assignment would be a non goal because it's beyond what vhost-user-net is about.

Maybe this point about extending kubevirtci should go in the "Goals" section?

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,166 @@
...
+
+- As a result, when we have a host with DPDK dataplane running in Userspace, the traffic from kubevirt VM will go through the kernel and reach the dataplane running in userspace. This longer path slows down the traffic between the application and the dataplane when both are running in userspace.
+
+- In order to improve the traffic performance and make use of the dpdk, we need to use a different path that can by-pass the kernel and reach the dataplane from the user space directly. This so called fast path will run within the user space and will allow fast packet processing.
+
+## Goals
+- The kubevirt VM when created with DPDK features, should be able to use fast datpath if the host supports the DPDK dataplane allowing fast packet processing to improve performance of the traffic.
+
+## Non Goals
+- The Userspace CNI with the dataplane running in userspace like dpdk/vpp, should make sure proper handling of mounting and unmounting the volumes. These changes are CNI specific and might vary from CNI to CNI.
```
stefanha, line 17: I'm not sure what the first sentence means?

The second sentence sounds like the non goal is "Automatically supporting existing CNI plugins. It will be necessary to modify existing CNI plugins in order to support vhost-user-net.".

design-proposals/vhostuser/vhostuser-interface.md
stefanha: The vhost-user protocol allows the frontend and backend to act as the client or server. This is configurable and QEMU supports both acting as a socket client and a server.

In your proposal the socket seems to be created by QEMU (the frontend) and DPDK (the backend) connects to the UNIX domain socket. (I say this because you wrote "The __virt-launcher__ reads the DownwardAPI volume and retrieves the vhostsocket-file name specified in pod annotations and uses it to __create__ a unix socket in the VM".)

My question is about the order in which QEMU creates the UNIX domain socket and DPDK connects to it:

How does DPDK know when it is time to connect to the UNIX domain socket since the socket is created after virt-launcher Pod launches?

Does DPDK need to attempt to reconnect in a loop or is there a step where DPDK is notified that it's time to connect?

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,166 @@
...
+        - Add E2E test to send traffic between the 2 VMs created.
+
+## Annex
+
+
+A sample DownwardAPi mght look this way:
+
+```yaml
+k8s.v1.cni.cncf.io/network-status="[{\n    \"name\": \"k8s-kubemanager-kubernetes-CNI/default-podnetwork\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.244.2.4\"\n    ],\n    \"mac\": \"02:d7:b3:27:b3:b2\",\n    \"default\": true,\n    \"dns\": {}\n},{\n    \"name\": \"kubevirttest/vn-blue\",\n    \"interface\": \"net1\",\n    \"ips\": [\n        \"19.1.1.2\"\n    ],\n    \"mac\": \"02:c6:da:d3:ab:07\",\n    \"dns\": {},\n    \"device-info\": {\n        \"type\": \"vhost-user\",\n        \"vhost-user\": {\n            \"mode\": \"server\",\n            \"path\": \"2ea1931c-2935-net1\"\n        }\n    }\n}]"
+k8s.v1.cni.cncf.io/networks="[{\"interface\":\"net1\",\"name\":\"vn-blue\",\"namespace\":\"kubevirttest\",\"cni-args\":{\"interface-type\":\"virtio\"}}]"
```
stefanha, line 105: What does `"interface-type":"virtio"` mean? Is that related to vhost-user-net or is net1 exposed as a virtio-net device somehow?

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,166 @@
...
+kubernetes.io/config.seen="2023-05-16T16:50:04.224937236Z"
+kubernetes.io/config.source="api"
+kubevirt.io/domain="vm-single-virtio"
+kubevirt.io/migrationTransportUnix="true"
+post.hook.backup.velero.io/command="[\"/usr/bin/virt-freezer\", \"--unfreeze\", \"--name\", \"vm-single-virtio\", \"--namespace\", \"kubevirttest\"]"
+post.hook.backup.velero.io/container="compute"
+pre.hook.backup.velero.io/command="[\"/usr/bin/virt-freezer\", \"--freeze\", \"--name\", \"vm-single-virtio\", \"--namespace\", \"kubevirttest\"]"
+pre.hook.backup.velero.io/container="compute"
+```
+The CNI should update the /ect/podnetinfo/annotaions with something similar, to update the annotations of the virt-launcher pod.
```
stefanha, line 116: Should the path be /etc/podnetinfo/annotations?

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,166 @@
...
+
+```yaml
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: userspace-ovs-net-1
+spec:
+  config: '{
+        "cniVersion": "0.3.1",
+        "type": "userspace",
```
stefanha, line 128: Why is it called "userspace" and not "vhost-user-net"? Is the idea to support other userspace network interfaces in the future?

maiqueb: this must have the name of the CNI binary which will run on the host. 

AFAIU, the "reference" CNI to use is userspace CNI, whose binary is named `userspace`: https://github.com/intel/userspace-cni-network-plugin#work-standalone

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: Yes, the path is set that way to update the pod annotations.

design-proposals/vhostuser/vhostuser-interface.md
stefanha: @maiqueb @nvinnakota10 
Thanks for sharing the link to the Userspace CNI plugin! It seems the Userspace CNI plugin doesn't define how Pods need to be configured - it's left up to the user to create volumes, choose paths, and I'm not sure it uses annotations and the downward API.

I think a standard interface should be defined by the Userspace CNI plugin, not in KubeVirt. Then other container workloads can benefit from a standard interface that doesn't require coming up with configuration (volumes, paths, etc) from scratch.

The KubeVirt proposal would just be about implementing the Userspace CNI interface, not defining annotations, etc.

stefanha: As a first step towards this, I suggest splitting this proposal into two parts:
1. The standard Userspace CNI plugin interface that allows any Pod to detect vhost-user-net devices and consume them. This will eventually go into the Userspace CNI plugin project (and other CNI plugins can implement a compatible interface).
2. The KubeVirt-specific part that describes how to implement the interface in virt-launcher, etc.

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: @stefanha , The UNIX socket will be created in the host in the custom path linked to the shared directory  and the qemu then connects to it. The socket details will be passed to the qemu by the virtualization-launcher while creating the VM, which gets this information from the downwardAPI passed by the Kubemanager.

stefanha:
@nvinnakota10 I don't see a URL for a video conference in the kick-off event calendar invite you sent. Can you share the URL?

EdDev:
> @nvinnakota10 I don't see a URL for a video conference in the kick-off event calendar invite you sent. Can you share the URL?

https://meet.google.com/adw-nnfz-uad?authuser=0

design-proposals/vhostuser/vhostuser-interface.md
stefanha: Okay. There is a typo in the path on line 116.

design-proposals/vhostuser/vhostuser-interface.md
stefanha: Does the NAD below need to be updated too?

```
        "host": {
                "engine": "ovs-dpdk",
                "iftype": "vhostuser",
                "netType": "bridge",
                "vhost": {
                        "mode": "client"
                },
                "bridge": {
                        "bridgeName": "br-dpdk0"
                }
        },
...
        "container": {
                "engine": "ovs-dpdk",
                "iftype": "vhostuser",
                "netType": "interface",
                "vhost": {
                        "mode": "server"
                }
        },
```

Maybe "mode" needs to be reversed for "host" and "container".

Do you intend to support both configurations, or just DPDK acting as the server and QEMU acting as the client?

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: If the user is trying to create a vhostUser interface, we are passing the interface type as virtio in the multus cni arguments in the multusNetworkAnnotation from kubevirt code. This is the reason we see that in the pod annotations.

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: The changes proposed are specific to kubevirt itself. The changes are generic. Since interface addition and deletion is handled by a CNI, CNI should handle the mount and unmount the volumes where the sockets can be created.

This is not achieved by the changes proposed, User should make sure that CNI is capable of doing this or need to make necessary changes to the CNI to use this feature. This is the reason it was mentioned in the Non Goals section.

design-proposals/vhostuser/vhostuser-interface.md
stefanha: Why "virtio" and not "vhost-user-net"?

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,166 @@
...
+    4. **The virt-launcher reads the DownwardAPI volume and retrieves the vhostsocket-file name specified in pod annotations and uses it while launching the VM using libvirtd.**
+    4. **The virt-launcher is modified to skip establishing the networking between VM and the virt-launcher pod using bridge(Refer in Kubevirt Networking section in https://kubevirt.io/2018/KubeVirt-Network-Deep-Dive.html).**
+    5. **Instead of using the bridge through the launcher pod, the vHostUser interface of the VM will be directly connected to the DPDK datplane using the vhost socket**
+
+- Removing the VMs will follow the below chnages:
+    1. **The CNI should delete the socket files of each interface. If 'n' interfaces are present, 'n' number of socket files will be deleted allowing k8s to proceed with a clean pod deletion.**
+    
+   No additional changes are required while deleting the VM, it will be deleted as a regular VM, the CNI should handle the deletion of interfaces. In the case of vHostUser, CNI should delete the socket files before k8s initiates pod deletion.
```
alicefr, line 51: I apologize if I insist on this point, but does kubelet request again to CNI the deletion of the network interface if there is a failure? If so, then the cleanup shouldn't be a problem

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,166 @@
...
+
+- Creating the VMs will follow the same process with a few changes highlighted as below:
+    1. **Once the VM spec virt-controller will add two new volumes to the virt-launcher pod.**\
+        a. **EmptyDir volume named (shared-dir) "/var/lib/cni/usrcni/" from virt-launcher pod is used to share the vhostuser socket file with the virt-launcher pod and dpdk dataplane. This socket file acts as an UNIX socket between the VM interface and the dataplane running in the UserSpace.**\
+        b. **DownwardAPI volume named (podinfo) is "/etc/podnetinfo/annotations" used to share vhostuser socket file name with the virt-launcher via pod annotations. The downwardAPI is used here to have the pod know the socket details like name and path, which is created by CNI/kubemanager while bringingup the virt-launcher pod and this info is only availbale after the pod is created**
+    2. **The CNI should mount the shared-dir i.e., EmptyDir volume's host path /var/lib/kubelet/pods/<podID>/volumes/kubernetes.io~empty-dir/shared-dir by linking it to a custom path CNI prefers to have. The CNI should create the sockets in the host in the custom path. For each interface, CNI should create a socket.**
```
alicefr, line 42: Can we guarantee that the socket is always created before we start QEMU?

design-proposals/vhostuser/vhostuser-interface.md
maiqueb: AFAIU no; depending on the implementation, kubelet may not even be aware something failed.

The [CNI spec](https://github.com/containernetworking/cni/blob/main/SPEC.md#del-remove-container-from-network-or-un-apply-modifications) in fact indicates a CNI DEL should **not** return an error:
> Plugins should generally complete a DEL action without error even if some resources are missing.

Some entity should reconcile the resources afterwards ensuring those are gone.

alicefr: Then, we might still have a problem here :/ @nvinnakota10 if you have a working prototype, I'd be really interested to know what happens if you try to kill the component doing the unmount and at the same time deleting the VMI/pod.

design-proposals/vhostuser/vhostuser-interface.md
maiqueb: If the socket is created by CNI, it is indeed guaranteed; first CNI creates creates the pod sandbox, and only after that are the pod's containers started.

design-proposals/vhostuser/vhostuser-interface.md
nvinnakota10: It might have come up there as we use that name in out CNI. We can change it to host-user-net and have it as a standard.

nvinnakota10: I have updated the cni arguments in the annotations. This value might not be used by every CNI, however any CNI wanted to use the CNI arguments can use this as a standard if needed.

stefanha: Thanks. "virtio" or "virtio-net" can mean something different from "vhost-user-net" (e.g. a virtio-net PCI or vdpa-net device). Calling it "vhost-user-net" is clearer.

design-proposals/vhostuser-interface.md
nvinnakota10: @maiqueb , @EdDev 

Can you please confirm if we can proceed with the interface binding method, or do we have any work in progress for the same as mentioned in the  comment:
https://github.com/kubevirt/community/pull/218#discussion_r1183788198

Thanks

design-proposals/vhostuser/vhostuser-interface.md
TReZ42: > @EdDev , Also the CNI/KubeManager will be creating the socket files while creating the virtio interfaces. So, to have the consistency, we expect that the CNI uses the socket name reperesenting the poduid+intfname+id. This info will not be available from the VM YAML spec.
> 
> DownwardAPI will help us maintain consistency as the CNI creates the sockets in a custom path, which will be shared to the launcher pod. This socket name will be passed to qemu.

@nvinnakota10, if socket name is built after `poduid+intfname+id`, live migration of VM will fail because a new pod  will be created where the running VM will be transferred to, and qemu will be started with the same parameters including the path/socketname.
But obviously `poduid` changed and the CNI created a new socket named after `newpodui+intfname+id`, and uses also this name to configure the vswitch port. qemu will then hang waiting the socket `poduid+intfname+id` to be opened by the client (vswitch).
We did experience this issue with userspace CNI, and had to explicitly specify socket name in the userspace CNI config in order to have a successful VM live migration:

```yaml
        "cniVersion": "0.3.1",
        "type": "userspace",
        "name": "userspace-vpp-250-1",
        "KubeConfig": "/etc/cni/net.d/multus.d/multus.kubeconfig",
        "SharedDir": "/var/lib/cni/usrcni/",
        "logFile": "/var/log/userspace-vpp-250-1.log",
        "logLevel": "debug",
        "host": {
                "engine": "vpp",
                "iftype": "vhostuser",
                "netType": "bridge",
                "vhost": {
                        "mode": "client",
                        "socketfile": "net1"
                },
                "bridge": {
                        "bridgeName": "250"
                }
        }
```

The socket name is `poduid+intfname+id` is the default one in userspace CNI. Maybe it worth changing this default naming in the CNI (standard interface?) or change the qemu parameters regarding the vhost-user-net socketname when live migrate a VM?

nvinnakota10: @TReZ42 , Yes, we will be have to look at the Migration part, as the new pod created might still have the same socket name, but the downwardAPI might create a new socket config based on the podUID. This might cause the VM to fail waiting for the socket.

I will check the case with multi interface as well. I am not sure if the interface supports live migration. Though it is a secondary interface, it can fail live migration to fail if not handled appropriately, as the socket names might change, as PodUId changes and the VM should have the appropriate socket name.

nvinnakota10: @TReZ42 , One possible way is to hashing the combination of name+namespace of the VM created. But for this, can you please confirm if we can Live Migrate the VMs using VirtualMachineInstanceReplicaSet kind. How do we handle the names of the VM when VirtualMachineInstanceReplicaSet kind is used to create the VMs.

TReZ42: @nvinnakota10, VMI names are created after the template `name` metadata if it exists, or VirtualMachineInstanceReplicaSet name if not, appended with an id: `testvmi5wfxp`, `testvmibjkpz`.

We can also think about VirtualMachinePool: the VMs and VMIs names are created after VirtualMachinePool name, appended with -0, -1, etc. : `test-vmp-0`, `test-vmp-1`.

We can Live Migrate each of the VMI created by VirtualMachineInstanceReplicaSet or VirtualMachinePool, but not directly the VirtualMachineInstanceReplicaSet or the VirtualMachinePool.

VMI name does not change after live migration, so use a hash of `vmi+namespace+intfname` as socket name could be OK. But userspace CNI must also be adapted to use socket names according to this naming convention...

Don't you think it is possible to let the CNI defines the socket name, and for live migration patch the domain of the target VMI with the new socket name ?

nvinnakota10: @TReZ42 , Yes. This part needs to be handled by the CNI. For the changes I am planning to upstream for vHostUser interface, Kubevirt need no information on how the socket name is derived. But the CNI needs to make sure that the socket name needs to be consistent, if they need Live Migration.

Also, if we create VMs using replicasets, the hash appended at the end of the VMI name is immutable. So, even if we Live Migrate these VMs they still have the same name in the VMI object, which shouldn't affect the Socket name consistency.

I was asking the question to make sure if the approach make sense, or do I need to look at other possibilities for having CNI to maintain consistent names.

TReZ42: @nvinnakota10, if I understand correctly, you want the CNI to derive the socket name from the VM name (VMI, VMIRS, VMPool) ? But is this info available to the CNI in Downward API? Perhaps in annotations?

It's not possible to change the target domain with updated socket name, generated by the CNI as usual, the way you are doing in `createDomainInterfaces` with the vhostuser case?

nvinnakota10: Yes, I am using the createDomainInterfaces. As soon as the CNI creates the socket name and interface name, it will store the annotations in the path mounted for DownwardAPI for the launcher pod.

The createDomainInterfaces will retrieve this information to create the socket using qemu args.

design-proposals/vhostuser/vhostuser-interface.md
```diff
@@ -0,0 +1,166 @@
+
+# Overview
+- User Space networking is a feature that involves packet path from the VM to the data plane running in userspace like OVS-DPDK etc, by-passing the kernel.
```
kvaps, line 3: Hi, any idea to make vhost-user interface available for block devices and qemu-storage-daemon as well?

vasiliy-ul:
Hey, @nvinnakota10 and all. Raised this also in the mailing list, but just decided to duplicate here. I am curious about the status of this PR. Any chance we can move this forward? What are the current blockers? Ping @maiqueb, @EdDev, @xpivarc, @alicefr.

alicefr:
@vasiliy-ul I might be wrong, but I think this is blocked by the fact that we would like to separate the network plugin externally and we are missing the infrastructure for this

vasiliy-ul:
> @vasiliy-ul I might be wrong, but I think this is blocked by the fact that we would like to separate the network plugin externally and we are missing the infrastructure for this
> 
> UPDATE: this might be already available from this PR: [kubevirt/kubevirt#10284](https://github.com/kubevirt/kubevirt/pull/10284)

@alicefr, thanks for the link :+1:  I was not aware of that PR. Will take a closer look.

I've seen this concern raised here in this PR, but from the conversations it is not clear if it's a blocker or not: https://github.com/kubevirt/community/pull/218#discussion_r1210607218 and https://github.com/kubevirt/community/pull/218#pullrequestreview-1419855883

So, does that mean that from now on this new API is the way to go if we need to introduce network bindings?

alicefr:
> I've seen this concern raised here in this PR, but from the conversations it is not clear if it's a blocker or not: [#218 (comment)](https://github.com/kubevirt/community/pull/218#discussion_r1210607218) and [#218 (review)](https://github.com/kubevirt/community/pull/218#pullrequestreview-1419855883)
> 
> So, does that mean that from now on this new API is the way to go if we need to introduce network bindings?

This was at least my understanding, @EdDev could you please confirm this?

design-proposals/vhostuser/vhostuser-interface.md
stefanha: @alicefr and @albertofaria have investigated vhost-user-blk devices in the past separately from this PR. Perhaps they can share a proposal or the current state of the discussion with you.

maiqueb:
The networking bindings were already refactored, and now it is possible for community provided bindings to be provided out of (KubeVirt) tree.

Should this proposal be revisited under these new terms ?

sbrivio-rh:
> Should this proposal be revisited under these new terms ?

Which prompts an update from my side: we're about to merge [vhost-user support](https://archives.passt.top/passt-dev/20240202141151.3762941-1-lvivier@redhat.com/) in passt.

I guess it's going to be transparent for KubeVirt, that is, libvirt will eventually pass different options to both passt (`--vhost-user` and perhaps different socket path) and QEMU (essentially the `memory-backend-memfd` object).

So, that support doesn't aim in any way at replacing this proposal, and probably it doesn't integrate with it at all, but please let me know if opinions differ.

maiqueb:
> > Should this proposal be revisited under these new terms ?
> 
> Which prompts an update from my side: we're about to merge [vhost-user support](https://archives.passt.top/passt-dev/20240202141151.3762941-1-lvivier@redhat.com/) in passt.
> 
> I guess it's going to be transparent for KubeVirt, that is, libvirt will eventually pass different options to both passt (`--vhost-user` and perhaps different socket path) and QEMU (essentially the `memory-backend-memfd` object).
> 
> So, that support doesn't aim in any way at replacing this proposal, and probably it doesn't integrate with it at all, but please let me know if opinions differ.

Well, while it doesn't intend to replace this proposal, it surely brings forward a "cheaper" alternative. 

It would be interesting to see an "alternatives" section in this proposal focusing on that @sbrivio-rh . 

Thanks for raising it.

fabiand:
@alicefr @EdDev @phoracek  in the light of the other plugin based vhost user proposal - What should we do about this one? Can it be closed?

alicefr:
@fabiand I would, at least, wait until we see the other proposal. Otherwise, @nvinnakota10 are you still working on it?

fabiand:
Fair point.
When are we expecting the new proposal?

bgaussen:
Hi All,

We are still working on a proper implementation of the vhostuser network binding plugin and all stuff around:

- device plugin
- device-info spec implementation
- downward API support for plugin in Kubevirt 1.3
- and a workaround for Kubevirt 1.2 -> mutating webhook

Benoit.

/remove-lifecycle rotten
/remove-lifecycle stale

PR #219, opened 2023-05-02
proposal: Monitoring code refactor

machadovilaca:
This design doc is proposing a code refactor for the monitoring logic in all
KubeVirt components.

### DIFF
--- a/design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
+++ b/design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
@@ -1,0 +1,99 @@
+# Monitoring code refactor
+
+**Authors:** João Vilaça <jvilaca@redhat.com>
+
+## What
+
+This design document is proposing a code refactor for the monitoring logic in
+all KubeVirt components. The goal is to have a consistent monitoring package
+between all projects, a code struct that is easy to maintain and evolve, while
+moving closer to the Kubernetes metric implementation style.
+
+## Why
+
+It’s important to refactor the monitoring code to separate monitoring logic from
+business logic. It will make the codebase more modular, readable, and
+maintainable. It will reduce complexity, reduce the risk of introducing errors,
+and allow us to test each component independently.
+
+Right now, every change regarding monitoring needs to happen at many different
+places. For example, when adding the linter for metrics names, it was necessary
+to go to all KubeVirt components repositories and implement the linter
+validation in each. This is demanding, very time-consuming, requires going
+through multiple duplicated code reviews, uses many CI resources, and makes it
+mentally harder to decide to change anything.
+
+### Pitfalls of the current solution
+
+We are facing multiple challenges with the current solution:
+
+- It is hard to keep a mental model of how monitoring is implemented in all the
+different repositories and is difficult to maintain all of them in sync with new
+features
+
+- Hard to add new features to metrics and alerts, for example, because of how
+metrics are being created it is hard to add a new metric stability field to all
+of them since they are independently created in many different places
+
+- Complex to collect all existing metrics and alerts for automatic documentation
+
+- High levels of intertwined monitoring and business logic code
+
+### Goals
+
+- Decouple monitoring logic from business logic
+- Encapsulate the monitoring best-practices and the common patterns into a
+library and have it as a dependency for all KubeVirt components
+- Keep monitoring code and utilities easy to maintain and evolve
+- Document how to use the monitoring library for adding new metrics and rules
+- Have a structure and tools to accurately and easily generate monitoring
+documentation, lint metrics and alerts, define allow-list, deny-list and
+opt-in metrics and other future features without having to change the code in
+multiple places
+
+### Non-Goals
+
+- Update any metrics or alerts
+- Apply these ideas to events in this phase
+- Create a new component to run in the cluster alongside other components to
+collect metrics and create alerts
+
+## How
+
+### General relationship diagram:
+
+Each component will use a generic package for operator observability,
+abstracting all common operations between the components (such as registering
+them in `controller-runtime` metrics `Registry`), deduplicating code, and
+providing a way simplify the definition of metrics, recording rules, and alerts.
+
+The `docs/` package will list the metrics, alerts, and recording rules and
+generate the documentation files using the operator observability docs
+utilities.
+
+### Utilities
+
+The operator observability package will also have utilities to:
+- Generate the documentation for metrics, alerts, and recording rules
+- Lint metrics and recording rules names
+- Validate alerts configuration
+- Generate resource definitions PrometheusRule, ServiceMonitor, etc
+- Define allow-list, deny-list, and opt-in metrics
+
+## Action Plan
+
+### Decouple monitoring logic from business logic
+
+1) Move the monitoring logic from the business logic code to a dedicated package
+in each component repository
+
+### Use a generic operator-observability package
+
+1) Create the operator-observability package with the generic monitoring code
+   and utilities
+
+2) Document how to use the operator-observability package for adding new
+   metrics, alerts, and recording rules
+
+3) Update the monitoring code in each component to use the 
+   operator-observability package


design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,84 @@
...
+- Have a structure and tools to accurately generate monitoring documentation,
+lint metrics and alerts, and define allow-list, deny-list and opt-in metrics
+
+### Non-Goals
+
+- Update any metrics or alerts
+- Apply these ideas to events in this phase
+- Create a new component to handle all project monitoring
+
+## How
```
EdDev, line 48: Can you please add a diagram that shows the relation between:
- The central monitoring tooling package (i.e. a lib I guess) and the individual project that consumes it.
- The monitoring package in each project and how it interacts with the project itself (to collect the data). For example, what is the dependency direction.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,84 @@
...
+
+- Create a new separate “monitoring” package, in the "internal" folder,
```
EdDev, line 50: Not all projects are using the `internal` folder/s.
But if the suggestion here is to do so, it would be useful to explain why it helps and how it serves the needs listed above.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,84 @@
...
+- Update the monitoring documentation generator
+
+- Add unit tests to check metrics are being added to the Prometheus registry and
+PrometheusRule contains all alerts
+
+![code struct](struct.png)
+
+### Example implementation in [operator-sdk](https://github.com/operator-framework/operator-sdk) example operator
+
+Monitoring commit:
```
EdDev, line 66: I hope that with a diagram of what is expected, it will be easier to see how this works.

Perhaps you can consider a tutorial or workshop to promote this initiative.
I would be interested to attend one.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,84 @@
...
+- Refactor the implementation of the metrics in KubeVirt
+- Implement the metrics changes in other components
+- Refactor the implementation of the recording rules and alerts in KubeVirt
+- Implement the recording rules and alerts changes in other components
+- Add linter to the library package
+- Implement allow-list, deny-list and opt-in metrics logic
+
+## Open Questions
+
+- Should we add a mechanism to prevent developers to create and register metrics
```
EdDev, line 83: This depends on the build system.
But if the structure is clear, contributors will likely add it to the place it makes sense. Most will just follow the current norm, unless they think it is totally wrong.
I would worry about this only if there is proof that it is a problem, and not before.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,84 @@
...
+## Action Plan
+
+- Refactor the implementation of the metrics in KubeVirt
+- Implement the metrics changes in other components
+- Refactor the implementation of the recording rules and alerts in KubeVirt
+- Implement the recording rules and alerts changes in other components
+- Add linter to the library package
+- Implement allow-list, deny-list and opt-in metrics logic
+
+## Open Questions
```
EdDev, line 81: Who is suppose to maintain the monitoring subpackage in each project?
What happens today and are you looking to change it with this change proposal?

EdDev (commented):
Very nice initiative, thanks!

iholder101:
Thanks @machadovilaca! Separating monitoring logic from business logic is very important! great to see this being pushed forward. :+1: 

----

I feel like further explanation is needed here. Specifically, I'd like to understand:
What is the difference between the current situation and what you're proposing and how does the new state helps to reach the goals you mentioned, specifically to help separate monitoring from business logic?

IOW, is this change trying to solely achieve a uniformity between the different repositories, or does it also actually help separate monitoring from business logic? I'm asking since even now we have a dedicated `/monitoring` directory, which sounds pretty similar to what you suggest here.

---------

In addition, in [the proposal](https://github.com/kubevirt/community/pull/189) I've created once, I've thought of this problem at the opposite direction. Instead aiming for every repo to align with a certain structure, I think it's reasonable for every repo to adopt some kind of a declarative configuration that would describe which fields need to be monitored and how. I'm not sure how feasible this direction really is, but my point is to remove the burden from the developers to automation as much as possible, especially because I think that a lot of repositories use monitoring pretty similarly (e.g. track X and Y fields, alert when X is bigger than 123, etc).

I'm not experienced enough with the monitoring field to know if that's feasible and how, but I would like to hear your perspective regarding that direction.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
machadovilaca: I added the diagram, please check if it helps answer these questions. If needed I can try to add more info/detail

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
machadovilaca: Per [Golang project layout standards](https://github.com/golang-standards/project-layout#internal), I think it makes sense to use `internal/` to hold the application-specific metrics and rules code. The most important thing is that every project uses the same directories. But whether it is `internal/monitoring`, `pkg/monitoring` or even only `monitoring/` is up for discussion

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
machadovilaca: Most repositories already have someone who takes care of each component's monitoring aspects. But, for example, @assafad and I do a lot of monitoring work in which we need to edit and update all repositories, and I find it complex to keep a mental model of how every project does monitoring and I need to keep context switching.

To sum up, I would say each component should keep its monitoring owners, but we aim to simplify and uniformize the experience through all of them, and allow everyone to more easily contribute to monitoring in all components.

machadovilaca:
@iholder101 First of all thank you for your review.

And I agree with you. I think a declarative configuration for metrics and rules is a solution I would like to have one day. Unfortunately, for now, I think that goal would be way down the road since, in my opinion, we have increasingly more technical debt and it is becoming more difficult to update and 'evolve' monitoring features.

In this stage, my goal is to, as you said, align all repositories with a certain structure, have a package that all repositories import and use (reducing code duplication and making it easier to add new features for monitoring in the future), and cleanup business logic code.

This way, in the future, I think it would be easier to update the library package to parse and registry metrics and rules from declarative configurations, move the collection logic of external resources to external repositories, and even have an external controller/operator if needed/wanted.

---

Related to

> IOW, is this change trying to solely achieve uniformity between the different repositories, or does it also actually help separate monitoring from business logic? I'm asking since even now we have a dedicated /monitoring directory, which sounds pretty similar to what you suggest here.

Yes, KubeVirt already has a monitoring directory, some other components like HCO, also have it. But I still think there is still a lot of work to be done. Besides the components creating and registering metrics in very different ways, we see that not all metrics are registered there (for example [kubevirt_virt_controller_leading](https://github.com/kubevirt/kubevirt/blob/64a7cd17b60456364641ed036a9dabdb3a7727dc/pkg/virt-controller/watch/application.go#L122)),. We see that if we want to follow [Kubernetes Metrics Stability framework](https://kubernetes.io/blog/2021/04/23/kubernetes-release-1.21-metrics-stability-ga/) in kubevirt/kubevirt alone it would take a lot of manual repetitive work. We see that the metric documentation generator misses a lot of metrics.

I think this is hard work, but necessary to allow us to continuously improve our monitoring systems, and important if we ever want to reach some of the ideas you also had.

machadovilaca:
@enp0s3 Didn't test it, but since we are working with operators and `controller-runtime` we have one limitation. `controller-runtime` [creates a new Registry](https://github.com/kubernetes-sigs/controller-runtime/blob/73949d01fbbebeb5e2ce20f7e3b014832237186f/pkg/metrics/registry.go#L30) which usually does not work with these packages, the only workaround I was able to do was override the `controller-runtime` registry, but we would lose [all default exported metrics](https://book.kubebuilder.io/reference/metrics-reference.html). And `go-belt` [is managing its own repository](https://github.com/facebookincubator/go-belt/blob/fd57c819c9ec3aa76fdd5de385ba21d382e0ba05/tool/experimental/metrics/implementation/prometheus/metrics.go#L187). 

At the time, I found this issue investigating a library that, personally, I think was even more interesting than `go-belt`, [Kubernetes component base for metrics](https://github.com/kubernetes/component-base/tree/master/metrics), which is used by most of the Kubernetes components, but unfortunately has these compatibility issues with `sigs.k8s.io/controller-runtime/pkg/metrics`. They already have support for all metric types, stability levels, and also tracing and logs.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,90 @@
...
+metrics are being created it is hard to add a new metric stability field to all
+of them since they are independently created in many different places
+
+- Complex to collect all existing metrics and alerts for automatic documentation
+
+- High levels of intertwined monitoring and business logic code
+
+### Goals
+
+- Have a centralized place for the creation of monitoring resources
```
enp0s3, line 35: I would add `Ease the maintenance of monitoring code stability` as the justification of your action plan to add a linter.

design-proposals/monitoring-code-refactor/relationship.png
enp0s3, line 1: Can you please also add to the legend the different kinds of the colored squares? what is the meaning of each square color?

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,90 @@
...
+- Have a code struct that is easy to maintain and evolve
+- Move closer to the Kubernetes metric implementation style
+- Have a structure and tools to accurately generate monitoring documentation,
+lint metrics and alerts, and define allow-list, deny-list and opt-in metrics
+
+### Non-Goals
+
+- Update any metrics or alerts
+- Apply these ideas to events in this phase
+- Create a new component to handle all project monitoring
```
enp0s3, line 46: Is this means that the design doesn't cover refactoring of the way the operators and operands code are instrumented for monitoring the control-plane components?

iholder101:
> @iholder101 First of all thank you for your review.
> 
> And I agree with you. I think a declarative configuration for metrics and rules is a solution I would like to have one day. Unfortunately, for now, I think that goal would be way down the road since, in my opinion, we have increasingly more technical debt and it is becoming more difficult to update and 'evolve' monitoring features.
> 
> In this stage, my goal is to, as you said, align all repositories with a certain structure, have a package that all repositories import and use (reducing code duplication and making it easier to add new features for monitoring in the future), and cleanup business logic code.
> 
> This way, in the future, I think it would be easier to update the library package to parse and registry metrics and rules from declarative configurations, move the collection logic of external resources to external repositories, and even have an external controller/operator if needed/wanted.

Hey @machadovilaca!

Let's back down a bit and discuss what is the problem we really aim to solve here.

The basic problem here, as I see it, is that **every change regarding monitoring needs to happen at many different places**. For example, if another label needs to be added to a certain metric, there's a need to perform changes at many different repositories. This is both hard, takes a lot of time, requires going through code-reviews, costs CI resources and makes it mentally harder to decide to change anything.

The reality is that it's very hard maintaining monitoring, that's supposed to be aligned and uniform, between many different repositories. A **side-effect** of this approach is both that repositories diverge from one another over time, making it even harder to maintain as time goes by.

As I see it, this proposal suggests to invest a lot of effort in trying to solve the **side-effect** problem, by making conventions and linters that would try to reduce the divergence of repos from one another. My humble opinion is that we should invest, even if it would take longer, in a solution that solves the root cause of evil here, which is performing changes in many repos manually.

What concerns me the most in this approach:
1. This effort is not negligible at all. It demands once again to perform changes at many different places, which obviously includes code reviews, CI resources, etc.
2. I don't see how this can be enforced. Even with linters and conventions, it's likely that some of the repositories would diverge over time.
3. And this is the most important bit, we don't solve the root issue. All of this effort is invested just to keep the problem of making PR to many different repos for every change, but make it a bit less severe.

I think that a better approach going forward is to introduce some kind of a framework / library that repositories can import and have under `/vendor`. This way, if something needs to change, all needs to be done is to make the changes in **one repo**, then only **bump the version for many repos**. Even if it's harder to accomplish and would take more time, it's at least directed at solving the root problem. This would make a huge difference, would ensure uniformity between all repos, save time (for example w.r.t. code-reviews), CI resources, and would make it mentally easier for you to decide about future changes.

WDYT?

enp0s3:
@machadovilaca @iholder101 More or less what I was thinking of:

* Encapsulate the monitoring best-practice and the common patterns into a framework/library and put in in `/vendor`
* De-couple the monitroing bits from the main flow and put them into a consolidated place in each project, put `OWNERS` file under the directory thus making the monitoring team the maintainers of the that piece of code.
* Enforce stability by automation as much as possible, leaving the non-trivial parts for human review, thus implementing linters.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,112 @@
...
+`Registry`), deduplicating code, and providing a way to define the metrics,
+recording rules , and alerts in a declarative manner.
+
+### Relationship diagram for each component monitoring code:
+
+<img src="component_monitoring_relationship.png" alt="relationship diagram" width="800"/>
+
+Each component package in `kubevirt/monitoring`, will have a `metrics/` package
+with the metrics declarations, the methods to update/consult the metrics values
+and the collectors logic. The `rules/` package will also have the declarative
```
iholder101, line 90: Please correct me if I'm wrong here, but it seems like the `metrics/` directory will have shared logic between all components. Do we need a different directory for each component? Am I missing something here?

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,112 @@
...
+configuration of the resources, in this case, alerts and recording rules.
```
iholder101, line 91: Can you show an example of this declarative configuration? I think it would be much easier to picture this solution that way

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,112 @@
...
+
+## How
+
+### General relationship diagram:
+
+<img src="relationship.png" alt="relationship diagram" width="800"/>
+
+With this refactor, we intend to move out monitoring logic from all the KubeVirt
+components' business logic code. As we see in the above diagram, each KubeVirt
+component will now import a package in the `kubevirt/monitoring` repository.
```
iholder101, line 72: Can you clarify: how does a Kubevirt component uses the package it imports from `kubevirt/monitoring`? It it being used to declaratively define the metrics/alerts it wants to expose? Is this declarative definition in-code, or with a file e.g. yaml?

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,112 @@
...
+- Decouple monitoring logic from business logic
+- Have a centralized place for the configuration of monitoring resources
+- Encapsulate the monitoring best-practices and the common patterns into a
+library and have it as a dependency for all KubeVirt components
+- Have a declarative way to define metrics and alerts
+- Keep monitoring code and utilities easy to maintain and evolve
+- Document how to use the monitoring library for adding new metrics and rules
+- Have a structure and tools to accurately and easily generate monitoring
+documentation, lint metrics and alerts, define allow-list, deny-list and
+opt-in metrics and other future features without having to change the code in
```
iholder101, line 54: Can you explain how it would be possible to opt in/out of metrics/alerts?

iholder101 (commented):
Hey @machadovilaca!

Thanks a lot for shifting this proposal into a new much more clean and general approach, this already looks much much better!

I feel like I need some more examples to fully grasp what you're suggesting here :) so forgive me if I'm totally wrong in my comments below. It would be great if you can add more examples and clarifications to make the idea clearer.

One general thought: you describe one of the pitfalls of the current approach as:

> It is hard to keep a mental model of how monitoring is implemented in all the different repositories and is difficult to maintain all of them in sync with new features

I wonder - what is the gap between having a kubevirt-specific solution and having a general enough solution so it could be used even outside of Kubevirt? As I see it (and maybe I'm wrong here), every operator could essentially declaratively express what are the metrics and alerts it would like to expose. As a silly example, let's say that there's a k8s operator out there that creates a CRD called `Human`, and wishes to expose the age of this human as a metric. Ideally, we could declaratively express that `spec.age` needs to be exposed as a metric, and if `spec.age > 120` we wish to fire an alert. Perhaps the solution that you're suggesting here can be even more general to support it as well.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
machadovilaca: the shared logic will be in the `operator-observability` package/module, and each `metrics/` will contain the specific definitions for each component resources (using `operator-observability`).

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
machadovilaca: added

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
machadovilaca: When moving the monitoring code to `kubevirt/monitoring`, it would already be a declarative definition of metrics and other resources

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
machadovilaca: the `operator-observability` package will already provide the idea of a `group` of metrics which will help with this.

We will have 2 flags:

opt-in: a list of groups and/or metrics that should be the only ones registered in the Prometheus Registry
opt-out: a list of groups and/or metrics that should be excluded from the registration in the Prometheus Registry (while everything else is registered)

machadovilaca:
> I wonder - what is the gap between having a kubevirt-specific solution and having a general enough solution so it could be used even outside of Kubevirt? As I see it (and maybe I'm wrong here), every operator could essentially declaratively express what are the metrics and alerts it would like to expose. As a silly example, let's say that there's a k8s operator out there that creates a CRD called Human, and wishes to expose the age of this human as a metric. Ideally, we could declaratively express that spec.age needs to be exposed as a metric, and if spec.age > 120 we wish to fire an alert. Perhaps the solution that you're suggesting here can be even more general to support it as well.

the `operator-observability` package will be the place where we will move all the generic code. That should be usable for any operator. 

As an initial POC: https://github.com/machadovilaca/operator-observability

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
iholder101: Got you! Thanks for clarifying

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
iholder101: Thanks! Much clearer now!

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
iholder101: In the form of YAML files IIUC, right?

iholder101:
> > I wonder - what is the gap between having a kubevirt-specific solution and having a general enough solution so it could be used even outside of Kubevirt? As I see it (and maybe I'm wrong here), every operator could essentially declaratively express what are the metrics and alerts it would like to expose. As a silly example, let's say that there's a k8s operator out there that creates a CRD called Human, and wishes to expose the age of this human as a metric. Ideally, we could declaratively express that spec.age needs to be exposed as a metric, and if spec.age > 120 we wish to fire an alert. Perhaps the solution that you're suggesting here can be even more general to support it as well.
> 
> the `operator-observability` package will be the place where we will move all the generic code. That should be usable for any operator.
> 
> As an initial POC: https://github.com/machadovilaca/operator-observability

That's AWESOME to hear. Thank you. I think this is the right direction going forward!

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,214 @@
...
+          stability: stable
+          labels:
+            - name: another_label
+              value: only_for_this_metric
+
+      recording_rules:
+        - name: example_operator_number_of_pods
+          help: Number of example operator pods in the cluster
+          type: gauge
+          expr: sum(up{namespace='$NAMESPACE', pod=~'example-operator-.*'}) or vector(0)
```
iholder101, line 129: I think that we can take this a step forward by presenting common promQL generators that would be declarative and help the user define these without having to deal with promQL itself.

Some thoughts that come into mind are, for metrics:
`expr: ObjectField("spec.age")`
`expr: NumberOfInstances("my-namespace", "virt-api-*")`
`expr: ReadyPods("my-namespace", "virt-api-*")`

for alerts:
`expr: ObjectField("spec.age").IsNumerically(">", 123)`
`expr: DeploymentInstances("deploymentName").IsZero()`

As I see it, `operator-availability` could have pre-made promQL generators (or however we want to call them) that are relevant to most operators out there. Then, every operator (e.g. Kubevirt) would be able to expand them with generators specific to their scope. As an example, for Kubevirt, we can introduce `NodeWithKVM()`, `LiveMigration()`, etc.

I think that with these in place this repo might be very appealing to many other operators as it eases the monitoring effort into a level that it's accessible by developers that aren't familiar with PromQL at all.

WDYT?

sradco: This doesn't make sense to me. I think we should keep the PromQL language.
If someone is adding a metric they should know PromQL. This is a must.

iholder101: Can you provide a reasoning for your opinion?

Why is it wrong to wrap very basic operations with functions that would make it easier? Why does a user that wants to add a straight-forward metric needs to struggle with promQL code? Why do we need to duplicate the PromQL logic between many different metrics?

I just don't understand what your concerns are / what's the value of requiring to write explicit promQL code.

akrejcir:
This looks great, thanks!

I have a question about YAML definitions: Have you considered using golang `structs` directly for the declarative definitions? It would make sense, if the yaml files will only be stored in the codebase and then parsed to golang `structs` at runtime.

machadovilaca:
@akrejcir I think in the end, we will somewhat support both approaches. Having the YAML option, I think it allows users to more easily add their own custom metrics, recording rules and alerts, without having to touch the code.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,214 @@
...
+
+- High levels of intertwined monitoring and business logic code
+
+### Goals
+
+- Decouple monitoring logic from business logic
+- Have a centralized place for the configuration of monitoring resources
+- Encapsulate the monitoring best-practices and the common patterns into a
+library and have it as a dependency for all KubeVirt components
+- Have a declarative way to define metrics and alerts
```
RamLavi, line 49: The current way is already declarative, is it not?

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,214 @@
...
+all KubeVirt components. The goal is to have a centralized place for the
+creation of monitoring resources, a consistent monitoring package between all
+projects, a code struct that is easy to maintain and evolve, while moving closer
+to the Kubernetes metric implementation style.
+
+## Why
+
+It’s important to refactor the monitoring code to separate monitoring logic from
+business logic. It will make the codebase more modular, readable, and
+maintainable. It will reduce complexity, reduce the risk of introducing errors,
```
RamLavi, line 17: I don't understand how this proposal is offering to reduce errors, as you reason here.
If I create an alert, that is basically garbage, how this new object validate it?

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,214 @@
...
+and the collectors logic. The `rules/` package will also have the declarative
+configuration of the resources, in this case, alerts and recording rules.
+
+Finally, the `docs/` package will list the metrics, alerts, and recording rules
+and generate the documentation files using the operator observability docs
+utilities.
+
+### Declarative configuration
+
+The declarative configuration will be done using a YAML files. The YAML files
```
RamLavi, line 99: typo: YAML files (no `a`)

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,214 @@
...
+## Action Plan
+
+### Phase 1 - Decouple monitoring logic from business logic
+
+1) Move the monitoring logic from the business logic code to a dedicated package
+in each component repository
+
+### Phase 2 - Use a generic operator-observability package
+
+1) Create the operator-observability package with the generic monitoring code
```
RamLavi, line 191: Can you specify which utilities?

RamLavi (commented):
Added questions. 
Plus, I assume that components that can also run independently to kubevirt (like CNAO, KMP, etc..) to need to deploy the controller that will do the logic of talking to prometheus, right?

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
machadovilaca: yes

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
machadovilaca: not really, we want to abstract all the calls to `prometheus.MustRegister`, `prometheus.NewCounter`,  `prometheus.NewCounterVec`, and so on, all of that should be handled by the common lib

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
machadovilaca: `the risk of introducing errors` here refers to the fact that now that the KubeVirt is in a more advanced stage and we are seeing a considerable increase in monitoring features added across all repositories, having business logic intertwined with monitoring logic, makes it harder to contribute to either one and changes more prone to introduce errors

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
machadovilaca: I will update the text with these examples, but to name a few, documentation generators, metric and recording rules name validators, alert configuration validators, building `PrometheusRule` resource definition, ...

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
machadovilaca: will fix

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
RamLavi: Can you give me an example of such a case where monitoring logic interfered with business logic?

machadovilaca: https://github.com/kubevirt/kubevirt/pull/7465/files#diff-1baf192bb9c4560d7d40f008b7ba5c680ebce669b8f1b7639445abd3e7269688

that we ended up removing later

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,223 @@
...
+
+2) Document how to use the operator-observability package for adding new
+   metrics, alerts, and recording rules
+
+3) Update the monitoring code in each component to use the
+operator-observability package
+
+### Phase 3 - Have a declarative way to define metrics and alerts
+
+1) Update the operator-observability package to have a declarative way to
```
sradco, line 211: I think we should keep the Prometheus way of creating the metrics and alert and not create another layer on top of it.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,223 @@
...
+define metrics and alerts
+
+2) Update the monitoring code in each component to use the declarative way to
+define metrics and alerts
+
+### Phase 4 - Move existing monitoring code to kubevirt/monitoring
```
sradco, line 217: I think this can bring complexity regarding versions dependency, like we have now with kube-state-metrics and that its important that we keep the metrics and alerts in each sub operators separately.
Please outline pros and cons for this step. Im not convinced.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
sradco: This would be a great addition, similar to how we use the metrics linter.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
sradco: The overhead of maintaining such a layer is not worth the effort IMHO

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
sradco: Its like adding a layer over SQL language. It doesn't make sense.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
sradco: I think the idea is, for metrics, to add the metrics to a list where each item includes all the required details and the list would be used for defining the based on its typpe, registering it, adding it to the auto docs generator and so on. Am I correct?

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
enp0s3: @sradco Hi. Can you please elaborate on the versions dependency issue you had with kube-state-metrics?

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
enp0s3: Hi. I think that creating a language that is rendered to a query language is a bit overkill here, although if someone manage to create such thing and it plays nicer than promQL itself it should be suggested to prometheus folks as well :)
But instead, having some subset of Go funcs that will return common PromQL expressions could be wonderful.

enp0s3: > I think that we can take this a step forward by presenting common promQL generators that would be declarative and help the user define these without having to deal with promQL itself.
> 
> Some thoughts that come into mind are, for metrics: `expr: ObjectField("spec.age")` `expr: NumberOfInstances("my-namespace", "virt-api-*")` `expr: ReadyPods("my-namespace", "virt-api-*")`
> 
> for alerts: `expr: ObjectField("spec.age").IsNumerically(">", 123)` `expr: DeploymentInstances("deploymentName").IsZero()`
> 
> As I see it, `operator-availability` could have pre-made promQL generators (or however we want to call them) that are relevant to most operators out there. Then, every operator (e.g. Kubevirt) would be able to expand them with generators specific to their scope. As an example, for Kubevirt, we can introduce `NodeWithKVM()`, `LiveMigration()`, etc.
> 
> I think that with these in place this repo might be very appealing to many other operators as it eases the monitoring effort into a level that it's accessible by developers that aren't familiar with PromQL at all.
> 
> WDYT?

I don't think its needed for metrics instrumentation. The way to write metrics using Prometheus Go client is pretty straight-forward.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
enp0s3: @sradco @iholder101 I think the approach here is different for metrics and for alerts. With regards to metrics, are we talking about creating a code generator here?

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
sradco: When you create a query you need to mine the filtering of the metric (namespace/name/pod/etc.) , what you sum/avg/max/min by and the metric type (Gauge/Counter/Histogram/Summery) will require different functions for doing the same thing.
For example: 
sum(rate(metric_name[time period])) (for counter) vs 
sum_over_time(metric_name[time period]) (for gauge) or perhaps increase(for counters).
We should use the plain promQL and not add extra complexity to it.

iholder101: Obviously I didn't mean to create another layer of promQL or extend the language, but rather introduce helper functions for very frequent operations. I have a feeling that many metrics would be very similar to one another (not only just in Kubevirt, but in many other repos as well) and if that's the case I don't see a reason why not to provide a way of simplification and avoiding code duplication.

I think it's somewhat similar to how we're using gomega. We could have written:
```go
if value == nil {
    return err
}
```

But since it's a very common operation, we rather use something like:
```go
Expect(value).NotTo(BeNil())
```

I wouldn't say that this is an extra layer to Go, just helper functions for frequent operations.

BTW, @machadovilaca's example [here](https://github.com/machadovilaca/operator-observability/blob/main/examples/metrics/custom_resource_collector.go#L16) is somewhat similar. There, a function is being called to count the number of k8s objects in a generic way that doesn't require any promQL knowledge.

I think the advantages are:
1. Simplicity
2. Less error-prone
3. Easier to add straight-forward metrics (which are pretty common) by people who aren't comfortable with PromQL.

I'm also not convinced by the argument that it would be a burden to maintain, since again I'm talking about the most common and straight forward promQL code anyways. The argument for being more error-prone is not compelling to me either, since it's much easier to make a typo in a promQL language IMO.

iholder101: In any case, this discussion should block @machadovilaca from merging this PR.
We can asses this together and can tackle this in the future. Perhaps the right place to discuss this in within an issue in the operator-observability repo

sradco: This should not be a goal for this feature.
If you (In general) think of things that can help you can create a PR for them.
But it should not be part of the scope for the code refactoring which aims to separate the monitoring code from the core operator code.
We need to scope the effort for this feature. This is a nice to have improvement at should be separated out of this feature please.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,223 @@
...
+business logic. It will make the codebase more modular, readable, and
+maintainable. It will reduce complexity, reduce the risk of introducing errors,
+and allow us to test each component independently.
+
+Right now, every change regarding monitoring needs to happen at many different
+places. For example, when adding the linter for metrics names, it was necessary
+to go to all KubeVirt components repositories and implement the linter
+validation in each. This is demanding, very time-consuming, requires going
+through multiple duplicated code reviews, uses many CI resources, and makes it
+mentally harder to decide to change anything.
```
sradco, line 25: I disagree with the statement that it "makes it mentally harder to decide to change anything."
We never thought about this in this way.

I still think that there are significant advantages for having the monitoring tied to the specific operator that its related to.

We have a hard time in the past with other operators that we struggled to understand where their monitoring come from and how they are calculated, when they were not in the same repository.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,223 @@
...
+
+We are facing multiple challenges with the current solution:
+
+- It is hard to keep a mental model of how monitoring is implemented in all the
+different repositories and is difficult to maintain all of them in sync with new
+features
+
+- Hard to add new features to metrics and alerts, for example, because of how
+metrics are being created it is hard to add a new metric stability field to all
+of them since they are independently created in many different places
```
sradco, line 37: I don't understand this statement. Why is it hard? 
Why can't we add the same logic to each operator?

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
iholder101: @enp0s3 +1

We need more details here w.r.t. the cons.
In general, I think it's very valuable in many ways:
1. We can ensure that monitoring code does not change business logic
2. PRs would be much simpler to review and add when the monitoring and kubevirt-core code would be decoupled. The monitoring experts could own their repo including reviewing, approving and be responsible for the monitoring code in an independent way.
3. CI could look entirely different on both Repos. For example, in the monitoring repo we can add monitoring linters and specific monitoring tests.
4. From the Kubevirt perspective: replace the burden of maintaining monitoring logic and instead focus on testing the monitoring package with Kubevirt.
5. Easier to fork and customize
6. A central place to add docs, bug reports, issues for enhancement requests, monitoring discussions, etc

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
sradco: I think the emphasis here is that :
```
Since metrics are being created independently in many different places, its hard to get the full list of metrics that the operator is exposing and this is causing issues. For example:
- We miss metrics when doing automatic documentation
- We don't test all metrics with the needed tests
- We can apply additional feature to the metrics that we are not aware of
```

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
iholder101: > I disagree with the statement that it "makes it mentally harder to decide to change anything."

Don't you agree that adding code to multiple repositories, having to code-review and get feedback from multiple of them, while having to align the logic between them brings complexities and intimidates from introducing more complex monitoring logic?

I think what @machadovilaca wrote there makes a lot of sense, and my past experience is similar.

> I still think that there are significant advantages for having the monitoring tied to the specific operator that its 
related to.

Can you outline the advantages?

> We have a hard time in the past with other operators that we struggled to understand where their monitoring come from and how they are calculated, when they were not in the same repository.

I think entirely the opposite.
When the monitoring code is part of the core operator in many different repositories it's very hard to reason about it. When the code resides at a centralized repository it's easy to find and understand. We can also make sure it's well-documented.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
```diff
@@ -0,0 +1,214 @@
...
+It should be possible to:
+
+- create a single file with all groups of metrics, alerts, and recording rules
+
+- create multiple files, each with a specific set of groups. With each group
+containing its metrics, alerts, and recording rules
+
+- create multiple files, each with multiple groups with only one type of
+resource (metrics, alerts, or recording rules). For example, one file with all
+metrics, another with all alerts, and another with all recording rules.
```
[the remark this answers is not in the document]
sradco: I agree, This would be a great improvement.
We need to enable an allow/deny list .

But, I dont think we should have metrics, alerts, and recording rules in a single file.
And we should have the structure for having an allow/deny list fro what we expose.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
iholder101: Could you please provide your thought on the following scenario:
- Someone is trying to add monitoring code to repos A, B and C.
- In repo A, the community accepts the changes.
- In repo B, the community accepts the changes in general, but demands several core changes.
- In repo C, there's a significant push-back regarding the code and they don't want to adopt it without significant changes.

sradco: Please try to think of it differently.
When you take out the monitoring from the repo, then you also take away the maintainers and the repo community control over it.
Can you really prevent each repo from adding new metrics and alerts that are asked for? You may end up with having monitoring for the same repo in several places.
Yes, Adding functionality that we need to several repos has its challenges, but if its good, like the linter, the docs generator, tests and so on, the community will accept it. 
Another point is that getting the feedback and requests from the different operators makes the code much better at the end.

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
sradco: > @enp0s3 +1
> 
> We need more details here w.r.t. the cons. In general, I think it's very valuable in many ways:
> 
> 1. We can ensure that monitoring code does not change business logic

Having it in a separate package inside the repo should allow that

> 2. PRs would be much simpler to review and add when the monitoring and kubevirt-core code would be decoupled. The monitoring experts could own their repo including reviewing, approving and be responsible for the monitoring code in an independent way.

Same, if its in a separate package inside the repo

> 3. CI could look entirely different on both Repos. For example, in the monitoring repo we can add monitoring linters and specific monitoring tests.

I don't think there would be a big advantage there. We need to test them together, see that alerts are fires, metrics are propagated, etc. This can actually create duplicated tests in my opinion.

> 4. From the Kubevirt perspective: replace the burden of maintaining monitoring logic and instead focus on testing the monitoring package with Kubevirt.

I dont see how this cant be separated completely, since there are metrics/logs that are gathers on the VMIs that cant be collected from an external operator.
If all the monitoring code is separated to a different package inside the repo, I don't see why it would still be a burden. The package can have its own maintainers list.

> 5. Easier to fork and customize

If you are using only Kubevirt, why would you want a repo that includes monitoring for the other operators?
> 6. A central place to add docs, bug reports, issues for enhancement requests, monitoring discussions, etc

This would be better scoped when its tied to the specific operator.
It would make it harder to track each individual bug/RFE for the different operators and many times the developers that will need to work on it are the the core operator developers and not the monitoring team developers.

sradco: Im all for an observability library like Joao suggested, but against separating out the monitoring code from each of the operators to a separate operator.

RamLavi, line 217: I find it would be complex to implement and maintain having the alerts/metric apart from the repo using them, especially across so many kubevirt products. 
But maybe I would see things differently if you do a POC about it with the more specific design details.

sradco:
We agreed to split this proposal to two proposal:

1. Separating the monitoring code to a separated package within the same repo and adding an operator observability library for cross operators tools and code.
Goal: Clear place for adding/updating monitoring related code, easier maintenance and better control, separated owners file, align the common code across the operators.

2. Declarative code for defining metrics and alerts and moving existing monitoring code to kubevirt/monitoring
Goal: Additional improvements. Nice to have at this point. Will need to asses this after the first part.

enp0s3:
> We agreed to split this proposal to two proposal:
> 
> 1. Separating the monitoring code to a separated package within the same repo and adding an operator observability library for cross operators tools and code.
>    Goal: Clear place for adding/updating monitoring related code, easier maintenance and better control, separated owners file, align the common code and tests across the operators.
> 2. Declarative code for defining metrics and alerts and moving existing monitoring code to kubevirt/monitoring
>    Goal: Additional improvements. Nice to have at this point. Will need to asses this after the first part.

Sounds good to me.

machadovilaca:
> We agreed to split this proposal to two proposal:
> 
> 1. Separating the monitoring code to a separated package within the same repo and adding an operator observability library for cross operators tools and code.
>    Goal: Clear place for adding/updating monitoring related code, easier maintenance and better control, separated owners file, align the common code and tests across the operators.
> 2. Declarative code for defining metrics and alerts and moving existing monitoring code to kubevirt/monitoring
>    Goal: Additional improvements. Nice to have at this point. Will need to asses this after the first part.

I created a separate Pull request for later review with "Separating the monitoring code" and " Declarative code for defining metrics and alerts" -> https://github.com/kubevirt/community/pull/234

Please take a new look and review the current PR independently. 
@EdDev @sradco @RamLavi @enp0s3 @iholder101

design-proposals/monitoring-code-refactor/monitoring-code-refactor.md
RamLavi: OK thanks

RamLavi (commented):
> ### Use a generic operator-observability package

Question: Does this proposal still want to create a different repo/library? if not, how do you plan to expand these tools outside of the kubevirt repo? Or does this proposal mean to focus on kubevirt?

machadovilaca:
> > ### Use a generic operator-observability package
> 
> Question: Does this proposal still want to create a different repo/library? if not, how do you plan to expand these tools outside of the kubevirt repo? Or does this proposal mean to focus on kubevirt?

@RamLavi we still want to create a new library with common code for all components, see goal:

"- Encapsulate the monitoring best-practices and the common patterns into a
library and have it as a dependency for all KubeVirt components"

RamLavi (approved):
Sounds like a good proposal!

/lgtm

vladikr:
/approve

Thanks. I had a hard time understanding which code is going to be moved out exactly. Thank @machadovilaca for the explanation!

iholder101:
/lgtm
Well done!

PR #224, opened 2023-06-28
Add Helm charts proposal

atanasdinov:
This design proposal addresses the Helm chart installation method alternative. For reference: https://github.com/kubevirt/kubevirt/issues/8347

### DIFF
--- a/design-proposals/helm-chart.md
+++ b/design-proposals/helm-chart.md
@@ -169,6 +169,11 @@ The CDI project has the same `CRD-Operator-CR` setup so all of the above applies
 Note that the file structures above are purely for illustrative purposes.
 Different Kubernetes resources (e.g. `rbac`) could be extracted as separate templates (files) in order to keep the charts cleaner and easier to navigate.
 
+### Decision
+
+After careful evaluation of all the different options by the community, [Approach 2](#approach-2-one-crd-chart-and-one-cr--operator-chart) 
+is by far the most favoured one taking into account popularity, maintainability, as well as extensibility.
+
 ## Chart Management
 
 The second problem which needs to be addressed is how Helm charts are going to be created and modified.
@@ -204,9 +209,13 @@ but it should be stable and both manifests and charts would be released simultan
 however, we need to keep in mind that the complete manifests are not version controlled.
 This would also mean that updated Helm charts are released separately from the manifests.
 
-Unfortunately, I'm still very new to KubeVirt so feedback in this section would be more than welcome.
-
 I'm not familiar with CDI's current manifest creation flow yet, but I *believe* it is similar.
+
+### Decision
+
+[Approach 2](#approach-2-manually-curate-a-helm-chart) is the preferred option by the community.
+The Helm charts will be created based on the KubeVirt (and CDI) manifests that are published
+with each KubeVirt (and CDI) release and will be hosted in a separate GitHub repository (e.g. `kubevirt/charts`).
 
 ## Helm repository
 
@@ -231,17 +240,35 @@ We should be able to use Quay OCI considering all other artefacts are already st
 
 ## API Examples
 
-Working with Helm charts is fairly straightforward. Users should be able to manage the applications by:
+Working with Helm charts is fairly straightforward. Users should be able to manage the applications via Helm CLI.
+
+### Add Chart Repository 
 
 ```shell
 $ helm repo add kubevirt <repo-url>
+```
+
+Note that this is not necessary if we use OCI registry as opposed to a standard HTTP(s) repository.
+
+### Install Charts
+
+```shell
+$ helm install kubevirt-crd kubevirt/kubevirt-crd --version <version>
 $ helm install kubevirt kubevirt/kubevirt --version <version> \
       # additional configuration
+$ helm install cdi-crd kubevirt/cdi-crd --version <version>
 $ helm install cdi kubevirt/cdi --version <version> \
       # additional configuration
+```
+
+### Uninstall Charts
+
+```shell
 $ helm uninstall kubevirt
 $ helm uninstall cdi
 ```
+
+Note that it is a good practice to also uninstall the CRD charts (following the same example) in order not to have unused CRDs on the cluster.
 
 ## Scalability
 
--- a/design-proposals/helm-chart.md
+++ b/design-proposals/helm-chart.md
@@ -208,6 +208,27 @@ Unfortunately, I'm still very new to KubeVirt so feedback in this section would 
 
 I'm not familiar with CDI's current manifest creation flow yet, but I *believe* it is similar.
 
+## Helm repository
+
+Chart repository is a location where packaged charts can be stored and shared.
+
+We need to configure one and publish KubeVirt and CDI charts there.
+
+There are multiple options to achieve this. The ones worth considering are:
+
+- GitHub pages
+
+This is the most popular way of setting up repositories for GitHub based projects.
+GitHub can be configured to serve static content using simple configuration.
+[Example guide](https://helm.sh/docs/topics/chart_repository/#github-pages-example).
+
+- OCI registry
+
+The more interesting option is to use OCI registry as Helm repository.
+[This feature](https://helm.sh/docs/topics/registries/#using-an-oci-based-registry) has graduated to GA since Helm v3.8.0.
+We should be able to use Quay OCI considering all other artefacts are already stored there.
+[Example guide](https://cloud.redhat.com/blog/quay-oci-artifact-support-for-helm-charts)
+
 ## API Examples
 
 Working with Helm charts is fairly straightforward. Users should be able to manage the applications by:
@@ -245,5 +266,5 @@ The same tests should be performed for CDI.
 # Implementation Phases
 
 - Create a KubeVirt Helm chart
-- Configure a Helm repository (using GitHub pages is probably the best option)
+- Configure a Helm repository
 - Create a CDI Helm chart and add it to the same Helm repository
--- a/design-proposals/helm-chart.md
+++ b/design-proposals/helm-chart.md
@@ -1,0 +1,249 @@
+# Overview
+
+Provide and support Helm charts as an alternative installation method for KubeVirt and CDI.
+
+## Motivation
+
+Helm is one of the most popular choices when it comes to managing Kubernetes applications and their release lifecycle.
+Many software platforms prefer Helm charts for their configuration, automation and distribution capabilities.
+
+## Goals
+
+- Provide Helm charts for all upcoming versions of KubeVirt and CDI
+
+## Non Goals
+
+- Removing manifests installation method is out of scope.
+  Users will continue to have the option to deploy KubeVirt and CDI using the manifests provided at build time.
+
+## Definition of Users
+
+This feature is intended for administrators of Kubernetes clusters who are managing KubeVirt's (and CDI's) release lifecycles.
+
+## User Stories
+
+- As a cluster administrator, I'd want to install, upgrade and uninstall KubeVirt using Helm chart
+- As a cluster administrator, I'd want to install, upgrade and uninstall CDI using Helm chart
+
+## Repos
+
+* [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+* [kubevirt/containerized-data-importer](https://github.com/kubevirt/containerized-data-importer)
+
+# Design
+
+In order to establish a solution design we need to answer the following questions:
+* How are the Helm charts going to look like?
+* How will the Helm charts be managed?
+
+## Chart Design
+
+Both KubeVirt and CDI installations depend on CRD, CR and Operator Deployment resources.
+
+The current installation procedure involves creating the CRD and Operator Deployment first closely followed by creating the CR.
+
+Helm, however, has limitations when it comes to [handling CRDs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/).
+Upgrading and deleting is not currently supported (unless templated) and CRDs must be present before creating a CR.
+The CR is required in order for the operator to deploy the rest of KubeVirt's dependencies and applications.
+
+This leaves us with the following Helm chart layouts:
+
+### Approach 1: Single chart
+
+```text
+├── Chart.yaml
+├── crds
+│   └── kubevirt-crd.yaml
+├── templates
+│    ├── kubevirt-cr.yaml
+│    ├── kubevirt-hooks.yaml
+│    └── kubevirt-operator.yaml
+└── values.yaml
+```
+
+This method allows us to have all resources in the same chart.
+
+The `CustomResourceDefinition` is extracted out of `kubevirt-operator.yaml` under the special `/crds` directory which Helm will pick up and install first.
+Note that `kubevirt-crd.yaml` cannot be templated.
+
+The installation process for this approach is simple (`$ helm install kubevirt <repo-url>`), however,
+there are some obstacles when it comes to upgrade and uninstall procedures.
+
+Upgrading a CRD is not currently possible in Helm. Once installed, further upgrades will only detect that the CRD already exists and will ignore it.
+
+Uninstalling depends on the CR being removed before the Operator. The CRD removal is also not taken care of by Helm automatically.
+
+In order to resolve these issues, we can leverage [Helm's hooks mechanism](https://helm.sh/docs/topics/charts_hooks/).
+It allows us to intervene at certain points of the release lifecycle and perform custom operations.
+These operations will be defined in `kubevirt-hooks.yaml` and will consist of the following:
+- `pre-upgrade` hook manually applying the CRD manifest in order to perform modifications to it
+- `pre-delete` hook removing the CR first in order for the operator to gracefully uninstall KubeVirt's components before Helm removes it
+- `post-delete` hook removing the leftover CRD
+
+This approach was used in [SUSE Edge's initial Helm chart](https://github.com/suse-edge/charts/tree/main/charts/kubevirt/0.1.0).
+[PR](https://github.com/suse-edge/charts/pull/13) for additional context.
+
+### Approach 2: One CRD chart and one CR + Operator chart
+
+Perhaps the most common solution to tackle the CRD problem is to extract the CRD in a separate Helm chart.
+
+```text
+├── Chart.yaml
+├── templates
+│   └── kubevirt-crd.yaml
+```
+
+```text
+├── Chart.yaml
+├── templates
+│    ├── kubevirt-cr.yaml
+│    ├── kubevirt-hooks.yaml
+│    └── kubevirt-operator.yaml
+└── values.yaml
+```
+
+Here we will install the CRD chart first and the CR / Operator chart after that.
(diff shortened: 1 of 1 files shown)


design-proposals/helm-chart.md
```diff
@@ -0,0 +1,249 @@
...
+- Ensure all components are working as expected and the configurations have been properly updated
+- Uninstall KubeVirt chart
+- Ensure all components have been removed
+
+The same tests should be performed for CDI.
+
+# Implementation Phases
+
+- Create a KubeVirt Helm chart
+- Configure a Helm repository (using GitHub pages is probably the best option)
```
vasiliy-ul, line 248: I do not see it mentioned in the proposal, but Helm also supports charts deployment from OCI registries: https://helm.sh/docs/topics/registries/. I think it's also an interesting option to consider. Though, I do not know if there are limitations or potential compatibility issues from the perspective of quay.io. Quick googling suggests that it should work https://cloud.redhat.com/blog/quay-oci-artifact-support-for-helm-charts

vasiliy-ul (commented):
Hi @atanasdinov, great proposal. Thanks a lot for bringing it up here.

I had an initial look and here are my thoughts on the topic.

From the end-user perspective, I personally like the `approach #1`. It is simple and less error-prone (no chance for the user to uninstall the resources in wrong order or smth like that). However, the simplicity comes at the cost of additional logic within the hooks that needs to be maintained.

`approach #3` looks like a good alternative which basically replicates the existing flow with the manifests. It is well-known to the users already. Also, from the implementation and maintenance point of view, it does not seem to add any complexity. And in fact I think that by using the hooks it might be possible to e.g. prevent uninstallation of the operator if the CR is still present in the cluster. Thus enforcing proper uninstallation order. The only downside I see for now with this approach is related to potential inconsistency between operator/cr charts versions. When updating KubeVirt/CDI you usually do not need to update the CR, only the operator (and CRD). Therefore, after an update, the CR chart will stay at version N while the operator will move to N+1. Maybe this is in fact not a big issue at the end. But still maybe worth thinking about it... if it may cause some problems or confusions for the users.

Regarding `approach #2` I think there might be an issue if it always installs/updates the CR resource. Users may have some settings defined in the CR, and I assume those will be overwritten by an update.

Just to summarize: I am not sure about `#2` but `#1` and `#3` look fine.

design-proposals/helm-chart.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Approach 1: Generate Helm chart(s)
+
+Helm charts are in general not that different from plain Kubernetes manifests.
+It all depends on how much customization we'd allow the user but for the SUSE Edge initial chart we kept as little modifications as possible.
+
+We should be able to plug in the functionality to generate a `Chart.yaml`, `values.yaml` and possibly hooks and organize those in the necessary file structure.
+The charts can then be generated, packaged and pushed to the configured Helm repository at build time.
+
+### Approach 2: Manually curate a Helm chart
```
vasiliy-ul, line 188: Helm charts in fact use the content of the manifests. Since this content is generated for every build, it should be relatively easy to also structure it properly for Helm and produce the related artifacts. Those can be released as `*.tgz` or via `quay.io`. IMHO, it would be a better way, comparing to manual maintenance in a separate repository.

design-proposals/helm-chart.md
```diff
@@ -0,0 +1,249 @@
...
+## Update/Rollback Compatibility
+
+The feature will not impact older versions of KubeVirt or CDI.
+Helm charts will, however, be a viable way of updating & rolling back future releases.
+
+## Functional Testing Approach
+
+- Deploy KubeVirt from Helm chart
+- Ensure all components are created and work as expected
+- Slightly modify the configuration of the CRD, CR and operator's Deployment and upgrade KubeVirt
```
vasiliy-ul, line 238: I think this should be rather "update KubeVirt from N-1 to N release using Helm charts"

design-proposals/helm-chart.md
atanasdinov: This is indeed a viable option. I'll include it.

design-proposals/helm-chart.md
atanasdinov: The `Generate Helm chart(s)` definitely has the better end result. However, most charts are highly customisable and even if we don't take such approach, I imagine we'd still template some values so the manifests will not be completely the same.

This way the users would be able to work with their preferred configuration without relying on `kubectl patch` and / or manually modifying manifests before applying them e.g. configuring custom options for the CR would be possible via something like `$ helm install kubevirt-cr <repo-url> --set configuration.developerConfiguration.cpuAllocationRatio=5`

design-proposals/helm-chart.md
atanasdinov: This is definitely the better wording, will use it. Thanks!

atanasdinov:
Thank you for your input, @vasiliy-ul!

> Regarding `approach #2` I think there might be an issue if it always installs/updates the CR resource. Users may have some settings defined in the CR, and I assume those will be overwritten by an update.

Custom options will not be overwritten on upgrade. However, I ran into https://github.com/kubevirt/kubevirt/issues/2533 when upgrading KubeVirt with the [single chart approach](https://github.com/suse-edge/charts/tree/main/charts/kubevirt/0.1.0) which requires further investigation.

I'm currently testing all different setups and will update the proposal accordingly after I'm done.

vasiliy-ul:
> Custom options will not be overwritten on upgrade. 

Just wanted to double-check this. So, assuming there is a valid KubeVirt installation running on my cluster and I have enabled some feature gates by editing the CR resource `k edit kv -n kubevirt kubevirt`:
```yaml
...
spec:
  certificateRotateStrategy: {}
  configuration:
    developerConfiguration:
      featureGates:
      - WorkloadEncryptionSEV
...
```
The generated `kubevirt-cr.yaml` now looks like below:
```yaml
---
apiVersion: kubevirt.io/v1
kind: KubeVirt
metadata:
  name: kubevirt
  namespace: kubevirt
spec:
  certificateRotateStrategy: {}
  configuration:
    developerConfiguration:
      featureGates: []
  customizeComponents: {}
  imagePullPolicy: IfNotPresent
  workloadUpdateStrategy: {}
```
I assume it will look similar when generated for the Helm chart. If I do smth like `k apply -f _out/manifests/release/kubevirt-cr.yaml` then definitely I will end up in overwriting the existing feature gates with `featureGates: []`.

But what will happen on update with Helm? Does Helm perform the same `kubectl apply ...` or does it merge the existing resource with the new one?

atanasdinov:
Helm 3 will consider the previous manifest, the current live state and the new manifest.

The flow you described will keep the manually edited feature gate `WorkloadEncryptionSEV` even after :

1. `$ helm install kubevirt <repo-url>`
2. `$ kubectl edit kv -n kubevirt kubevirt` OR `$ kubectl patch kv kubevirt -n kubevirt --type=json -p '[{ "op": "add", "path": "/spec/configuration/developerConfiguration/featureGates/0", "value": "WorkloadEncryptionSEV" }]'`
3. `$ helm upgrade kubevirt <repo-url> --version <version>`

We can introduce a relevant functional test covering such scenarios.

More info on the topic from the [official Helm documentation](https://helm.sh/docs/faq/changes_since_helm2/#improved-upgrade-strategy-3-way-strategic-merge-patches).

atanasdinov:
> `approach #3` looks like a good alternative which basically replicates the existing flow with the manifests. It is well-known to the users already. Also, from the implementation and maintenance point of view, it does not seem to add any complexity. And in fact I think that by using the hooks it might be possible to e.g. prevent uninstallation of the operator if the CR is still present in the cluster. Thus enforcing proper uninstallation order. The only downside I see for now with this approach is related to potential inconsistency between operator/cr charts versions. When updating KubeVirt/CDI you usually do not need to update the CR, only the operator (and CRD). Therefore, after an update, the CR chart will stay at version N while the operator will move to N+1. Maybe this is in fact not a big issue at the end. But still maybe worth thinking about it... if it may cause some problems or confusions for the users.

Regarding the versioning concerns, [upgrading the operator](https://kubevirt.io/user-guide/operations/updating_and_deletion/#updating-kubevirt-control-plane) using the manifests will actually also upgrade the CR. I've confirmed that this also works for Helm upgrades with `approach #1`.

This means that the CR chart can stay at version N while the Operator / CRD chart is at version N + M. We'd only **need** to upgrade if there was a change to the CRD and / or there's a particular feature which we'd want to enable / disable in the standard CR manifest.

All this also applies to `approach #2` where we keep a separate CRD chart instead. There wouldn't be a reason to upgrade unless there is a change there and as the product matures such changes will be less and less frequent.

Regenerating the same CR or CRD charts (depending on the approach) is definitely not necessary and we can opt for doing so only when something has changed. However, I agree that it may bring confusion to the users.

---

There is also the option to attempt creating subcharts (either of CR or CRD) as dependencies of the main chart and install a single chart only, however we'd need to have multiple hooks taking care of the ordering which would make it so complex that it didn't even make the exotic options list.

---

Please let me know if any of the comments have not been fully addressed, thanks!

aburdenthehand:
/remove-lifecycle stale
@atanasdinov @vasiliy-ul - am I right to assume that we still want to pursue this? If so, I will see about getting some more eyes on this.

atanasdinov:
Hello, @aburdenthehand! We're still interested in contributing a Helm chart and would definitely appreciate it if we are able to get more people review the proposal and decide on the design and implementation options.

In the meantime, we plan on using a manually curated one which we'll bump to the GA release soon.

Please keep in mind that (depending on the implementation approach) we might not be able to start working on it right away due to limited capacity.

design-proposals/helm-chart.md
scydas: In my helm, I use approach `approach #1`.
But when upgrading crd need to do it manually, when deleting use re-delete to delete crd.

davenull:
As an outsider, but a very heavy user of this project, is there any assistance that I can offer to help speed this through review?

dhiller:
@vasiliy-ul do you have any updates on this one? Not exactly sure what the state is here?

vasiliy-ul:
@dhiller, please refer to the previous comment from @atanasdinov https://github.com/kubevirt/community/pull/224#issuecomment-1749101403

vladikr:
@atanasdinov It's a nice proposal. My only concern is how it can be maintained and supported as part of core KubeVirt? I doubt we have any expertise around Helm... 
Could this be set up as a sub-project?  Simar to @kubevirt/application-aware-quota ?
Since it's part of the KubeVirt org all kubevirt related tests can run on any new PR?

davenull:
The main issue with not having a chart or kustomize is CD, Argo and other cd systems are built around using charts, but kubevirt/cdi requires some funny dancing around to shoehorn into the workflow.

jkleinlercher:
Regarding the dependency between CRDs and CRs: there are other very popular charts which deploy CRDs as templates in the same chart as the CRs, e.g.: https://github.com/akuity/kargo/tree/main/charts/kargo/templates

see also https://github.com/akuity/kargo/releases#chart-improvements

so from this it should work to have CRDs in the same chart and have them as templates so that CRD upgrades should also work

aburdenthehand:
/remove-lifecycle stale
@vladikr Did the above comments allay your concerns?

vladikr:
> /remove-lifecycle stale @vladikr Did the above comments allay your concerns?

My preference would be to keep this chart outside of the kubevirt/kubevirt repo but in the kubevirt org.

design-proposals/helm-chart.md
Infinoid: I am not a kubevirt developer, but I had a thought.

I understand the motivation for generating one from the other.  It's [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself), right?  But generating a helm chart from the manifests seems backwards to me.  Couldn't you generate manifests from the helm chart, instead, using `helm template` and default values?

Patching templates into manifests sounds like it will become tricky to keep up with changes to the manifest in future releases.  Working in the other direction seems trivial to me (just apply the template).

That way, you'd get:
* a single source of truth (the helm chart)
* an easy, low-maintenance release process (just run `helm template`)
* guaranteed consistency between the two installation methods
* consistency of template parameters and values across versions
    * no scramble to update patches for new versions
* ability to add as much customization as you want, without exponentially increasing complexity

I may be wrong, maybe it's a terrible idea.  But I was surprised to see no one mention it yet.

aburdenthehand:
@davidvossel @mhenriks @awels can you please take a look at this?

Jmainguy:
As an end user, I wanted to say I really like this proposal, was searching for the helm chart today to discover one did not exist yet (managed by the project). The proposal is very well written and I can tell a good amount of work has gone into it and this ongoing discussion. I really appreciate everyone involved and look forward to when this is accepted.

design-proposals/helm-chart.md
```diff
@@ -0,0 +1,270 @@
...
+
+## User Stories
+
+- As a cluster administrator, I'd want to install, upgrade and uninstall KubeVirt using Helm chart
+- As a cluster administrator, I'd want to install, upgrade and uninstall CDI using Helm chart
+
+## Repos
+
+* [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+* [kubevirt/containerized-data-importer](https://github.com/kubevirt/containerized-data-importer)
```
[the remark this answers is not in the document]
atanasdinov: Yes, that's perfectly fine. I can update the proposal to mark that manually curated Helm chart in a separate Git repository is the selected approach.

We still have to decide on the chart layout, though. I assume the safest route is with the [second option](https://github.com/kubevirt/community/blob/0ff5608277b88a66892828c651db3692fe55604c/design-proposals/helm-chart.md#approach-2-one-crd-chart-and-one-cr--operator-chart). Does the community agree?

design-proposals/helm-chart.md
atanasdinov: All of this makes perfect sense. The initial goal of the proposal was not to be invasive with regards to how the manifests are currently generated hence why I didn't include it.

design-proposals/helm-chart.md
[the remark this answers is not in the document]
jkleinlercher: why not in the same chart as mentioned in https://github.com/kubevirt/community/pull/224#issuecomment-2039776484 ?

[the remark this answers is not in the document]
kfox1111: Both CRDS and CR's that use the CRD cant reliably be in the same chart. long standing helm issue. :/

As most folks want to deploy the operator, and then use it to immediately launch kubevirt, which object do you seperate? The CRDs or the CRs?

I generally like to manually apply static CRDs via kubectl, as they can be rather fiddly, then launch everything else with helm.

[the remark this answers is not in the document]
jkleinlercher: Okay I see, then I would prefer also the second option as mentioned above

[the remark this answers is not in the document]
davralin: > Both CRDS and CR's that use the CRD cant reliably be in the same chart. long standing helm issue. :/

I know that [helm](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/) says it's not reliable, but kube-prometheus-stack does it - and that is a rather popular project which heavily uses CRDs.

But they can't be [upgraded](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#upgrading-chart) directly.


One can use a dedicated chart that only contains the CRDs, also like what [kube-prometheus](https://artifacthub.io/packages/helm/prometheus-community/prometheus-operator-crds) does.

There are also other, less standard options, like putting all the CRD's in the template-dir for helm, instead of in the CRD-directory, like what [fluxcd-community-charts](https://github.com/fluxcd-community/helm-charts/blob/main/charts/flux2/templates/helm-controller.crds.yaml) do

It would be _really_ nice to be able to use a _only_ helm-chart-way of managing clusters - automations with gitops and helm is very nice, and having to drop to a command-line (or write an automation that does) kubectl apply isn't great at scale, [fluxcd](https://fluxcd.io/flux/components/helm/helmreleases/#controlling-the-lifecycle-of-custom-resource-definitions) has ways of handling this already, I assume ArgoCD has the same.

(But, I'm also a completely new person to kubevirt, so I have no idea of how kubevirt handles/changes CRDs over time, this might be a bigger issue to handle than what I know. I am subscriped to this topic, and eagerly await a helm-chart'y-way to install this project, as the current implementation comes with a bigger initial cost for me/my company, since we so heavily invested in helm and flux. I wanted to chime in as the approach of "just kubectl apply the CRD-upgrades" isn't a great scaling-option, and other options exist.)

davenull:
This is usually solved with an “install-cards: true/false” or two charts,
once for app, one for CRD. Mayastor for instance does the second method.

design-proposals/helm-chart.md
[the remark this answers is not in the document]
kfox1111: a crds chart separate from the main chart is a common solution. It still works with gitops.

sticking them in the templates directory does not work around the issue of having CRDS and CR's in the same chart.

xpivarc:
@aburdenthehand How do we move this forward?

scydas:
Two years have passed

aburdenthehand:
@xpivarc - Unless I'm mistaken, this was waiting for the KubeVirt community to voice a preference for the chart layout. This lack of preference could be taken as agreement with the suggestion of option2. I will ping some of the ppl previously tagged in this PR and raise it again in the community meeting for input. 
@atanasdinov It looks like the proposal has not been updated to specify that the charts will be maintained in a new and separate repo - is that right or am I misreading? It also looks like you need to sign one or more of your commits - if it's easiest, you can squash all commits into a single signed commit after updating it. If we're creating a new repo in the kubevirt project, it would be good to have a second person to help maintain these charts.

atanasdinov:
@aburdenthehand I'll update this with the chosen approaches (even if we hadn't had that much input from the community) in order to have this go under final review and hopefully get merged.

On a separate note, I've had all commits signed but I'll double check as soon as I update this. Thanks!

atanasdinov:
Updated the different sections with a "Decision" subsections with the selected approaches respectively. I intentionally did not rewrite the initial options in order to showcase those were consider and to keep the constraints clear.

We still need to decide where to publish and host the charts (i.e. HTTP(s) repository vs OCI registry). We can choose to address this in the scope of this proposal or at a later stage.

Let me know how you'd want to proceed.

CC: @aburdenthehand

aburdenthehand:
/lgtm
@atanasdinov Thank you for the update! I think that the name/details of the repo can be decided later. 
I'll aim to get some more +1s here so we can merge it. In the meantime, can I ask you have a look at the community membership guidelines and apply: https://github.com/kubevirt/community/blob/main/membership_policy.md#member
You're welcome to tag me to sponsor.

vladikr:
@atanasdinov Thank you. Could you please open an issue to track this proposal in kubevirt/enhancements repo?

vladikr:
@aburdenthehand @atanasdinov Thank you for this effort! 
Everything looks good to me. However, something is still not clear to me. Are we going to create a new repo for this? Will you update the proposal with the name?
Beside this this, I'm +1
/approve
/hold 
I'll let @lyarwood and @xpivarc have a final look.

PR #225, opened 2023-06-29
Proposal: mass machine type transition

prnaraya:
When RHEL 10 is released, machine types prior to RHEL-9 will no longer be compatible. This proposes a method for users to mass update the machine types of any VMs with unsupported machine types while minimizing disruptions to workload.

### DIFF
--- a/design-proposals/mass_machine_type_transition.md
+++ b/design-proposals/mass_machine_type_transition.md
@@ -1,0 +1,208 @@
+﻿# Overview
+This proposal discusses a method to automate updating the machine type of VMs to the latest machine type version, if the VM has a machine type that is no longer compatible. Currently we can support manually changing the machine type of individual VMs, but we need an automated process that will be able to change the machine types of multiple (e.g. thousands of) VMs, limiting workload interruptions and preventing the user from having to manually update the machine type of every VM with a machine type version that is no longer supported. Centos 9 will maintain compatibility with all `pc-q35-rhel8.x.x` machine types throughout its lifecycle. However, with the transition to Centos 10, compatibility with machine types prior to `pc-q35-rhel9.0.0` will not be maintained. This means each workload must be updated to use a `pc-q35-rhel9.x.x` machine type.
+
+## Goals
+* Creating an automated method for mass converting machine types of VMs
+	* if currently running, update VM machine type and add a label to alert the user the VM must be restarted for the change to take effect
+	* if currently running, take VM offline immediately and update machine type
+	* if not currently running, update machine type immediately
+* Allow user to specify certain VMs for updating machine type automation (e.g. by namespace)
+
+## Non Goals
+While these additions could be beneficial to cluster-admins who wish to use the mass machine type transition (MMTT), these are not covered in the scope of this design proposal: 
+* Allowing user to filter VMs by running state: a cluster admin may want to only update VMs that are offline.
+* Allowing user to filter VMs in ways other than by namespace; e.g. with label-selector
+
+Also not covered in this design proposal is the method in which users/cluster admins will be informed that they have VMs with unsupported machine types.
+
+## User Stories
+As a cluster-admin, I want an automated way to update the machine type of all VMs to be compatible with RHEL 10 without interrupting my workflow.
+As a cluster-admin, I want to be able to include only certain VMs to be updated (e.g. by namespace).
+As a cluster-admin, I want to be in control of the behavior of running VMs being updated; I can determine whether I want to manually restart running VMs being updated or let the automation restart all of them for me. 
+
+## Repos
+[kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+Create a new virtctl command that will allow the user to automatically update the machine type of any VMs with a machine type that is no longer supported. This command invokes a k8s `Job` that iterates through all VMs within a specified namespace (or all namespaces if none is specified), determines if the machine type is no longer supported, and updates it to the latest supported version. 
+
+If the VM is running, the update will not take effect until it is restarted, so the label `restart-vm-required` will be applied to the VM until the VMI is removed (indicating the VM has been stopped), upon which the label will be cleared. The job will only terminate once all `restart-vm-required` labels have been cleared. By default, the user must handle stopping/restarting the labelled VMs manually. This is to allow the user to safely update the machine types of their VMs at a time they choose without worrying about workload interruptions. However, the user also has the option to allow the job to automatically restart every running VM that is being updated and apply the changes immediately.
+
+The options to select a specific namespace or restart the VMs immediately are configurable with the `NAMESPACE` and `RESTART_NOW` environment variables respectively.
+
+Given a VM with the following spec:
+````
+---
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+...
+spec:
+  running: false
+  template:
+    spec:
+      domain:
+        machine:
+          type: pc-q35-rhel8.2.0
+````
+When the MMTT is invoked, the VM spec will be updated to: 
+````
+---
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+...
+spec:
+  running: false
+  template:
+    spec:
+      domain:
+        machine:
+          type: pc-q35-rhel9.2.0
+````
+where `pc-q35-rhel9.2.0` is the latest machine type version.
+
+Along with machine types being in the format `pc-q35-rhelx.x.x`, a VM may have a machine type of `q35`, an alias for the current latest machine type version *at the time the VM is started*.
+
+For example, this VM:
+````
+---
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+...
+spec:
+  running: true
+  template:
+    spec:
+      domain:
+        machine:
+          type: q35
+````
+may have been started when a RHEL-8 machine type was the latest machine type version and not been stopped or restarted since. Its VMI would thus look like this:
+````
+---
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+...
+spec:
+  domain:
+    machine:
+      type: q35
+status:
+  machine:
+    type: pc-q35-rhel8.2.0
+````
+and need to have its machine type updated.
+
+To handle this, any running VMs with the machine type `q35` will be labelled with `restart-vm-required`:
+````
+---
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+	labels: "restart-vm-required"="true"
+spec:
+  running: true
+  template:
+    spec:
+      domain:
+        machine:
+          type: q35
+````
+Once the VM has been stopped or restarted, the `restart-vm-required` label will be removed:
+````
+---
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+	labels:
+spec:
+  running: false
+  template:
+    spec:
+      domain:
+        machine:
+          type: q35
+````
+Now when a new VMI is created when starting this VM, it should have the latest machine type version:
+````
+---
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+...
+spec:
+  domain:
+    machine:
+      type: q35
+status:
+  machine:
+    type: pc-q35-rhel9.2.0
+````
+
+## Update/Rollback Compatibility
+As both the minimum supported machine type version and the latest machine type version change in the future, these values are global constants in the MMTT package. As the versions are updated in the future, these global constants can also be updated accordingly, allowing for easy maintainability.
+
+## Functional Testing Approach
+* Functional tests will follow the same basic procedure:
+	* Create (a) VM(s) with the necessary machine type for the test case; start the VM if testing functionality of running VMs
+	* Configure environment variables `NAMESPACE` and `RESTART_NOW` as necessary for the test case
+	* Ensure VM spec has the correctly updated machine type version
+	* Ensure the VM has/doesn't have the `restart-vm-required` label
+	* Ensure all VMs have been updated accordingly
+* Test cases:
+	* Single VM (each of these will be their own individual test)
+		* Running
+			* VM machine type version less than the minimum supported version
+				* `RESTART_NOW` is **false**
+				* `RESTART_NOW` is **true**
+			* VM machine type is `q35`
+				* `RESTART_NOW` is **false**
+				* `RESTART_NOW` is **true**
+			* VM machine type version is greater than or equal to the minimum supported version
+		* Not running
+			* VM machine type version less than the minimum supported version
+			* VM machine type version is greater than or equal to the minimum supported version
+			* VM machine type is equal to `q35`
+	* Multiple VMs (these cases will be split into 4 functional tests based on the environment variable configuration: `NAMESPACE`  specified/unspecified and `RESTART_NOW` **true**/**false**)
+		* `NAMESPACE` is specified
+			* `RESTART_NOW` is **true**
+				* Running
+					* VM machine type version less than the minimum supported version
+					* VM machine type version is greater than or equal to the minimum supported version
+					* VM machine type is equal to `q35`
+				* Not running
+					* VM machine type version less than the minimum supported version
+					* VM machine type version is greater than or equal to the minimum supported version
+					* VM machine type is equal to `q35`
+			* `RESTART_NOW` is **false**
+				* Running
+					* VM machine type version less than the minimum supported version
+					* VM machine type version is greater than or equal to the minimum supported version
+					* VM machine type is equal to `q35`
+				* Not running
+					* VM machine type version less than the minimum supported version
+					* VM machine type version is greater than or equal to the minimum supported version
+					* VM machine type is equal to `q35`
+		* `NAMESPACE` is not specified
+			* `RESTART_NOW` is **true**
+				* Running
+					* VM machine type version less than the minimum supported version
+					* VM machine type version is greater than or equal to the minimum supported version
+					* VM machine type is equal to `q35`
+				* Not running
+					* VM machine type version less than the minimum supported version
+					* VM machine type version is greater than or equal to the minimum supported version
+					* VM machine type is equal to `q35`
+			* `RESTART_NOW` is **false**
+				* Running
+					* VM machine type version less than the minimum supported version
+					* VM machine type version is greater than or equal to the minimum supported version
+					* VM machine type is equal to `q35`
+				* Not running
+					* VM machine type version less than the minimum supported version
+					* VM machine type version is greater than or equal to the minimum supported version
+					* VM machine type is equal to `q35`
+
+# Implementation Phases
+
+ - [ ] Create MMTT package that Kubernetes job will use as an image
+ - [ ] Create Kubernetes job yaml to be invoked with MMTT subcommand
+ - [ ] Insert package and job into kubevirt/pkg/virtctl and implement the subcommand (name TBD)
--- a/design-proposals/mass_machine_type_transition.md
+++ b/design-proposals/mass_machine_type_transition.md
@@ -1,5 +1,14 @@
 ﻿# Overview
-This proposal discusses a method to automate updating the machine type of VMs to the latest machine type version, if the VM has a machine type that is no longer compatible. Currently we can support manually changing the machine type of individual VMs, but we need an automated process that will be able to change the machine types of multiple (e.g. thousands of) VMs, limiting workload interruptions and preventing the user from having to manually update the machine type of every VM with a machine type version that is no longer supported. For example, CentOS Stream 9 will maintain compatibility with all `pc-q35-rhel8.x.x` machine types throughout its lifecycle. However, with the transition to CentOS Stream 10, compatibility with machine types prior to `pc-q35-rhel9.0.0` will not be maintained: 
+This proposal discusses a method to automate updating the machine type of VMs
+to the latest machine type version, if the VM has a machine type that is no
+longer compatible.Currently we can support manually changing the machine type of
+individual VMs, but we need an automated process that will be able to change the
+machine types of multiple (e.g. thousands of) VMs, limiting workload interruptions
+and preventing the user from having to manually update the machine type of every
+VM with a machine type version that is no longer supported. For example, CentOS
+Stream 9 will maintain compatibility with all `pc-q35-rhel8.x.x` machine types
+throughout its lifecycle. However, with the transition to CentOS Stream 10,
+compatibility with machine types prior to `pc-q35-rhel9.0.0` will not be maintained: 
 ```
 $ cat /etc/centos-release 
 CentOS Stream release 9
@@ -32,26 +41,63 @@ This means each workload must be updated to use a `pc-q35-rhel9.x.x` machine typ
 * Allow user to specify certain VMs for updating machine type automation (e.g. by namespace)
 
 ## Non Goals
-While these additions could be beneficial to cluster-admins who wish to use the mass machine type transition (MMTT), these are not covered in the scope of this design proposal: 
+While these additions are beneficial to cluster-admins who wish to use the
+mass machine type transition (MMTT), these will not be included in the initial implementation:
 * Allowing user to filter VMs by running state: a cluster admin may want to only update VMs that are offline.
 * Allowing user to filter VMs in ways other than by namespace; e.g. with label-selector
+* Subcommands that allow the user to monitor and manage the status of the MMTT job and affected VMs
 
-Also not covered in this design proposal is the method in which users/cluster admins will be informed that they have VMs with unsupported machine types.
+These will eventually be implemented in follow ups once the initial design is implemented.
+
+Additionally, this design proposal does not cover the method in which users/cluster admins
+will be informed that they have VMs with unsupported machine types.
 
 ## User Stories
-As a cluster-admin, I want an automated way to update the machine type of all VMs to be compatible with CentOS Stream X without interrupting my workflow.
+As a cluster-admin, I want an automated way to update the machine type of all
+VMs to be compatible with CentOS Stream X without interrupting my workflow.
 As a cluster-admin, I want to be able to include only certain VMs to be updated (e.g. by namespace).
-As a cluster-admin, I want to be in control of the behavior of running VMs being updated; I can determine whether I want to manually restart running VMs being updated or let the automation restart all of them for me. 
+As a cluster-admin, I want to be in control of the behavior of running VMs being
+updated; I can determine whether I want to manually restart running VMs being
+updated or let the automation restart all of them for me. 
 
 ## Repos
 [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
 
 # Design
-Create a new virtctl command that will allow the user to automatically update the machine type of any VMs with a machine type that is no longer supported. This command invokes a k8s `Job` that iterates through all VMs within a specified namespace (or all namespaces if none is specified), determines if the machine type is no longer supported, and updates it to the latest supported version. Both the minimum supported machine type and the target machine type are constants that can be updated as future versions are released and more machine types are deprecated upon these releases. 
+Create a new virtctl command `virtctl convert-machine-types` that will allow
+the user to automatically update the machine type of any VMs with a machine type
+that is no longer supported. This command invokes a k8s `Job` that iterates
+through all VMs within a specified namespace (or all namespaces if none is
+specified), determines if the machine type is no longer supported, and updates
+it to the latest supported version. If the Job fails or is killed, it will be
+restarted with `restartPolicy=OnFailure` in the Job spec. The `restart-vm-required`
+label may then remain on a VM that has already been shut down, so the Job will
+first check VMs with the label and confirm if the machine types have been updated.
+If they have, then the label will be removed, but if not, the label will remain
+and the VM will continue to be tracked until it is restarted and its machine
+type is updated. Both the minimum supported machine type and the target machine
+type will be determined (retrieved) and configured internally when the command is
+invoked, and these values will automatically be updated as future versions are
+released and more machine types are deprecated upon these releases.
 
-If the VM is running, the update will not take effect until it is restarted, so the label `restart-vm-required` will be applied to the VM until the VMI is removed (indicating the VM has been stopped), upon which the label will be cleared after verifying that its machine type has been successfully updated. The job will only terminate once all `restart-vm-required` labels have been cleared. By default, the user must handle stopping/restarting the labelled VMs manually. This is to allow the user to safely update the machine types of their VMs at a time they choose without worrying about workload interruptions. However, the user also has the option to allow the job to automatically restart every running VM that is being updated and apply the changes immediately.
+If the VM is running, the update will not take effect until it is restarted,
+so the label `restart-vm-required` will be applied to the VM until the VMI is
+removed (indicating the VM has been stopped), upon which the label will be cleared
+after verifying that its machine type has been successfully updated. The job will
+only terminate once all `restart-vm-required` labels have been cleared. By default,
+the user must handle stopping/restarting the labelled VMs manually. This is to allow
+the user to safely update the machine types of their VMs at a time they choose
+without worrying about workload interruptions. However, the user also has the
+option to allow the job to automatically restart every running VM that is being
+updated and apply the changes immediately.
 
-The options to select a specific namespace or restart the VMs immediately are configurable with the `NAMESPACE` and `RESTART_NOW` environment variables respectively.
+The options to specify by namespace, label-selector, or restart the VMs
+immediately are configurable with the `--namespace`, `--label-selector`,
+and `--restart-now` respectively. In follow-ups to the initial implementation,
+other methods of specifying which VMs to convert the machine types of will be
+added, such as specifying a single VM by name, or configuring a limit on the
+number of VMs that can be restarted at once when restarting the running VMs
+immediately.
 
 Given a VM with the following spec:
 ```yaml
@@ -83,7 +129,8 @@ spec:
 ```
 where `pc-q35-rhel9.2.0` is the latest machine type version.
 
-Along with machine types being in the format `pc-q35-rhelx.x.x`, a VM may have a machine type of `q35`, an alias for the current latest machine type version *at the time the VM is started*.
+Along with machine types being in the format `pc-q35-rhelx.x.x`, a VM may have
+a machine type of `q35`, an alias for the current latest machine type *at the time the VM is started*.
 
 For example, this VM:
 ```yaml
@@ -99,7 +146,8 @@ spec:
         machine:
           type: q35
 ```
-may have been started when a RHEL-8 machine type was the latest machine type version and not been stopped or restarted since. Its VMI would thus look like this:
+may have been started when a RHEL-8 machine type was the latest machine type and
+not been stopped or restarted since. Its VMI would thus look like this:
 ```yaml
 ---
 apiVersion: kubevirt.io/v1
@@ -115,7 +163,8 @@ status:
 ```
 and need to have its machine type updated.
 
-To handle this, any running VMs with 'q35' machine type and an outdated machine type (reported in VMI's status.machine.type) will be labelled with `restart-vm-required`:
+To handle this, any running VMs with 'q35' machine type and an outdated machine
+type (reported in VMI's status.machine.type) will be labelled with `restart-vm-required`:
 ```yaml
 ---
 apiVersion: kubevirt.io/v1
@@ -143,7 +192,7 @@ spec:
         machine:
           type: q35
 ```
-Now when a new VMI is created when starting this VM, it should have the latest machine type version:
+Now when a new VMI is created when starting this VM, it should have the latest machine type:
 ```yaml
 ---
 apiVersion: kubevirt.io/v1
@@ -159,7 +208,12 @@ status:
 ```
 
 ## Update/Rollback Compatibility
-As both the minimum supported machine type version and the latest machine type version change in the future, these values are global constants in the MMTT package. As the versions are updated in the future, these global constants can also be updated accordingly, allowing for easy maintainability.
+As both the minimum supported machine type version and the latest machine type
+change in the future when new versions are released, Kubevirt is already able
+to determine the latest machine type, so that will automatically be updated.
+The command will also be fetching the minimum supported machine type using
+the list of supported and deprecated QEMU machine types, thus minimizing the
+need to manually maintain these values.
 
 ## Functional Testing Approach
 * Functional tests will follow the same basic procedure:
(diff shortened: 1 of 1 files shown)


design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,208 @@
...
+...
+spec:
+  running: false
+  template:
+    spec:
+      domain:
+        machine:
+          type: pc-q35-rhel9.2.0
+````
+where `pc-q35-rhel9.2.0` is the latest machine type version.
```
lyarwood, line 61: I guess the job is using `/usr/libexec/qemu-kvm -machine ?` to determine this?

Could we also consider the target machine type being provided by the admin?

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,208 @@
...
+...
+spec:
+  running: true
+  template:
+    spec:
+      domain:
+        machine:
+          type: q35
+````
+may have been started when a RHEL-8 machine type was the latest machine type version and not been stopped or restarted since. Its VMI would thus look like this:
```
lyarwood, line 79: Odd, any idea why we mutate the `VMI` with this? QEMU should be able to track this across migrations so I can't think why we would need to track it on our side.

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,208 @@
...
+  domain:
+    machine:
+      type: q35
+status:
+  machine:
+    type: pc-q35-rhel9.2.0
+````
+
+## Update/Rollback Compatibility
+As both the minimum supported machine type version and the latest machine type version change in the future, these values are global constants in the MMTT package. As the versions are updated in the future, these global constants can also be updated accordingly, allowing for easy maintainability.
```
lyarwood, line 141: Ah so these are going to be constants? Could you reference this aspect in the design section?

lyarwood (changes requested):
Excellent proposal, a few comments inline but this is pretty cool thanks!

design-proposals/mass_machine_type_transition.md
stu-gott: I think you're both saying the same thing. migrations aren't restarts.

stu-gott: As for why it needs to be tracked at the VMI level? Only virt-launcher has access to qemu directly.

lyarwood: I'm asking specifically why we expand the QEMU alias of `q35` from the VM to `pc-q35-rhel8.2.0` in the VMI instead of just using the `q35` alias for both and passing that to QEMU?

lyarwood: Ah! We don't mutate the `VirtualMachineInstance` but somehow a static type is listed in the domain XML?

```
$ ./cluster-up/kubectl.sh get vm/cirros -o json | jq .spec.template.spec.domain.machine.type
selecting podman as container runtime
"q35"

$ ./cluster-up/kubectl.sh get vmi/cirros -o json | jq .spec.domain.machine.type
selecting podman as container runtime
"q35"

$ ./cluster-up/kubectl.sh exec -ti virt-launcher-cirros-kjn2n bash
selecting podman as container runtime
[..]
$ virsh dumpxml 1 | grep q35
    <type arch='x86_64' machine='pc-q35-rhel9.2.0'>hvm</type>
```

Anyway labelling any `VirtualMachines` using an alias is still valid but the suggestion that the `VirtualMachineInstance` is mutated needs to be dropped.

lyarwood: This does get reflected in the `status` of the `VirtualMachineInstance` FWIW:
```
./cluster-up/kubectl.sh get vmi/cirros -o json | jq .status.machine
selecting podman as container runtime
{
  "type": "pc-q35-rhel9.2.0"
}
```

design-proposals/mass_machine_type_transition.md
prnaraya: In the case of allowing the admin configure the target machine type, do we want to limit it to only supported machine types (in this case, just the RHEL-9 machine types)?

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,208 @@
+﻿# Overview
+This proposal discusses a method to automate updating the machine type of VMs to the latest machine type version, if the VM has a machine type that is no longer compatible. Currently we can support manually changing the machine type of individual VMs, but we need an automated process that will be able to change the machine types of multiple (e.g. thousands of) VMs, limiting workload interruptions and preventing the user from having to manually update the machine type of every VM with a machine type version that is no longer supported. Centos 9 will maintain compatibility with all `pc-q35-rhel8.x.x` machine types throughout its lifecycle. However, with the transition to Centos 10, compatibility with machine types prior to `pc-q35-rhel9.0.0` will not be maintained. This means each workload must be updated to use a `pc-q35-rhel9.x.x` machine type.
```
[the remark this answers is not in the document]
prnaraya: Since `pc-i440fx-rhel7.6.0` is listed as deprecated here, when encountering a machine type of the alias `pc` should it be handled the same way as `q35` and do we want to replace it with a `q35` machine type entirely?

design-proposals/mass_machine_type_transition.md
prnaraya: Yes these will both be constants, I've updated the design section.

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Goals
+* Creating an automated method for mass converting machine types of VMs
+	* if currently running, update VM machine type and add a label to alert the user the VM must be restarted for the change to take effect
+	* if currently running, take VM offline immediately and update machine type
```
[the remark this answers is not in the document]
prnaraya: Added.

design-proposals/mass_machine_type_transition.md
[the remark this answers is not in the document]
iholder101: As written in QEMU's docs [1]:
> These old machine types are quite neglected nowadays and thus might have various pitfalls with regards to live migration. Use a newer machine type instead.

IOW, this machine type is deprecated. However, it is a different machine type than q35.

I think there might be a high probability that whoever chose this machine type might be interested in it specifically.
WDYT about avoiding changing it automatically to q35, but instead adding a warning label?

[1] https://www.qemu.org/docs/master/about/deprecated.html

design-proposals/mass_machine_type_transition.md
iholder101: I think it would be ideal that the admin would choose between a specified type and the `latest` type that would be determined dynamically. Perhaps we can utilize `node-labeller` to populate the latest machine type supported, then we could fetch the latest version that's supported on all nodes.

> do we want to limit it to only supported machine types (in this case, just the RHEL-9 machine types)?

We could, via `/usr/libexec/qemu-kvm -machine ?`, parse the different versions to give the admin a choice, but I would avoid providing specific hard-coded versions that would demand us updating the versions manually whenever needed.

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,230 @@
+﻿# Overview
+This proposal discusses a method to automate updating the machine type of VMs to the latest machine type version, if the VM has a machine type that is no longer compatible. Currently we can support manually changing the machine type of individual VMs, but we need an automated process that will be able to change the machine types of multiple (e.g. thousands of) VMs, limiting workload interruptions and preventing the user from having to manually update the machine type of every VM with a machine type version that is no longer supported. CentOS Stream 9 will maintain compatibility with all `pc-q35-rhel8.x.x` machine types throughout its lifecycle. However, with the transition to CentOS Stream 10, compatibility with machine types prior to `pc-q35-rhel9.0.0` will not be maintained:
```
iholder101, line 2:
```suggestion
This proposal discusses a method to automate updating the machine type of VMs to the latest machine type version, if the VM has a machine type that is no longer compatible. Currently we can support manually changing the machine type of individual VMs, but we need an automated process that will be able to change the machine types of multiple (e.g. thousands of) VMs, limiting workload interruptions and preventing the user from having to manually update the machine type of every VM with a machine type version that is no longer supported. For example, CentOS Stream 9 will maintain compatibility with all `pc-q35-rhel8.x.x` machine types throughout its lifecycle. However, with the transition to CentOS Stream 10, compatibility with machine types prior to `pc-q35-rhel9.0.0` will not be maintained: 
```

Adding "For example" here just to emphasize that this could be used in the future for other versions as well.

design-proposals/mass_machine_type_transition.md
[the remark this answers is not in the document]
iholder101: I agree with @lyarwood, the first two sub-points seem to contradict each other.
You can turn the second sub-point to:

> * Optionally, if currently running, take VM offline immediately and update machine type

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,230 @@
...
+pc-q35-rhel7.6.0     RHEL-7.6.0 PC (Q35 + ICH9, 2009) (deprecated)
+none                 empty machine
+This means each workload must be updated to use a `pc-q35-rhel9.x.x` machine type.
+````
+## Goals
+* Create an opt-in workflow that automates mass converting machine types of VMs
+	* if currently running, update VM machine type and add a label to alert the user the VM must be restarted for the change to take effect
+	* if currently running, take VM offline immediately and update machine type
+	* if not currently running, update machine type immediately
+* Allow user to specify certain VMs for updating machine type automation (e.g. by namespace)
```
iholder101, line 31: How about also supporting a label-selector?

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Non Goals
+While these additions could be beneficial to cluster-admins who wish to use the mass machine type transition (MMTT), these are not covered in the scope of this design proposal: 
+* Allowing user to filter VMs by running state: a cluster admin may want to only update VMs that are offline.
+* Allowing user to filter VMs in ways other than by namespace; e.g. with label-selector
+
+Also not covered in this design proposal is the method in which users/cluster admins will be informed that they have VMs with unsupported machine types.
+
+## User Stories
+As a cluster-admin, I want an automated way to update the machine type of all VMs to be compatible with RHEL 10 without interrupting my workflow.
```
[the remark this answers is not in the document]
iholder101: I would actually go with `CentOS Stream X`, just to emphasize this is tool is not just about a specific version

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,230 @@
...
+  domain:
+    machine:
+      type: q35
+status:
+  machine:
+    type: pc-q35-rhel8.2.0
+````
+and need to have its machine type updated.
+
+To handle this, any running VMs with the machine type `q35` will be labelled with `restart-vm-required`:
```
iholder101, line 117:
```suggestion
To handle this, any running VMs with an outdated machine type (reported in VMI's status.machine.type) will be labelled with `restart-vm-required`:
```

We should do that only for VMs with an outdated machine type.
Am I missing anything?

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,230 @@
...
+kind: VirtualMachine
+metadata:
+	labels:
+spec:
+  running: false
+  template:
+    spec:
+      domain:
+        machine:
+          type: q35
```
iholder101, line 145: nit: removing empty label section

```suggestion
apiVersion: kubevirt.io/v1
kind: VirtualMachine
spec:
  running: false
  template:
    spec:
      domain:
        machine:
          type: q35
```

iholder101: BTW, a nice trick: if you provide the file format:

> \```yaml
> ....
> \```

You get syntax highlight for free :)

```yaml
spec:
  running: true
  template:
    spec:
      domain:
        machine:
          type: q35
```

design-proposals/mass_machine_type_transition.md
iholder101: As said above I'd prefer dynamic values over constant ones

iholder101 (commented):
Great work @prnaraya!
See comments below

design-proposals/mass_machine_type_transition.md
lyarwood: > I think it would be ideal that the admin would choose between a specified type and the `latest` type that would be determined dynamically. Perhaps we can utilize `node-labeller` to populate the latest machine type supported, then we could fetch the latest version that's supported on all nodes.

They can already use the aliases to track the `latest` machine type, you could have another way for them to pick and pin to the latest machine type but that might be over complicating things a little.

> > do we want to limit it to only supported machine types (in this case, just the RHEL-9 machine types)?
> 
> We could, via `/usr/libexec/qemu-kvm -machine ?`, parse the different versions to give the admin a choice, but I would avoid providing specific hard-coded versions that would demand us updating the versions manually whenever needed.

Getting the supported versions dynamically might be awkward tbh, in the past for OpenStack I just ended up defining some basic regex covering the known type names (note we had to handle upstream *and* downstream QEMU while AFAIK KubeVirt only supports downstream CentOS Stream QEMU right?).

https://github.com/openstack/nova/blob/4b454febf73cdd7b5be0a2dad272c1d7685fac9e/nova/virt/libvirt/machine_type_utils.py#L24-L39

design-proposals/mass_machine_type_transition.md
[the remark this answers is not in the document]
lyarwood: It's deprecated and was scheduled to be removed in EL 9 before being punted to EL 10 AFAIK and prompted the OpenStack tooling that I previously worked on.

Totally agree that we shouldn't migrate VMs automatically as the `pc` chipset differs greatly to `q35` and could require manual changes to the VM before it's bootable.

The following OpenStack docs cover some of this:

https://docs.openstack.org/nova/latest/admin/hw-machine-type.html

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,230 @@
...
+## User Stories
+As a cluster-admin, I want an automated way to update the machine type of all VMs to be compatible with CentOS Stream 10 without interrupting my workflow.
+As a cluster-admin, I want to be able to include only certain VMs to be updated (e.g. by namespace).
+As a cluster-admin, I want to be in control of the behavior of running VMs being updated; I can determine whether I want to manually restart running VMs being updated or let the automation restart all of them for me. 
+
+## Repos
+[kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+Create a new virtctl command that will allow the user to automatically update the machine type of any VMs with a machine type that is no longer supported. This command invokes a k8s `Job` that iterates through all VMs within a specified namespace (or all namespaces if none is specified), determines if the machine type is no longer supported, and updates it to the latest supported version. Both the minimum supported machine type and the target machine type are constants that can be updated as future versions are released and more machine types are deprecated upon these releases.
```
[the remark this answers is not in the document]
0xFelix: Is there a benefit in running a `Job` instead of doing it directly in `virtctl`? I guess it is about run time of the command vs. having a separate binary to run inside the `Job`?

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,230 @@
...
+
+If the VM is running, the update will not take effect until it is restarted, so the label `restart-vm-required` will be applied to the VM until the VMI is removed (indicating the VM has been stopped), upon which the label will be cleared. The job will only terminate once all `restart-vm-required` labels have been cleared. By default, the user must handle stopping/restarting the labelled VMs manually. This is to allow the user to safely update the machine types of their VMs at a time they choose without worrying about workload interruptions. However, the user also has the option to allow the job to automatically restart every running VM that is being updated and apply the changes immediately.
```
0xFelix, line 51: I'm not sure if this should run inside a `Job` or if an "operator" would be more suited to do this? The job could potentially run for a very very long time and might even be stopped/killed before all VMs were updated?

0xFelix (commented):
Great idea! Added some comments.

design-proposals/mass_machine_type_transition.md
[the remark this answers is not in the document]
iholder101: Hey @0xFelix!

As I see it, the main benefit of running this inside a job is that you can get a status of the mass conversion process. For example, you can see that you had 20 VMs that were outdated, but now 7 had restarted so there are 13 more to go, the remaining VM names, etc. In addition, you can clearly see when the whole operation succeeds, meaning that you're sure your cluster is fully updated. A job would also automatically add/remove warning labels.

Doing so directly with virtctl would put the burden of tracking the outdated VMs, removing warning labels, and tracking the status of the overall conversion task on the user.

Essentially, this approach is more declarative than imperative. Since as an admin I would like to declaratively state "I want all of the outdated VMs to convert their machine type into a new one", and not mind about how exactly this is being done behind the scenes, I think a job fits perfectly here.

WDYT?

design-proposals/mass_machine_type_transition.md
iholder101: My opinion is that, in general, we already treat our components as monoliths more than micro-services. For example, virt-controller, just like `util.go`, is a generic name for "insert any logic you need here". This caused `virt-controller`, just like many of our other core components, to not be cohesive and deal with many unrelated tasks. IMHO this is a serious problem that hurts us in many ways, e.g. performance, scalability, maintainability, and will only get more severe over time if we don't change how things are being done.

IMO, This is a perfect example.

We're talking about a rare operation that an admin would likely to perform once in a few years. This operation is self-contained, has a very specific goal that is unrelated to virt-contoller's role, and I see no reason it should be handled by virt-controller or any other core component.

> The job could potentially run for a very very long time

The job would run as long as the cluster is not fully updated. As I see it, this is a feature, not a bug. BTW, by "running" I basically mean that the process would sleep, waiting for events about VMIs being updated, so in terms of performance I find it negligible. But, in terms of usability, I think that's actually great. As an admin I would know the status of the operation in a single and dedicated place.

design-proposals/mass_machine_type_transition.md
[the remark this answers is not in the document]
0xFelix: I was just concerned about the run time of the Job and its state. Say the cluster has a large amount of VMs and the Job may gets killed inbetween. How does the user know? Will there be any state left-overs from the Job? Can it be restarted to resume the operation?

Also, how often would this Job be needed? How often do machine types change?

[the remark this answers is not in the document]
0xFelix: IOW it may become a bit imperative, as the users needs to check on the job to make sure it succeeded or to restart it. Declarative would be more like declare it once and forget about it?

design-proposals/mass_machine_type_transition.md
0xFelix: IMO it doesn't have to be in the `virt-controller`, but it could be in an independent micro-service too.

If it is really just a operation performed once in a few years then I'd agree with keeping it as simple as possible and using a Job instead of a long running operator.

As I said in my other comment, I was just concerned about the state of the Job and it maybe being killed before it was able to complete the operation. So an admin would have to monitor it until it eventually succeeded.

design-proposals/mass_machine_type_transition.md
prnaraya: I think adding support for a label-selector is a good idea, although I think I want to get the base version with just specifying by namespace working first, then add the label-selector support in a follow up.

design-proposals/mass_machine_type_transition.md
prnaraya: > the latest type that would be determined dynamically

Would replacing the machine types with just `q35` and restarting the VMs be sufficient then for updating the machine types to the latest version, since 'q35' is the alias for the latest version? Then if the admin wishes to use a different target version they can specify (choosing from a list of supported machine types? or specifying any machine type?)

design-proposals/mass_machine_type_transition.md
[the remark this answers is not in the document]
iholder101: Thanks a lot @0xFelix for this discussion!

> Would be good if the two points were added to the proposal.

I completely agree.
@prnaraya WDYT?

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,229 @@
+﻿# Overview
+This proposal discusses a method to automate updating the machine type of VMs to the latest machine type version, if the VM has a machine type that is no longer compatible. Currently we can support manually changing the machine type of individual VMs, but we need an automated process that will be able to change the machine types of multiple (e.g. thousands of) VMs, limiting workload interruptions and preventing the user from having to manually update the machine type of every VM with a machine type version that is no longer supported. For example, CentOS Stream 9 will maintain compatibility with all `pc-q35-rhel8.x.x` machine types throughout its lifecycle. However, with the transition to CentOS Stream 10, compatibility with machine types prior to `pc-q35-rhel9.0.0` will not be maintained:
```
fabiand, line 2: Doing a line break after 80 chars or so would make this proposal easier to read :)

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,229 @@
...
+## User Stories
+As a cluster-admin, I want an automated way to update the machine type of all VMs to be compatible with CentOS Stream X without interrupting my workflow.
+As a cluster-admin, I want to be able to include only certain VMs to be updated (e.g. by namespace).
+As a cluster-admin, I want to be in control of the behavior of running VMs being updated; I can determine whether I want to manually restart running VMs being updated or let the automation restart all of them for me. 
+
+## Repos
+[kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+Create a new virtctl command that will allow the user to automatically update the machine type of any VMs with a machine type that is no longer supported. This command invokes a k8s `Job` that iterates through all VMs within a specified namespace (or all namespaces if none is specified), determines if the machine type is no longer supported, and updates it to the latest supported version. Both the minimum supported machine type and the target machine type are constants that can be updated as future versions are released and more machine types are deprecated upon these releases.
```
fabiand, line 50: What would be the virtctl command?

fabiand, line 50: It's good to look at the "mass" case, but what about a more narrow one, updating one vm?

I did like Felix comment that what about simply making it a direct functionality of virtctl instead of virtctl creating  Job.
Will the virtctl command for creating the job also have a way to cancel the job?

fabiand, line 50: How does the system know what machine types are outdated on the given cluster?

And how does the system know to what version to update the machine type of a vm?
Would it be enough to just unset `machine.type` as our admission webhooks will auto fill it in with the most recent?

Please explain if the system has the knowledge (and how it is gaining it) or if the user is specifying this. And if we assuem the user, then where (the heck!) the user will get this info from.

design-proposals/mass_machine_type_transition.md
fabiand: No. `q35` is a "floating" type and iwll always "point" to the latest q35.
IIUIC then we rather want to use the most recent specific (non-floating) machine type.
Ths is required in order to then keep the machine type stable until the next update of a machine type.

fabiand (commented):
A few comments.

design-proposals/mass_machine_type_transition.md
iholder101: Yeah, that works for me. Can you add these plans under "Implementation Phases"?

design-proposals/mass_machine_type_transition.md
iholder101: > It's good to look at the "mass" case, but what about a more narrow one, updating one vm?
>
> I did like Felix comment that what about simply making it a direct functionality of virtctl instead of virtctl creating Job.

Firstly, I think that the "mass" case is more common, as usually an admin would convert machine types only if he has to, and usually before upgrades. But as for the the case of a single VM (or a few VMs), I think it's still convenient that you have an object that reflects the status of the conversion, and that makes this whole operation more declarative.

Here's another use-case: let's say that an admin converts the machine types of some VMs, then does so for a few more VMs, and so on. Tracking this without jobs would be messy IMO. With jobs, you know exactly which groups of VMs were aimed for conversion, when, and what's the status of the conversion.

In general, if we look at it from a high-level perspective, there are 3 phases to the conversion:
1. Listing the VMs that we want to convert
2. Waiting for them to shut down / restart
3. "Finalizing" the conversion, or letting the user know that the conversion is done

I think that performing steps 2-3 imperatively would be a burden for the administrator.

If we would go with the imperative virtctl-only approach, how would an administrator be able to track progress in an easy way? How could he be notified when the conversion is done, or get a clear status of the operation's state? How would the administrator perform re-tries is something goes wrong? I'm open to discuss this front if we overcome these challenges.

cc @0xFelix 

> Will the virtctl command for creating the job also have a way to cancel the job?

Great question.
Perhaps to make this a whole solution we need something like the following commands:
1. `virtctl machine-type-convert start` - to define the target group of VMs, desired machine type, etc
2. `virtctl machine-type-convert list` - to list the currently running jobs
3. `virtctl machine-type-convert delete` - to delete a job.

iholder101: This was briefly discussed [above](https://github.com/kubevirt/community/pull/225#discussion_r1249356835). I completely agree that either a hard-coded version or a user-specified version (which, as you said, the user won't know where to get from) is a bad idea. 

As I see it, we have a few reasonable solutions:
1. Fetching the version dynamically. As @lyarwood [mentioned](https://github.com/kubevirt/community/pull/225#discussion_r1250506880), this is a bit tricky since we would need to support both downstream and upstream versions via regex or something similar. Saying that, if the version formats don't change too often and are pretty consistent this might not be such a bad idea.
2. Store the most-recent version in HCO on a per-release basis. This is also not great, as someone needs to update it every time a release takes place, which is error-prone.
3. Provide the user a list of possible versions, then asking him to specify the desired version. At least with this solution the admin would have a general idea which version to choose, and we can ease it even further with some documentation.

WDYT?

fabiand: I did not mean to say that we do not need a mass tool at all.

However, I'm missing 1 critical step (and 1 use-case):
How does the admin try out the migration on one or so VMs before applying it to many more?
If I was an admin, I want to try the migration on all the different VM types I have, before changing all of them.
After all this operation CAN lead to functional regressions, this is the reason why we do not do this automatically.

Thus to me a 2 layered approch just looked reasonable, i.e.
1. `virtctl vm update machine-type $VMNAME` # first trial
2. `virtctl vm update machine-type --jobs $LABELSELECTOR` # now all

0xFelix: @fabiand Are you thinking of a hybrid approach where 1 is running locally in `virtctl` while 2 is running as a `Job` on the cluster? For the modification of a single VM the `virtctl` command could be quite trivial, but afterall this would also be covered by @iholder101's suggestion to define the target group of VMs (could also be a single VM). In this case I'd still vote for only the `Job` because this avoids duplicating the updater logic in two places.

To get back to @iholder101's concern about monitoring the state of the operation. I have to agree that it would be easier to track the state of it in a `Job`. How else would you know which VMs were already upgraded? Where do you store the information?

0xFelix: I would go with approach 3 as IMO it is the most versatile and predictable one. You can set a filter to replace specific versions or a wildcard to replace all (which makes it behave like approach 1 again).

prnaraya: I do think that in the case of a single VM, we could still utilize the same implementation method of invoking the `Job` on that single VM, but allowing the user to specify that VM by name. I currently have this listed as a follow up in the implementation phases section of the design proposal, but if you feel this would be important in the initial design I can also add this single VM use case to the initial design and implement a flag to specify a VM by name. I agree with @0xFelix that I would like to avoid duplicating the logic for updating the VMs and keeping the design limited to the `Job` approach.

prnaraya: At the moment, since KubeVirt knows the most recent machine type, the initial implementation can utilize that and also determine the minimum supported (i.e. oldest machine type that is not deprecated) from QEMU's list of supported machines to keep the initial implementation simple. However this is worth discussing and planning more and I would like to add it to the design as a follow up (as I have added in the Implementation Phases section of the design proposal).

design-proposals/mass_machine_type_transition.md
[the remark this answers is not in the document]
prnaraya: I've updated the design proposal to include how the Job will be handled if it is killed and how leftovers (leftover labels) will be handled upon the Job restarting.

design-proposals/mass_machine_type_transition.md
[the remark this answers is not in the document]
prnaraya: I agree that this would be good to add, I've added it as a follow up in the Implementation Phases section of the design proposal.

prnaraya:
I have updated the design proposal, most notably I have updated the Implementation Phases section and split it into an initial implementation phase and a future implementation (follow ups) phase. There are some things that have been brought up in discussion on this design proposal that are definitely worth looking into further, however in an effort to reduce feature creep I would like to relegate some of these to follow ups. Please let me know what you think, and if there is anything that I have listed as a follow up that would be imperative to the initial implementation.

Also please let me know if I missed something or there are any leftovers/outdated parts in the rest of the design proposal. Thanks everyone for the feedback!

@fabiand @0xFelix @iholder101

design-proposals/mass_machine_type_transition.md
```diff
@@ -32,26 +41,63 @@ This means each workload must be updated to use a `pc-q35-rhel9.x.x` machine typ
...
 
-Also not covered in this design proposal is the method in which users/cluster admins will be informed that they have VMs with unsupported machine types.
+These will eventually be implemented in follow ups once the initial design is implemented.
+
+Additionally, this design proposal does not cover the method in which users/cluster admins
+will be informed that they have VMs with unsupported machine types.
 
 ## User Stories
-As a cluster-admin, I want an automated way to update the machine type of all VMs to be compatible with CentOS Stream X without interrupting my workflow.
+As a cluster-admin, I want an automated way to update the machine type of all
```
0xFelix, line 56: Should this be limited to CentOS Stream only? IMO this could apply to any OS?

design-proposals/mass_machine_type_transition.md
```diff
@@ -171,59 +225,66 @@ As both the minimum supported machine type version and the latest machine type v
...
 					* VM machine type is equal to `q35`
 
 # Implementation Phases
-
- - [ ] Create MMTT package that Kubernetes job will use as an image
- - [ ] Create Kubernetes job yaml to be invoked with MMTT subcommand
- - [ ] Insert package and job into kubevirt/pkg/virtctl and implement the subcommand (name TBD)
+## Initial Phase
+This will be a bare-bones design that will be functional for the user, but with limited features that focus on the specific use case of upgrading the machine types of VMs with outdated or deprecated machine types to the latest supported machine type. The user will have the ability to use the command with the flags to specify by namespace, label-selector,  or to restart any running VMs immediately. In this stage, the minimum supported machine type and the latest machine type will be determined and stored internally; the user will not be able to configure what machine types they would like to use at this time. Additionally, to monitor and manage/delete the job, the user can utilize the `kubectl` commands. In future phases, we will add subcommands that allow the user to monitor and manage the specific machine type transition job directly.
```
0xFelix, line 281: The `virtctl` subcommand could output an example command to monitor the Job once it was created. WDYT?

0xFelix (commented):
Added two more comments, other than that looks good to me!

design-proposals/mass_machine_type_transition.md
iholder101: +1

Unless I'm missing something we should be able to use this mechanism for any machine type / OS

design-proposals/mass_machine_type_transition.md
iholder101: Nice idea!

design-proposals/mass_machine_type_transition.md
```diff
@@ -171,59 +225,66 @@ As both the minimum supported machine type version and the latest machine type v
...
+ - Dynamically retrieving the supported and deprecated QEMU machine types and allowing the user to select a specific machine type to migrate to. Some more discussion and planning will be required to flesh out exactly how to implement this; namely how the user will know what machine types are supported and any restrictions we want to place on the user when selecting a machine type to convert to.
+ - Subcommands that the user can use to monitor and manage the jobs. These could include `virtctl convert-machine-types delete` which will safely terminate and clean up the job and delete it, and `virtctl convert-machine-types list` which will allow the user to see the status of the VMs being updated by the job.
+ - Other flags that the user can specify VMs to be affected by. For example, if the user wants to only convert a single specific VM, we might want to implement a flag to allow the user to specify the VM by name.
+ - When using the `restart-now` flag, allow the user to configure how many VMs they want to restart at once; depending on the number of running VMs, it may be very intensive to trigger the restart of possibly thousands of VMs at once.
```
iholder101, line 290: I'd suggest that once this proposal merges you'd open an issue to track these future steps

design-proposals/mass_machine_type_transition.md
iholder101, line 281: nit: break into more lines so it's easier to comment about a specific sentence :)

iholder101 (commented):
Thanks a lot @prnaraya!

/lgtm

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,290 @@
...
+	* if currently running, update VM machine type and add a label to alert the user the VM must be restarted for the change to take effect
+	* alternatively, if currently running, take VM offline immediately and update machine type
+	* if not currently running, update machine type immediately
+* Allow user to specify certain VMs for updating machine type automation (e.g. by namespace)
+
+## Non Goals
+While these additions are beneficial to cluster-admins who wish to use the
+mass machine type transition (MMTT), these will not be included in the initial implementation:
+* Allowing user to filter VMs by running state: a cluster admin may want to only update VMs that are offline.
+* Allowing user to filter VMs in ways other than by namespace; e.g. with label-selector
```
stu-gott, line 47: So this says using label selectors are an explicit non-goal? "--label-selector" is mentioned as a valid option to pass to the virtctl command on line 95.

prnaraya: Oops, this was a leftover from before. I've updated it.

iholder101:
/lgtm

Great work @prnaraya!

prnaraya:
@fabiand Would you be able to take another look? It has an lgtm now so I would like to move this forward if possible :)

prnaraya:
> We may need to update the virtctl command name

I can update the doc with the updated virtctl command/structure (`virtctl update machine-type` vs `virtctl convert-machine-type`) and other than that I think all the comments have been addressed in the implementation phases section and can be merged.

iholder101:
/lgtm

@stu-gott I don't have approval right here, but from my side we're good to go.

p.s.: perhaps we want to add some approvers to this repo?

0xFelix (approved):
Thanks, looks good to me too!

/lgtm

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,299 @@
...
+it to the latest supported version. If the Job fails or is killed, it will be
+restarted with `restartPolicy=OnFailure` in the Job spec. The `restart-vm-required`
+label may then remain on a VM that has already been shut down, so the Job will
+first check VMs with the label and confirm if the machine types have been updated.
+If they have, then the label will be removed, but if not, the label will remain
+and the VM will continue to be tracked until it is restarted and its machine
+type is updated. Both the minimum supported machine type and the target machine
+type will be determined (retrieved) and configured internally when the command is
+invoked, and these values will automatically be updated as future versions are
+released and more machine types are deprecated upon these releases.
```
xpivarc, line 80: How will we determine the correct machine type?

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,299 @@
...
+
+If the VM is running, the update will not take effect until it is restarted,
+so the label `restart-vm-required` will be applied to the VM until the VMI is
+removed (indicating the VM has been stopped), upon which the label will be cleared
+after verifying that its machine type has been successfully updated. The job will
```
xpivarc, line 85: Who is responsible for removing the label?

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,299 @@
...
+  running: false
+  template:
+    spec:
+      domain:
+        machine:
+          type: pc-q35-rhel9.2.0
+```
+where `pc-q35-rhel9.2.0` is the latest machine type version.
+
+Along with machine types being in the format `pc-q35-rhelx.x.x`, a VM may have
```
xpivarc, line 131: I am not sure all machine types are in format of  `pc-q35-rhelx.x.x`,

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,299 @@
...
+...
+spec:
+  domain:
+    machine:
+      type: q35
+status:
+  machine:
+    type: pc-q35-rhel8.2.0
+```
+and need to have its machine type updated.
```
xpivarc, line 163: Why do we need to update it if `q35` is floating? Other than migration of course...

design-proposals/mass_machine_type_transition.md
prnaraya: Currently this is being manually handled and will need some more discussion before being fully implemented in a way that retrieves the correct machine type dynamically; this will be added in a follow-up implementation phase rather than the initial implementation. However, we are currently discussing how to expand the scope of this to not be vendor specific to RHEL/CentOS Stream, so it might end up being implemented anyway while solving this issue.

design-proposals/mass_machine_type_transition.md
prnaraya: The job itself will automatically remove the label when it detects

1. The VMI has been deleted (VM is stopped or restarted)
2. The machine type in VMI status has actually been updated.

design-proposals/mass_machine_type_transition.md
prnaraya: Sorry, in this case in the proposal i was using the `pc-q35-rhelx.x.x` machine type as an example pertaining specifically to the CentOS Stream example I mentioned in the overview; however the actual implementation will not just be limited to machine types of this format/vendor specific machine types.

design-proposals/mass_machine_type_transition.md
prnaraya: In this case, the update would take place by simply just restarting the VM (which would automatically end up updating the machine type in the VMI status.machine.type. The machine type in the VM spec would remain `q35` though.

design-proposals/mass_machine_type_transition.md
xpivarc: Can we address this in the design?

design-proposals/mass_machine_type_transition.md
xpivarc: Please update.

design-proposals/mass_machine_type_transition.md
xpivarc: Please leave a note outlining this detail.

fabiand:
Please also see the updates and comments in the PR https://github.com/kubevirt/kubevirt/pull/10222#issuecomment-1681729486

design-proposals/mass_machine_type_transition.md
prnaraya: The behavior of the `q35` alias is mentioned [here](https://github.com/kubevirt/community/blob/d3938a6d2651014ed197c85d304c7e7bb57b1f6b/design-proposals/mass_machine_type_transition.md?plain=1#L132).

design-proposals/mass_machine_type_transition.md
prnaraya: I have updated the design proposal with the proposed solution for determining the machine types. Please take a look at Itamar's comment on the draft PR for more context. [kubevirt/kubevirt#10222(comment)](https://github.com/kubevirt/kubevirt/pull/10222#issuecomment-1689491133)

prnaraya:
@xpivarc @fabiand Could you please take a look and see if I've addressed all your concerns? Thanks.

design-proposals/mass_machine_type_transition.md
xpivarc: I am not quit sure that you mentioned that no action other than restart is required for this alias.

xpivarc:
@jean-edouard Can you have a look as approver?

jean-edouard:
> @jean-edouard Can you have a look as approver?

Sorry but I'm only a reviewer in this repo...

xpivarc:
@vladikr @davidvossel @rmohr any thoughts?

xpivarc:
> > @jean-edouard Can you have a look as approver?
> 
> Sorry but I'm only a reviewer in this repo...

Sorry, I remembered otherwise.

iholder101:
> > @jean-edouard Can you have a look as approver?
> 
> Sorry but I'm only a reviewer in this repo...

off topic: I think we should converge the approvers on kubevirt/kubevirt and in here. This separation does not make sense to me :)

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,265 @@
...
+
+To handle this, any running VMs with 'q35' machine type and a machine type that matches
+the regex (reported in VMI's status.machine.type) will be labelled with `restart-vm-required`:
+```yaml
+---
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+	labels: "restart-vm-required"="true"
```
rmohr, line 179: Everything in KubeVirt is normally designed to not change metadata or spec. One reason for that is to avoid reconcile loops. For example when people are hosting their objects in git and a process is periodically checking if the repo state is in sync with the cluster.

I can understand the desire to "just" selected with an efficient kubectl query the relevant VMs. This seems to be however another case where we would actually finally want field selector support on CRDs, so that we can put info like this in the status.

Thoughts by others?

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,265 @@
...
+pc-q35-rhel8.1.0     RHEL-8.1.0 PC (Q35 + ICH9, 2009) (deprecated)
+pc-q35-rhel8.0.0     RHEL-8.0.0 PC (Q35 + ICH9, 2009) (deprecated)
+pc-q35-rhel7.6.0     RHEL-7.6.0 PC (Q35 + ICH9, 2009) (deprecated)
+none                 empty machine
+```
+This means each workload must be updated to use a `pc-q35-rhel9.x.x` machine type.
+
+## Goals
+* Create an opt-in workflow that automates mass converting machine types of VMs
+	* if currently running, update VM machine type and add a label to alert the user the VM must be restarted for the change to take effect
```
rmohr, line 38: I think a fundamental issue is, that for whatever reason we are not resolving the cluster-wide machine type during VMI creation. Instead it gets added via a webhook during VM creation. This is problematic since it makes it impossible to differentiate explicit user-choices and cluster-wide decision points.

One option would be to rethink this behaviour. Nowadays the resolved machine type is also visible in the VMI status (also resolving libvirt aliases), and if we would not default the machine type, a restart would automatically pick up the latest type if unset on  the VM level, which would allow clear intent separation.

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,265 @@
+﻿# Overview
```
rmohr, line 1: one general question on this: We already have a process which is taking care of reporting and updating virt-launcher images when they are outdated. Would this just be another attribute to trigger these already established paths? Here the PR I mean: https://github.com/kubevirt/kubevirt/pull/4686

stu-gott: In this case, live migration is not going to be able to accomplish the change, since it's outright changing the machine type of the VM. A restart of qemu is unavoidable.

design-proposals/mass_machine_type_transition.md
stu-gott: I think the point of the mass machine type transition effort overall is to actually change the user requested value--as requested by the user, so in this case the approach in the design document makes sense.

design-proposals/mass_machine_type_transition.md
prnaraya: Good point, I have updated the design proposal to create and use a field in the VM Status rather than updating the VM's labels.

prnaraya:
@rmohr I've updated the design document to reflect using a field in the VM Status to track VMs pending update rather than labels. WDYT?

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,302 @@
...
+(e.g. by namespace, label-selector).
+As a cluster-admin, I want to be in control of the behavior of running VMs being
+updated; I can determine whether I want to manually restart running VMs being
+updated or let the automation restart all of them for me. 
+
+## Repos
+[kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+Create a new virtctl command `virtctl update machine-types` that will allow
```
jean-edouard, line 75: Are we planning on adding more update-able fields in the future? If not, this could probably be a single command, like `update-machine-types`.
Edit: see comment in future phases for a better idea

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,302 @@
...
+types.
+
+## User Stories
+As a cluster-admin, I want an automated way to update the machine type of all
+VMs with unsupported machine types without interrupting my workflow.
+As a cluster-admin, I want to be able to include only certain VMs to be updated
+(e.g. by namespace, label-selector).
+As a cluster-admin, I want to be in control of the behavior of running VMs being
+updated; I can determine whether I want to manually restart running VMs being
+updated or let the automation restart all of them for me.
```
jean-edouard, line 69: This needs to either be a list or have 2 spaces at the end of each story. Right now this is displayed as 1 big block.

design-proposals/mass_machine_type_transition.md
```diff
@@ -0,0 +1,302 @@
...
+## Future Phases
+As follow-ups, we will implement the following features:
+ - Dynamically retrieving the supported and deprecated QEMU machine types as a
+ subcommand to provide the user with a list of machine types to choose from. This
+ will be implemented in conjuction with the ability for the user to specify a
+ target machine type to update to; this machine type will be selected from the
+ list of supported machine types provided to the user.
+ - Subcommands that the user can use to monitor and manage the jobs. These could
+ include `virtctl update machine-types delete` which will safely terminate and
+ clean up the job and delete it, and `virtctl update machine-types list` which will allow the user to see the status of the VMs being updated by the job.
```
jean-edouard, line 296: How about:
`virtctl machine-types update`
`virtctl machine-types delete`
`virtctl machine-types list`

design-proposals/mass_machine_type_transition.md
prnaraya: Originally it was a single command, but some earlier discussion brought up a two layered approach which involved the separate `update machine-types

jean-edouard (approved):

dhiller:
@aburdenthehand there's consensus about this PR looking good. Would you mind taking a look? Thank you!

jean-edouard:
@aburdenthehand I think we should make @dhiller an approver on this repo.

aburdenthehand:
/approve
Sounds good to me

PR #226, opened 2023-07-12
WIP instancetype: Add ControllerRevision migration proposal

lyarwood:


design-proposals/instancetype-controllerrevision-migration.md
```diff
@@ -0,0 +1,189 @@
...
+* Currently a `ControllerRevision` is capture for each `VirtualMachine` and
+  combination or instance type or preference. The deduplication of these
+  `ControllerRevisions` is saved for follow up work but might be an additional
+  step when migrating to a new `ControllerRevision` using labels to identify
+  existing `ControllerRevisions` containing the same version and importantly
+  generation of an object.
+
+## User Stories
+
+* As a developer I want deprecated versions of `instancetype.kubevirt.io` to
```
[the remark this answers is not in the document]
lyarwood: I'd say that's secondary to lowering the maintenance overhead, the user stories don't make sense without stating why we are removing things IMHO.

xpivarc:
Could we just handle the revision upgrades on the periodical/triggered resync of VM?

lyarwood:
> Could we just handle the revision upgrades on the periodical/triggered resync of VM?

I'll post a respin shortly and will include my thoughts on this by ultimately I think only doing this work as part of a VM resync has a number of issues:

- Triggering an upgrade requires a resync of the `VirtualMachine`, something that isn't guaranteed between upgrades for offline `VirtualMachines` and thus isn't guaranteed to happen between the time that functionality is introduced and the time we drop the older version support
- It's transparent to the user, operator and vendor support team outside of logs in the `virt-controller` and thus a pain to debug if/when something goes wrong
- It's also difficult to tell when all of the `VirtuialMachines` have been updated and we are thus ready to upgrade to a version of KubeVirt dropping support for the older `instancetype.kubevirt.io` versions. Have a single `Trigger` CRD or similar would give us that assurance that we are ready to go.

Happy to use the resync as one way of triggering the upgrade but I still think we need a standalone trigger as well.

xpivarc:
> > Could we just handle the revision upgrades on the periodical/triggered resync of VM?
> 
> I'll post a respin shortly and will include my thoughts on this by ultimately I think only doing this work as part of a VM resync has a number of issues:
> 
>     * Triggering an upgrade requires a resync of the `VirtualMachine`, something that isn't guaranteed between upgrades for offline `VirtualMachines` and thus isn't guaranteed to happen between the time that functionality is introduced and the time we drop the older version support

Resync is guaranteed. On upgrades, a controller will enqueue all VMs, without upgrading a periodic 12/24h resync.

> 
>     * It's transparent to the user, operator and vendor support team outside of logs in the `virt-controller` and thus a pain to debug if/when something goes wrong

I can't follow what you mean by this point, could you rephrase?

> 
>     * It's also difficult to tell when all of the `VirtuialMachines` have been updated and we are thus ready to upgrade to a version of KubeVirt dropping support for the older `instancetype.kubevirt.io` versions. Having a single `Trigger` CRD or similar would give us that assurance that we are ready to go

I did not read through this proposal but I would say you could achieve this also without new CRD.

> 
> 
> Happy to use the resync as one way of triggering the upgrade but I still think we need a standalone trigger as well.

lyarwood:
> > > Could we just handle the revision upgrades on the periodical/triggered resync of VM?
> > 
> > I'll post a respin shortly and will include my thoughts on this by ultimately I think only doing this work as part of a VM resync has a number of issues:
> > ```
> > * Triggering an upgrade requires a resync of the `VirtualMachine`, something that isn't guaranteed between upgrades for offline `VirtualMachines` and thus isn't guaranteed to happen between the time that functionality is introduced and the time we drop the older version support
> > ```
> 
> Resync is guaranteed. On upgrades, a controller will enqueue all VMs, without upgrading a periodic 12/24h resync.

Apologies I wasn't aware of this, any code or doc links?

> > ```
> > * It's transparent to the user, operator and vendor support team outside of logs in the `virt-controller` and thus a pain to debug if/when something goes wrong
> > ```
> 
> I can't follow what you mean by this point, could you rephrase?

The additional CRs just allow us to lookup if an individual or set of upgrades have succeeded and if not they should make it easy to identify why an upgrade failed. Without these we'd end up in a situation where we would either need to add this detail to the VirtualMachine status or dive into the controller logs.

> > ```
> > * It's also difficult to tell when all of the `VirtuialMachines` have been updated and we are thus ready to upgrade to a version of KubeVirt dropping support for the older `instancetype.kubevirt.io` versions. Having a single `Trigger` CRD or similar would give us that assurance that we are ready to go
> > ```
> 
> I did not read through this proposal but I would say you could achieve this also without new CRD.

Possibly with labels on the `VirtualMachine` but thus far I've been targeting `ControllerRevisions` that make it awkward to track progress without something like this IMHO.

xpivarc:
>The additional CRs just allow us to lookup if an individual or set of upgrades have succeeded and if not they should make it easy to identify why an upgrade failed. Without these we'd end up in a situation where we would either need to add this detail to the VirtualMachine status or dive into the controller logs.

Yes, but the work remains the same. We can simply reflect label/annotation on Kubevirt CR

design-proposals/instancetype-controllerrevision-migration.md
```diff
@@ -0,0 +1,228 @@
...
+  older deprecated versions to eventually be removed.
+
+## Non Goals
+
+* Currently a `ControllerRevision` is captured for each `VirtualMachine` and
+  combination or instance type or preference. The deduplication of these
+  `ControllerRevisions` is saved for follow up work but might be an additional
+  step when migrating to a new `ControllerRevision` using labels to identify
+  existing `ControllerRevisions` containing the same version and importantly
+  generation of an object.
```
xpivarc, line 51: +1

design-proposals/instancetype-controllerrevision-migration.md
```diff
@@ -0,0 +1,228 @@
...
+
+For
+
+* This would avoid the need to add additional CRDs and controllers to the codebase
+
+* `VirtualMachines` could be labelled to highlight that no upgrade was required *or* the upgrade was complete?
+
+Against
+
+* Difficult to tell when all `VirtualMachines` had been upgraded without additional labels
```
xpivarc, line 219: Why?

design-proposals/instancetype-controllerrevision-migration.md
```diff
@@ -0,0 +1,228 @@
...
+
+* `VirtualMachineStatus` would need to be extended to capture failures
```
xpivarc, line 221: Why?

design-proposals/instancetype-controllerrevision-migration.md
```diff
@@ -0,0 +1,228 @@
...
+
+* This would add overhead to each resync of a VirtualMachine using
```
xpivarc, line 223: This could be handled by a new controller but without the need for CRD as well if we find this argument appealing.

design-proposals/instancetype-controllerrevision-migration.md
```diff
@@ -0,0 +1,228 @@
...
+  `instancetype.kubevirt.io` ControllerRevisions
+
+* This would also possibly overload the VirtualMachine controller on upgrade if
```
xpivarc, line 226: Don't see how it differs with original proposal

xpivarc (commented):
Reviewed the alternative.

PR #227, opened 2023-07-14
Guest console logs design proposal

tiraboschi:
Add a design proposal to stream guest
console logs and collect them with the
standard Kubernetes log architecture.

### DIFF
--- a/design-proposals/vm-console-logs.md
+++ b/design-proposals/vm-console-logs.md
@@ -1,0 +1,784 @@
+# Overview
+Linux and Windows Server VMs can log boot and application data to their serial console.
+KubeVirt user can great benefit from having VMs logs, collected over the serial console, integrated into the Kubernetes logging architecture. 
+
+## Motivation
+VM logs streamed over the serial console contain really useful debug/troubleshooting info.
+While we are currently already capturing KubeVirt components logs (virt-controller, virt-launcher, libvirt, virt-handler, etc…) but the VM guest OS and application logs are not currently captured or indexed using cluster logging:
+currently KubeVirt lacks the level of integration that allows VM guests logs to be seamlessly streamed and captured as for generic container workloads ones.
+
+A similar feature al been somehow already required more than once at community level, (see https://groups.google.com/g/kubevirt-dev/c/qBdboekQ_Vk for instance).
+And a similar feature is a pretty common pattern across all the most popular HyperScalers, see as examples:
+- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-console.html#instance-console-console-output
+- https://learn.microsoft.com/en-us/azure/virtual-machines/boot-diagnostics
+- https://cloud.google.com/compute/docs/troubleshooting/viewing-serial-port-output#enable-stackdriver
+- ...
+
+## Goals
+- It should let the user collect serial console logs for his VM according to the standard k8s logging architecture: the main benefit of this choice is that it will allow it to be directly consumed by all the k8s logging aggregation solutions without the need for further customizations.
+- No guest OS sided changes should be required.
+- It should not prevent the usage of the serial console to interactively login into the VM.
+
+## Non Goals
+- VM serial console logs could be streamed to external logging stacks like Fluentd+Elasticsearch+Kibana/Loki/... but this specific part of the integration is out of scope for this proposal.
+- Defining custom/new format/structured for guest OS logs is out of scope. 
+
+## Definition of Users
+- VM owner: the user who owns a VM in his namespace on a Kubernetes cluster with KubeVirt
+- cluster-admin: the administrator of the cluster
+
+## User Stories
+- As a VM owner I want to view the VM guest OS logs using standard client tooling just as I would an application running in a pod.
+- As a VM user I want to query my logs pertaining to an application running within my VM workload using Kubernetes standard logging
+- As a cluster-admin, I would like to view all the logs of all the VMs running in my cluster in order to find and correct reoccurring bugs, security flaws or misconfigurations.
+
+## Repos
+- https://github.com/kubevirt/kubevirt
+
+# Design
+In [Libvirt the serial port devices](https://libvirt.org/formatdomain.html#console) can already be configured via a log sub-element, with a file attribute.
+```
+<log file="/var/run/kubevirt-private/b802512e-d13e-4592-83b4-db153293b666/virt-serial0-log" append="on"/>
+```
+that translates to something like:
+```
+-chardev socket,id=charserial0,fd=17,server=on,wait=off,logfile=/dev/fdset/0,logappend=on -device {"driver":"isa-serial","chardev":"charserial0","id":"serial0","index":0}
+```
+on the QEMU side.
+
+This means that we can add the <log/> element like:
+```yaml
+    <serial type='unix'>
+      <source mode='bind' path='/var/run/kubevirt-private/b802512e-d13e-4592-83b4-db153293b666/virt-serial0'/>
+--->  <log file='/var/run/kubevirt-private/b802512e-d13e-4592-83b4-db153293b666/virt-serial0-log' append='on'/>
+      <target type='isa-serial' port='0'>
+        <model name='isa-serial'/>
+      </target>
+      <alias name='serial0'/>
+    </serial>
+    <console type='unix'>
+      <source mode='bind' path='/var/run/kubevirt-private/b802512e-d13e-4592-83b4-db153293b666/virt-serial0'/>
+--->  <log file='/var/run/kubevirt-private/b802512e-d13e-4592-83b4-db153293b666/virt-serial0-log' append='on'/>
+      <target type='serial' port='0'/>
+      <alias name='serial0'/>
+    </console>
+```
+to the serial and the console entries in the libvirt XML for the KubeVirt VMs.
+
+Those log files are going to be rotated by `virtlogd` exactly as for libvirt logs so file size is not an issue as it's not for libvirt/qemu ones.
+
+Now that we have the guest console logs in a file on the kubevirt-private volume in the pod, we just need another container in the pod to stream them out as for generic container logs.
+This is exactly the standard and well adopted [Kubernertes Streaming sidecar container logging architecture](https://kubernetes.io/docs/concepts/cluster-administration/logging/#streaming-sidecar-container).
+
+So in the pod we should have another (optional) container named `guest-console-log` and defined as:
+```yaml
+  - args:
+       - --logfile
+       - /var/run/kubevirt-private/bf9afcc1-fe55-4f63-919a-807b59efff76/virt-serial0-log
+    command:
+       - /usr/bin/virt-tail
+    env:
+       - name: VIRT_LAUNCHER_LOG_VERBOSITY
+         value: "2"
+    image: registry:5000/kubevirt/virt-launcher@sha256:fb1c5e6656501a6af8445236106eaa9d0dc242f1f5c1783db4a8a7895dfc75e5
+    imagePullPolicy: IfNotPresent
+    name: guest-console-log
+    resources:
+       limits:
+          cpu: 15m
+          memory: 60M
+       requests:
+          cpu: 5m
+          memory: 35M
+    securityContext:
+       allowPrivilegeEscalation: false
+       capabilities:
+          drop:
+             - ALL
+       runAsNonRoot: true
+       runAsUser: 107
+    terminationMessagePath: /dev/termination-log
+    terminationMessagePolicy: File
+    volumeMounts:
+       - mountPath: /var/run/kubevirt-private
+         name: private
+         readOnly: true
+```
+
+Ideally the `tail` binary could potentially be enough to stream a log file to the `stdout` in the log streaming container
+but the relationship with other container in the `virt-launcher` pod, especially in the shutdown/cleanup phase is critical.
+
+If not all the containers of the virt-launcher pod are correctly going to terminate together, we could potentially risk partially shut-down virt-launcher pods, blocking migrations and reaching a clear Terminated state.
+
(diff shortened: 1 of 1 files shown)


machadovilaca:
+1

design-proposals/vm-console-logs.md
```diff
@@ -0,0 +1,743 @@
...
+      domain:
+        devices:
+          autoattachSerialConsole: true
+          logSerialConsole: true
+...
+```
+
+We can also introduce a cluster wide setting for the cluster-admin to let him tune the default behavior when nothing is specified at VM level.  
+
+## Scalability
```
EdDev, line 659: Is there an I/O and CPU cost to this? Especially when enabled on a large cluster (for cluster level configuration as suggested earlier)

design-proposals/vm-console-logs.md
```diff
@@ -0,0 +1,743 @@
+# Overview
+Linux and Windows Server VMs can log boot and application data to their serial console.
+KubeVirt user can great benefit from having VMs logs, collected over the serial console, integrated into the Kubernetes logging architecture. 
+
+## Motivation
```
EdDev, line 5: - Are there similar solutions for other VM management systems?
- Were there specific requests for something like this from users?

design-proposals/vm-console-logs.md
```diff
@@ -0,0 +1,743 @@
...
+
+## User Stories
+- As a VM owner I want to view the VM guest OS logs using standard client tooling just as I would an application running in a container.
+- As a VM user I want to query my logs pertaining to an application running within my VM workload using Kubernetes standard logging
+- As a cluster-admin, I would like to view all the logs of all the VMs running in my cluster in order to find and correct reoccurring bugs, security flaws or misconfigurations.
+
+## Repos
+- https://github.com/kubevirt/kubevirt
+
+# Design
```
EdDev, line 31: How will this work with migration?

The example below focuses on the virt-launcher pod, but this is not a common KubeVirt user API.
A VM owner may not even have access to pods (I would be surprise if he had).
Can the VM object be the one the user interact with? If it can resolve the right pod in the background, it would be very nice.

EdDev (commented):
This is very nice, thanks!
When debugging, it can be very useful to access these logs.

/cc

design-proposals/vm-console-logs.md
tiraboschi: > * Are there similar solutions for other VM management systems?

This is a pretty common pattern across all the most popular HyperScalers, see as examples:
- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-console.html#instance-console-console-output
- https://learn.microsoft.com/en-us/azure/virtual-machines/boot-diagnostics
- https://cloud.google.com/compute/docs/troubleshooting/viewing-serial-port-output#enable-stackdriver
...

> * Were there specific requests for something like this from users?

yes, https://groups.google.com/g/kubevirt-dev/c/qBdboekQ_Vk for instance

design-proposals/vm-console-logs.md
tiraboschi: It really depends on how much each VM will write to its serial console (but this is not so different than for a generic application that produces logs on its stdout).
I'll try to get some numbers for popular guest images.

design-proposals/vm-console-logs.md
EdDev: Great, thank you!
It will be valuable to add this to the design. I guess it is a sub-section of the motivation.

design-proposals/vm-console-logs.md
EdDev: > Exactly as for libvirt logs that are streamed in the logs of the compute container

The guest logs focus on the guest while the libvirt (and virt-launcher) ones focus on the infrastructure that runs it. The fact that a migration translates into two pods is an implementation detail, same as libvirt and having a pod is an implementation detail.
For VM as a frontend to users, we usually provide an abstraction so they will not need to care about the backend details.

> If the user wants to get all the logs he has to setup a log collection and aggregation system as for a generic application and query it by VM name.

IMO, the minimum is to provide the generic instructions of how users can work with it (e.g. track the pods that are used). A better solution is to provide an abstraction and not burden a user with this, but I recognize this may be a follow up and harder to implement.

> The whole idea behind Kubevirt is to be able to have a VM natively interact with regular pods in an hybrid scenario so I'd expect an user to be able to start VMs and regular pods

I think there is a distinction between working on the same platform (VMs and Pods) and having access to these resources. Users that create VM/s may be given access to create and modify the VM CR, but not access to control the low level Pod. Accessing pods is a pretty privileged operation, as it allows one to run whatever they want on the platform/cluster.
Usually abstractions are given, limiting what can be run.

But you could claim that the user in question is the project admin (i.e. the namespace admin) which has the highest priv in the namespace, and has access to pods and SAs.

If the user is just the one that owns a specific VM, or has access only to VM abstractions (the VM CR), then accessing pods should not be an option for them.
On the other hand, if the user in question is the project admin, then this will make more sense.

As long as this is clear from the design point of view and expressed in the documentation, I think the limitations here are acceptable. If we will need to enhance the solution in the future, abstractions can be build over this base.

design-proposals/vm-console-logs.md
```diff
@@ -0,0 +1,743 @@
...
+   2. virt-launcher pod should not contain a container named `guest-console-log`.
+   3. a file named `/var/run/kubevirt-private/<uuid>/virt-serial0-log` should not be present in the Kubevirt private volume.
+
+### Positive flow
+
+1. Start a VM with `/spec/template/spec/domain/devices/logSerialConsole=true`
+2. Wait for the VM to be running
+3. Check:
+    1. The first serial console and the console device in the libvirt XML should contain the <log/> element.
+    2. virt-launcher pod should not a container named `guest-console-log`.
```
orenc1, line 735:
```suggestion
    2. virt-launcher pod should contain a container named `guest-console-log`.
```

design-proposals/vm-console-logs.md
tiraboschi: > The guest logs focus on the guest while the libvirt (and virt-launcher) ones focus on the infrastructure that runs it. The fact that a migration translates into two pods is an implementation detail, same as libvirt and having a pod is an implementation detail.
For VM as a frontend to users, we usually provide an abstraction so they will not need to care about the backend details.

The main idea behind this is to be 100% tight to the [standard k8s logging architecture](https://kubernetes.io/docs/concepts/cluster-administration/logging/) to get the expected integration with all the logs aggregating solutions (like the ELK stack, Loki, ...) for free.

> IMO, the minimum is to provide the generic instructions of how users can work with it (e.g. track the pods that are used). A better solution is to provide an abstraction and not burden a user with this, but I recognize this may be a follow up and harder to implement.

virt-launcher pods are already labeled with something like:
```yaml
  labels:
    kubevirt.io: virt-launcher
    kubevirt.io/domain: rhel9-above-urial
    vm.kubevirt.io/name: rhel9-above-urial
```
this is only about using the right labels when querying the log management tool.

design-proposals/vm-console-logs.md
```diff
@@ -0,0 +1,743 @@
...
+logSerialConsole:
+  description: Whether to log the auto-attached default serial
+    console or not. Serial console logs will be collect to
+    a file and then streamed from a named 'guest-console-log'. Not
+    relevant if autoattachSerialConsole is disabled. Defaults
+    to false.
+  type: boolean
+```
+to let the VM owner require the serial console to be logged.
+It should be off by default since serial console logs could potentially disclosure sensitive data.
```
iholder101, line 637: +1
Might also negatively affect performance

design-proposals/vm-console-logs.md
```diff
@@ -0,0 +1,743 @@
...
+    <console type='unix'>
+      <source mode='bind' path='/var/run/kubevirt-private/b802512e-d13e-4592-83b4-db153293b666/virt-serial0'/>
+--->  <log file='/var/run/kubevirt-private/b802512e-d13e-4592-83b4-db153293b666/virt-serial0-log' append='on'/>
+      <target type='serial' port='0'/>
+      <alias name='serial0'/>
+    </console>
+```
+to the serial and the console entries in the libvirt XML for the KubeVirt VMs.
+
+Those log files are going to be rotated by `virtlogd` exactly as for libvirt logs so file size is not an issue as it's not for libvirt/qemu ones.
```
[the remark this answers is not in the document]
iholder101: Oh now I see that you write below it's 8MB. Thanks!

iholder101 (commented):
Thank you @tiraboschi! Looks great and very valuable!
/cc

design-proposals/vm-console-logs.md
iholder101: TBH I think I agree with @EdDev's point here.

Kubevirt's premise is that users would be able to run VMs alongside "regular" pods. However, the fact that a VM runs inside a pod is indeed an implementation detail. The user doesn't have to know or care about the virt-launcher pod itself and as of today, there's no reason to interact with this pod to inspect the guest itself, but only in order to inspect the virtualization stack that's running the guest. All of the user's interaction with the VM is done by the VirtualMachine CRD which is the user-facing k8s object.

However, implementation-wise I'm not sure if it's possible / easy to redirect the logs to the VM object itself. Is it common to perform operations like `kubectl logs vm my-vm` on other CRDs? How are other projects coping with this?

Perhaps as a compromise we can hide the implementation from the user by introducing a `virtctl log my-vm` command that would know to fetch this logs without complicating the users with the need to understand that the VM runs inside a pod, and that there's a side-car container that actually fetches these logs, etc.

WDYT?

tiraboschi: > However, implementation-wise I'm not sure if it's possible / easy to redirect the logs to the VM object itself. Is it common to perform operations like `kubectl logs vm my-vm` on other CRDs? How are other projects coping with this?

As for:
https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#advanced-features-and-flexibility
`Other Subresources - Add operations other than CRUD, such as "logs" or "exec".` is not supported on CRDs.

> Perhaps as a compromise we can hide the implementation from the user by introducing a `virtctl log my-vm` command that would know to fetch this logs without complicating the users with the need to understand that the VM runs inside a pod, and that there's a side-car container that actually fetches these logs, etc.
> 
> WDYT?

Introducing `virtctl logs -n <vmnamespace> <vmname>` as wrapper for `kubectl logs -n <vmnamespace> -l vm.kubevirt.io/name=<vmname>` is probably straightforward but I want also to mention that interactively fetching the logs of a running VM is just one of the possible use-cases. 
I'd expect that being able to stream the guest console of the VMs on the cluster to an external logging stack (like [Grafana Loki](https://grafana.com/oss/loki/)) to store, analyze, and query it there it will become pretty common.
All the k8s logging stack are designed to natively manage and filter pod logs according to the standard k8s logging architecture, if we design something significantly different it will risk to require further effort to be integrated into other tools.
Currently with this approach we can easily get something like this in Loki basically out of the box:
[guest console log in Grafana Loki](https://pasteboard.co/khxx1VDKhQUN.png)

design-proposals/vm-console-logs.md
```diff
@@ -0,0 +1,743 @@
...
+    image: registry:5000/kubevirt/virt-launcher@sha256:6cd10c0d6c634b8d5619700058920549fb501dbb7e97cc132047fafdc6099914
+    imagePullPolicy: IfNotPresent
+    name: guest-console-log
+    resources:
+      limits:
+        cpu: 100m
+        memory: 80M
+      requests:
+        cpu: 10m
+        memory: 1M
```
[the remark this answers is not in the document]
tiraboschi: Just an example, it has to be properly tuned.

design-proposals/vm-console-logs.md
tiraboschi: on https://github.com/kubevirt/kubevirt/pull/10110 I'm now trying to have it on by default for all the VMs and see if we can have it successfully passing the whole test suite with no evident regressions.

design-proposals/vm-console-logs.md
tiraboschi: done

design-proposals/vm-console-logs.md
```diff
@@ -0,0 +1,743 @@
...
+    - mountPath: /var/run/kubevirt-private
+      name: private
+      readOnly: true
+```
+
+Since we don't need any custom binary or script, we can simply reuse the virt-launcher image which already contains the `tail` binary.
+For security reason the virt-launcher `private` volume that contains the log can be mounted in read only mode.
+The tail command will be executed with options:
+- `--quiet` to not output tail headers (file name...) into the logs.
+- `-n+1` to avoid duplicating log lines on process restart.
```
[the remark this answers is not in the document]
tiraboschi: moving to custom tail binary

design-proposals/vm-console-logs.md
tiraboschi: done

design-proposals/vm-console-logs.md
tiraboschi: for a small centos-stream8 VM printing 10 log lines every second I see:
```bash
$ kubectl  top pod -n simone virt-launcher-centos-stream8-8v46h --containers --use-protocol-buffers
POD                                  NAME                CPU(cores)   MEMORY(bytes)   
virt-launcher-centos-stream8-8v46h   POD                 0m           0Mi             
virt-launcher-centos-stream8-8v46h   compute             30m          1205Mi          
virt-launcher-centos-stream8-8v46h   guest-console-log   7m           29Mi
```

design-proposals/vm-console-logs.md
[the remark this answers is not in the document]
tiraboschi: done

PR #231, opened 2023-07-26
Add virt-operator deployment of common-instancetypes proposal

lyarwood:
The `common-instancetypes` project provides a set of useful instance types and preferences for use when creating `VirtualMachines` within KubeVirt. Until now operators and users could manually deploy these resources from the upstream project tree, from a specific release or by using the SSP operator.

This proposal suggests moving the deployment of the cluster-wide `common-instancetypes` resources to `virt-operator` so that they are consistently available across KubeVirt deployments.

### DIFF
--- a/design-proposals/virt-operator-deploy-instancetypes.md
+++ b/design-proposals/virt-operator-deploy-instancetypes.md
@@ -1,0 +1,139 @@
+# Overview
+
+The [`kubevirt/common-instancetypes`](https://github.com/kubevirt/common-instancetypes)
+project provides a set of useful [instance types and preferences](https://kubevirt.io/user-guide/virtual_machines/instancetypes/) for
+use when creating
+[`VirtualMachines`](http://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachine)
+with KubeVirt. Until now operators and users could manually deploy these
+resources from the [upstream project
+tree](https://github.com/kubevirt/common-instancetypes#installation), from a
+[specific release](https://github.com/kubevirt/common-instancetypes/releases/tag/v0.3.1)
+or by using the [SSP operator](https://github.com/kubevirt/ssp-operator#ssp-operator).
+
+This proposal suggests moving the deployment of the cluster-wide
+`kubevirt/common-instancetypes` resources to `virt-operator` so that they are
+consistently available across KubeVirt deployments.
+
+## Motivation
+
+The ultimate goal of this proposal is to improve KubeVirt's default user
+experience when creating `VirtualMachines` across several different deployments.
+
+The deployment of the cluster-wide `kubevirt/common-instancetypes` resources by
+default will allow users to rely more freely on the presence of the resources
+and make use of them during their `VirtualMachine` creation flows.
+
+For example, thanks to recent improvements within `virtctl` the following
+command would deploy a simple Fedora `VirtualMachine` work across several
+different deployments:
+
+```sh
+$ virtctl create vm \
+  --instancetype n1.medium \
+  --preference fedora \
+  --volume-containerdisk name:fedora,src:quay.io/containerdisks/fedora:latest \
+  --name fedora | kubectl apply -f -
+```
+
+## Goals
+
+* For new deployments, the deployment of the cluster-wide
+  `kubevirt/common-instancetypes` resources by `virt-operator`
+
+* For upgraded deployments, to have `virt-operator` claim ownership of any
+  existing cluster-wide `kubevirt/common-instancetypes` resources deployed by
+  the SSP operator.
+
+## Non Goals
+
+* For upgraded deployments, `virt-operator` will not try to claim ownership of
+  any operator deployed cluster-wide `kubevirt/common-instancetypes` resources
+  other than by SSP.
+
+## User Stories
+
+* As a user I want access to a consistent default set of cluster-wide instance
+  types and preferences across all deployments I interact with
+
+* As an operator I want my users to have access to a consistent set of
+  cluster-wide instance types and preferences without the need for additional
+  deployment of operators or resources
+
+## Repos
+
+* kubevirt/kubevirt
+* kubevirt/common-instancetypes
+* kubevirt/ssp-operator
+
+# Design
+
+`virt-operator` will be extended to deploy bundles of cluster-wide resources by
+default from the `kubevirt/common-instancetypes` project.
+
+These bundles will be injected into the `virt-operator` container image at build
+time with the `kubevirt/common-instancetypes` project referenced as a submodule
+checked out from a specific release tag from within `kubevirt/kubevirt`.
+
+The submodule approach keeps much of the maintenance and testing cruft of the
+`kubevirt/common-instancetypes` project out of the already congested
+`kubevirt/kubevirt` project. It also allows us to retain the existing
+`kustomize` project structure of `kubevirt/common-instancetypes` making the repo
+useful for end operators and users looking to make modifications to or to only
+generate a subset of the resources.
+
+The `KubeVirt` `CR` will be extended to include a `deployInstancetypes`
+toggle, enabled by default, to allow operators to disable this behaviour.
+
+An additional toggle will also be introduced into the
+`kubevirt/hyperconverged-cluster-operator` project to control this behaviour
+when deploying `kubevirt/kubevirt`.
+
+The `kubevirt/ssp-operator` project will also be modified to ignore
+`kubevirt/common-instancetypes` resources deployed by the `virt-operator` in
+version `v0.18.N` and later to stop deploying `kubevirt/common-instancetypes`
+resources at all by default with version(s) >=`v0.19.0`.
+
+## Scalability
+
+The `kubevirt/common-instancetypes` project currently ships instance type and
+preference cluster-wide resources. Their deployment by default should not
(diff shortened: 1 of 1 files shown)


design-proposals/virt-operator-deploy-instancetypes.md
```diff
@@ -0,0 +1,97 @@
+# Overview
+
+The [`common-instancetypes`](https://github.com/kubevirt/common-instancetypes)
+project provides a set of useful [instance types and preferences](https://kubevirt.io/user-guide/virtual_machines/instancetypes/) for
```
0xFelix, line 4: `s/useful/pre-defined/`?

design-proposals/virt-operator-deploy-instancetypes.md
```diff
@@ -0,0 +1,97 @@
...
+This proposal suggests moving the deployment of the cluster-wide
+`common-instancetypes` resources to `virt-operator` so that they are
+consistently available across KubeVirt deployments.
+
+## Motivation
+
+The ultimate goal of this proposal is to improve KubeVirt's default user
+experience when creating `VirtualMachines` across several different deployments.
+
+The deployment by default of the cluster-wide `common-instancetypes` resources
```
0xFelix, line 22:
```suggestion
The deployment of the cluster-wide `common-instancetypes` resources by default
```

design-proposals/virt-operator-deploy-instancetypes.md
```diff
@@ -0,0 +1,97 @@
...
+* kubevirt/kubevirt
+* kubevirt/common-instancetypes
+* kubevirt/ssp-operator
+
+# Design
+
+**TODO** Add more context here once some draft code has been written
+
+`virt-operator` will be extended to by default deploy bundles of cluster-wide
+resources from the `common-instancetypes` project.
```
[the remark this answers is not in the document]
lyarwood: @davidvossel just following up, would a git sub-module based approach be acceptable or do you want the full tree imported?

design-proposals/virt-operator-deploy-instancetypes.md
fabiand: FWIW I like useful, because that is what they are.
It's a matter of fact that they are predfeined.

design-proposals/virt-operator-deploy-instancetypes.md
```diff
@@ -0,0 +1,97 @@
...
+
+This proposal suggests moving the deployment of the cluster-wide
+`common-instancetypes` resources to `virt-operator` so that they are
+consistently available across KubeVirt deployments.
+
+## Motivation
+
+The ultimate goal of this proposal is to improve KubeVirt's default user
+experience when creating `VirtualMachines` across several different deployments.
+
```
fabiand, line 21: Just to underline the usability enhancement, can we add the virtctl cmdline that illustrates the creation of a vm with IT and Prefs preinstalled?

design-proposals/virt-operator-deploy-instancetypes.md
```diff
@@ -0,0 +1,97 @@
...
+The deployment by default of the cluster-wide `common-instancetypes` resources
+will allow users to rely more freely on the presence of the resources and make
+use of them during their `VirtualMachine` creation flows.
+
+## Goals
+
+* For new deployments, the deployment of the cluster-wide `common-instancetypes`
+  resources by `virt-operator`
```
fabiand, line 29: Can an admin opt-out of this?

design-proposals/virt-operator-deploy-instancetypes.md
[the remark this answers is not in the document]
fabiand: I'm not sure if a submodule approach will fly. Mainly because /api is a reverse-submodule approach (we export from kubevirt/kubevirt to a dedicate repo).

The biggest benefit of having (some) ITs+prefs within kubevirt/kuevirt is testing.
But to recude the impact, we should slim down the default set of ITs+Prefs, this would then leave us with a "supplemental" set of ITs+Prefs.

FWIW if we inlcude them in kubevirt/kubevirt, then they should be really flat. To avoid adding to much magic to this repository.

design-proposals/virt-operator-deploy-instancetypes.md
```diff
@@ -0,0 +1,97 @@
...
+the SSP operator. The SSP operator itself will need to be updated at the same
+time to ensure it relinquishes ownership and no longer attempts to reconcile the
+resources.
+
+Rolling both `virt-operator` and SSP operator back to their previous versions
+should return control of these resources to the SSP operator.
+
+## Functional Testing Approach
+
+TBD
```
fabiand, line 87: Unit tests for all prefs+its
One e2e test per pref.
Pref is more sensible than ITs, thus I do not think that we have to test the full pref vs IT matrix.

design-proposals/virt-operator-deploy-instancetypes.md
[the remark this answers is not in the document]
lyarwood: > I'm not sure if a submodule approach will fly. Mainly because /api is a reverse-submodule approach (we export from kubevirt/kubevirt to a dedicate repo).

I don't really follow what the objection would be to the submodule approach then given what we do with `kubevirt/api`?

> The biggest benefit of having (some) ITs+prefs within kubevirt/kuevirt is testing. But to recude the impact, we should slim down the default set of ITs+Prefs, this would then leave us with a "supplemental" set of ITs+Prefs.

I think you mean the biggest drawback right? Having the full tree imported into `kubevirt/kubevirt` moves the burden of testing, bug and PR tracking into that repo.

> FWIW if we inlcude them in kubevirt/kubevirt, then they should be really flat. To avoid adding to much magic to this repository.

I'd rather keep the projects separated, I still think the structure of the `kubevirt/common-instancetypes` repo would be useful for certain end users looking to build on the resources we provide there.

design-proposals/virt-operator-deploy-instancetypes.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Goals
+
+* For new deployments, the deployment of the cluster-wide `common-instancetypes`
+  resources by `virt-operator`
+
+* For upgraded deployments, to have `virt-operator` claim ownership of any
+  existing cluster-wide `common-instancetypes` resources deployed by the SSP
+  operator.
+
```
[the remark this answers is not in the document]
lyarwood: I'm not sure that's a great idea, that's starting to suggest we also define the CRDs within the project IMHO.

design-proposals/virt-operator-deploy-instancetypes.md
lyarwood: Yes sorry I didn't include this in the design proposal but I want this configurable from the KV CR.

design-proposals/virt-operator-deploy-instancetypes.md
lyarwood: As I said before merging even a subset of the `common-instancetypes` tree into `kubevirt/kubevirt` migrates the maintenance and testing burden into an already busy repo. Using a submodule approach would allow us to focus on the specific fresh deployment and upgrade scenarios here.

design-proposals/virt-operator-deploy-instancetypes.md
```diff
@@ -0,0 +1,97 @@
...
+**TODO** Add more context here once some draft code has been written
+
+`virt-operator` will be extended to by default deploy bundles of cluster-wide
+resources from the `common-instancetypes` project.
+
+**TODO** Should we also support deploying resources from a kustomize URI as is
+         supported by SSP Operator?
+
+These bundles will be sync'd into the KubeVirt tree during the release of a
```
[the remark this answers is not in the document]
lyarwood: I don't think that really makes it any easier tbh, importing the bundles during the image build either through a submodule or direct download should be enough to get us started. If we needed a way to switch between bundles quickly then the kustomize URI approach would be the way forward IMHO.

design-proposals/virt-operator-deploy-instancetypes.md
```diff
@@ -0,0 +1,97 @@
...
+
+TBD
+
+# Implementation Phases
+
+TBD
+
+# Alternatives
+
+* Creation of a standalone `instancetype-operator` using
```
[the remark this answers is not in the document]
lyarwood: Yup just a suggested alternative and not necessarily desirable

design-proposals/virt-operator-deploy-instancetypes.md
```diff
@@ -0,0 +1,139 @@
...
+time with the `kubevirt/common-instancetypes` project referenced as a submodule
+checked out from a specific release tag from within `kubevirt/kubevirt`.
+
+The submodule approach keeps much of the maintenance and testing cruft of the
+`kubevirt/common-instancetypes` project out of the already congested
+`kubevirt/kubevirt` project. It also allows us to retain the existing
+`kustomize` project structure of `kubevirt/common-instancetypes` making the repo
+useful for end operators and users looking to make modifications to or to only
+generate a subset of the resources.
+
```
lyarwood, line 83: @davidvossel @fabiand I've tried to set out my reasoning for using a sub-module above, let me know if this is still an issue and I can think about reworking things into a flat import of these resources into `kubevirt/kubevirt`.

PR #240, opened 2023-08-17
Proposal for local storage migration

alicefr:
Reference PR for migrating local storage in KubeVirt: https://github.com/kubevirt/kubevirt/pull/10305


design-proposals/local-storage-migration/local-storage-migration.md
```diff
@@ -0,0 +1,261 @@
...
+  1. Switch storage class. For example, if the user wants to
+     change storage type, like in case the new storage provider provides
+     additional functionalities.
+  2. Move to a larger disk if the storage class doesn't support volume expansion
+  3. Move from container disks to PVC (this can be considered as a slightly
+     different flavor of the [Persistent containerDisk volumes](https://github.com/kubevirt/kubevirt/issues/6086))
+
+## Goals
+
+Design KubeVirt APIs for introducing non-shared storage migration.
```
fabiand, line 26: Should the goal be an api or to solve a problem? *hint*

fabiand (commented):
Nice to see this proposal.

design-proposals/local-storage-migration/local-storage-migration.md
alicefr: Well, both :)... I'll rephrase this

design-proposals/local-storage-migration/local-storage-migration.md
```diff
@@ -0,0 +1,261 @@
...
+
+The `MigrateNonSharedDisks` is a list of disks that represent the source volume to be migrated to the destination volume.
+
+The `ReclaimPolicySourcePvc` represents the policy that KubeVirt needs to apply to the source volume. Once the migration succeeds the source volume isn't used anymore and it can be treated in various ways. This is inspired by the [ReclaimPolicy](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaiming) of PersistentVolumes.
+
+In the VMI status under the `MigrateState` we report the migrate volumes:
+```golang
+type VirtualMachineInstanceMigrationState struct {
+
+        MigrationLocalDisks []MigrateNonSharedDisk `json:"migrationLocalDisks,omitempty"`
```
awels, line 89: Again would call it something like NonSharedStorage or something like that.

design-proposals/local-storage-migration/local-storage-migration.md
```diff
@@ -0,0 +1,261 @@
...
+
+// MigrateNonSharedDisk represents the source PVC to be migrated to the
+// destination PVC
+type MigrateNonSharedDisk struct {
+        SourcePvc      string `json:"sourcePvc,omitempty" valid:"required"`
+        DestinationPvc string `json:"destinationPvc,omitempty" valid:"required"`
+}
+
+// LocalStorageMigration represents the local storage to be migrated
+type LocalStorageMigration struct {
```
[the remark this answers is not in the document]
alicefr: Yes, naming is always challenging, well the proposal goal is also for this :)

alicefr:
I had some offline discussions on this topic, and to keep everyone on the same page, I'll summarize here the highlights:

1. The scope of this proposal is too broad and we should reduce a single use case (migrating local storage with the same PVC geometry and storage class)
2. Initially, we should avoid any APIs and KubeVirt will take care of the destination PVC creation. Later, once we gain more experience then we could introduce a new CRD where users can customize the PVC destination preferences
3. The migrate disks handling needs to be kept separated from the VirtualMachineInstanceMigration CRD, instead a new CRD can be added and then the migration controller could simply pick the information from there

After this feedback, I think the PoC PR is still valid, but I need to introduce the part that handles the destination PVC. Once, I finish, I'll update accordingly this proposal

@vladikr  @fabiand @davidvossel I hope I have captured the essence of our discussion. Please, correct if there is anything wrong

alicefr:
I'll close this proposal as we'll address separately every use-cases. I'll open a new one for the Storage Class migration

PR #242, opened 2023-08-31
Proposal for VM rollout strategy

vladikr:


jean-edouard (approved):

acardace (approved):

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,195 @@
...
+As a user, I want to be notified about changes that require a restart of the VM.
+
+## Repos
+
+kubevirt/kubevirt
+
+
+# Design
+
+A new field called rolloutStrategy will be introduced in the VM spec. This field will control the behavior of the VM object. Two possible values will be available:
```
fabiand, line 43: Can we add a section that is explaining why a new VM API field is the only way out?
Was it considered to start with cluster level (KubeVirt CR level) field only, and not extend the VM API?

If it was discussed, please decsribe why. If it was not discussed, then ... I'd encourage us to do so :)

fabiand (commented):
+1 on converging all the hotplug methods.

Thanks for this proposal. I left one comment as a starting point.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,195 @@
...
+2.  When `Stage` is used, `liveUpdateFeatures` will be ignored, except for special cases related to upgrades, see below.
+
+The above effectively means that the `LiveUpdateFeatures` field will be removed.
+
+The default value of the `rolloutStrategy` field if not specified will be `Stage`.
+
+If an existing VM (on upgrade) has the `LiveUpdateFeatures` field specified the `rolloutStrategy` field will be `LiveUpdate` and its `LiveUpdateFeatures` field will be removed.
+
+
+The `rolloutStrategy` field will be usable only if the `VMLiveUpdateFeatures` feature gate is enabled.
```
EdDev, line 69: If this was only a cluster-level option (i.e. in the KV CR), could we just use the FG instead of adding a new field?

EdDev (commented):
Thank you, I like the direction of this proposal.

I have for now one main question about how this will work:
If the `rolloutStrategy` is mutable, especially during the time the VM is running, how it will behave?
- Would it immediately try to live-update the VMI?
- What will happen to the condition suggested?
- What is the content of the condition?

The same could be asked the other direction: What will happen if the strategy changes to `Stage`? Although this seems a bit more trivial to me.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,195 @@
...
+        cpu:
+          sockets: 2
+          cores: 4
+          threads: 2
+          maxSockets: 8
+```
+  
+### Status API
+
+A condition called `RestartRequired` will be added to the VM requesting a restart if applied changes cannot be immediately applied.
```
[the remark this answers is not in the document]
vladikr: Sure, will add.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,195 @@
...
+
+
+# Design
+
+A new field called rolloutStrategy will be introduced in the VM spec. This field will control the behavior of the VM object. Two possible values will be available:
+
+1.  Setting RolloutStrategy to `LiveUpdate` means all changes to the VM object will be immediately applied to the VMI.
+    
+
+A condition called `RestartRequired` will be added to the VM if such changes are not possible to propagate,
```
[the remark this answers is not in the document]
vladikr: Well, there must be a way to communicate that a desired state is not reachable without an action.
Conditions fit the bill from my point of view.
What is your concern?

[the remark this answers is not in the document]
dankenigsberg: As a VM owner, I'd love to know that my VMI is not yet matching its desired state, and I'd like to know why. A condition like `RestartRequired` with details like `sockets > maxSockets; error hotplugging nic1;` would be very useful.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,195 @@
+# Overview
+
+Today changes to a VM object (i.e. changing VM.spec.template) get applied, but only take effect upon the next restart - we call these pending changes “staged”. This is a proposal to introduce a rollout strategy field to control how changes to the VM object will propagate to the VirtualMachineInstance object.
```
rthallisey, line 3: This was one of the proposed features for VMPools, but we didn't get to implementing it yet.   Can you add to your design how VMPools could leverage this feature on the VM API?

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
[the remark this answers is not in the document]
EdDev: > What is your concern?

I just wanted to know if this is needed for the users/clients or for an internal logic when processing the reconcile loop.
The only disadvantage that I can think of is that it is another thing to manage (e.g. if the VM is changed, the condition needs to be updated or removed accordingly).
If this is not needed for the internal logic, I would defer it to later and consider it an optional enhancement.

> As a VM owner, I'd love to know that my VMI is not yet matching its desired state, and I'd like to know why. A condition like RestartRequired with details like sockets > maxSockets; error hotplugging nic1; would be very useful.

Yes, that would be useful. However, a client could also calculate the diff in this regard and expose the gaps. The only thing it will not show is the reason it did not work (e.g. "not supported", "no permission").
But we also have logs and events that can be used to communicate things to users.

While I agree it may be nice to have it, we may (again) overdo things prematurely.
Said that, I do not have strong opinions on this, after all it does make sense.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
vladikr: For me, it would take a similar approach to the [DaemonSets](https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/#daemonset-update-strategy) update strategy: `OnDelete` and `RollingUpdate` or Deployments `strategy`. 

I think that if we only provide a cluster-level option, users will lose the flexibility to select different behavior for certain VMs.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: That's actually similar to the `DeploymentStrategy` of Deployments - which can be RollingUpdate or Recreate.

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#deploymentstrategy-v1-apps

One question that we have two streams of opinions about is:
Do we expose the different VM update strategies (liveupdate or (current) restart) on the VM or only on the Cluster level?

For VMPools I'd assume that we need a VM level control.
The VMPools either change the behavior if liveupdate is enabeld or not, or if liveupdate is enabled, then on VMPools the strategy can be selected.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: Ryan's VMPool example/request is along those lines as well.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
iholder101: A thought: it makes sense to me that some workloads must avoid migrations unless necessary. While I think that staging updates to the VM is probably not that useful in general, I think that using `stage` strategy is especially valuable to avoid being surprised by a migration.

Saying that, EdDev's idea regarding using selectors sounds reasonable and I think we should consider it.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
iholder101, line 48: If someone changes a configuration (e.g. `maxSockets`) with a VM using the `stage` strategy, then the condition is being added, will the condition be removed by reverting the configuration?

As a VM owner using a `staged` strategy, I think it's valuable to rely on this condition to check whether a configuration gets applied immediately or not.

In other words, VM owners could use the following approach if they wish to update a configuration:
1. Change a configuration
2. Check whether the condition is added

I think of a situation where someone wants to apply the config immediately if possible, but if he can't, then he doesn't want to stage it (one reason could be that he would be surprised by the changes in the next reboot and prefers to explicitly modify these configs only when the VM is shut down).

Not sure if I'm overthinking here.

iholder101:
Great stuff!
/cc

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
EdDev: Yea, the FG is not a fit to express how things should behave, only as means to enable/disable a feature and as part of its introduction lifecycle.

fabiand:
/hold

just to make sure that we don't forget about having this on the cluster level, and avoid the _requirement_ on a user to  configure this feature

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
vladikr: I suppose this evaluation would be part of the VM controller. It would add the condition when needed and remove it when the change has been applied or the strategy changed.
Similar to what we do with the live migration condition.

vladikr:
Thanks everyone for the valuable feedback. I've changed the proposal to start with a more simplistic approach.
The main change is that now the vmRolloutStrategy field will become a cluster wide setting that will apply to all VMs instead of a VM specific tunable.

dankenigsberg (changes requested):
minor English suggestions

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: Let#s defer the vmpool discussion pls

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,142 @@
...
+
+
+## Motivation
+
+In most cases, the user would prefer the changes to the VM object to immediately propagate to the VMI object, as long as the rollout does not interrupt the workload. This behavior would be similar to how the Deployments roll out DaemonSets.
```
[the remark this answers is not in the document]
vladikr: Let's go with `Eventually`

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,142 @@
...
+## Non Goals
+
+-   Successfully apply all the changes to the Virtual Machine Instance Object. Not all fields can be propagated to the guest immediately. Some fields should remain staged, such as those that affect the guest ABI (e.g. machine type)
+    
+-   Make LiveUpdate the default. We defer this decision.
+    
+
+## Definition of Users
+
+A user - a VM owner - with (at a minimum) create, update, and patch access for VM objects within a namespace.
```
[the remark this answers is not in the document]
vladikr: Done

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,142 @@
...
+
+
+## User Stories
+
+As a user I want changes that I applied to the VM object to propagate to the VMI immediately if possible.
+
+As a user, I want to be notified about changes that require a restart of the VM.
```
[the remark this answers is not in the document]
vladikr: Done

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,142 @@
...
+          sockets: 2
+          cores: 4
+          threads: 2
+          maxSockets: 8
+```
+  
+### Status API
+
+A condition called `RestartRequired` will be added to the VM requesting a restart if applied changes cannot be immediately applied.
+The condition will contain the reson for not being able to reach the desired state requested on the VM spec.
```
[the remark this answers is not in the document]
vladikr: Done

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,205 @@
...
+
+A new field called `vmRolloutStrategy` will be introduced in the KubeVirt CR. This field will control the behavior of the VM objects. Two possible values will be available:
+
+1. `LiveUpdate` means all changes to the VM object will be eventually applied to the VMI.
+    
+2. `Stage` would mean the current familiar behavior remains - all changes to the VM object will be staged and applied after a restart.
+
+In both cases, a condition called `RestartRequired` will be added to the VM to reflect the need for a restart and the reason why.
+
+The default value of the `rolloutStrategy` field if not specified will be `Stage`.
```
EdDev, line 56: `rolloutStrategy`or `vmRolloutStrategy`?

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,205 @@
...
+
+The `vmRolloutStrategy` field will be usable only if the `VMLiveUpdateFeatures` feature gate is enabled.
+
+
+## VM API
+
+The VM level `liveUpdateFeatures` field will be removed.
+VMs using the `liveUpdateFeatures` will be rejected by the API.
+
+If `RestartRequired` condition is present, any additional changes to the VM spec - to live-updatedable or non-liveupdateable fields - will be staged. Thus the VM has to be restarted.
```
EdDev, line 66: Can you explain this?
This condition can appear because a change that requires restart is set, but a different change after may not require a restart and perhaps is applied immediately.

The way it is defined here, sounds as if the following changes, even if they can be applied immediate, will not.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,205 @@
...
+A default value for the `MaxGuest` memory value will continue to be internally calculated based on the cluster setting of `maxHotplugRatio` 
+
+All other hot plug behavior wil remain the same.
+
+## Drawbacks / Limitations
+
+-   No cluster- and VM-level fine-grained control over rollout settings of specific features.
+-   No VM-specific controll over the rollout strategy.
+
+## API
```
EdDev, line 92: How do you know what is the **actual** maximum resources (CPU, memory) available? Is it reported through the status?

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,205 @@
...
+
+Today changes to a VM object (i.e. changing VM.spec.template) get applied, but only take effect upon the next restart - we call these pending changes “staged”. This is a proposal to introduce a custer-level rollout strategy field to control how changes to the VM object will propagate to the VirtualMachineInstance object without interupting the workload.
+
+
+## Motivation
+
+In most cases, the user would prefer the changes to the VM object to eventually propagate to the VMI object, as long as the rollout does not interrupt the workload. This behavior would be similar to how the Deployments roll out ReplicaSets.
+
+
+## Goals
```
EdDev, line 11: How is this going to work/sync with the network hotplug interfaces?
At the moment, it has a FG and is not controlled by any strategy (it is by default `LiveUpdate`).

I am in favor to graduate the network hotplug in v1.2 or v1.3. depending on user feedback.
But if we bind it to this new FG or to the rollout strategy parameter, we may have a backward compatibility issue on our hand.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,205 @@
...
+> **Note:** Below we are only discussing changes to fields that support live updates.
+                  References to `all` fields only refers to fields that can be live-updated. 
+
+## KubeVirt CR
+
+A new field called `vmRolloutStrategy` will be introduced in the KubeVirt CR. This field will control the behavior of the VM objects. Two possible values will be available:
+
+1. `LiveUpdate` means all changes to the VM object will be eventually applied to the VMI.
+    
+2. `Stage` would mean the current familiar behavior remains - all changes to the VM object will be staged and applied after a restart.
```
EdDev, line 52: I am unsure about the CPU and memory, but for network interfaces there are 3 options to apply changes:
- Restart
- Live/On-the-fly
- Migration

I have no seen the migration option mentioned in this proposal.
It does not play any role with CPU and memory?

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
vladikr: Yes. The VMI spec has the `spec.domain.cpu.maxSockets` and `spec.domain.memory.maxGuest`. These fields cannot be updated on a running VMI. These can be set only during creation.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
vladikr: Yes, I understand the potential complications.

I think the network hotplug interface will need to preserve its current enablement.
We definitely need to have a follow-up discussion about the harmonization of all hot pluggable interfaces (Network and Storage) - how can we make all interfaces be driven by the rollout strategy.

The fact that the the network hotplug is behind a feature gate means that we didn't commit to this API and effectively don't have to provide an upgrade path if we are unable to find one.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: It's just two cases:
- Restart (interrupt workload)
- LiveUpdate (Do not interrupt workload.

Then, live migration and on the fly are startegies for liveupdate.

IOW: Two states are enough IMO, and live update will lead to LM or on-the-fly according to what is possible - it's an implementation detail.

The same is btw true for mem/cpu once in-place pod updates / mutatable pods will GA.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: We need to align all hotplug mechanisms.

@mhenriks is currently working to prepare storage hotplug for a unified UX.

Before you @EdDev can GA network, we should sync and see how we can provide this unified UX.

IOW, you want to hold you GA until it's clear how a unified UX looks.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: +1 on vmRolloutStrategy - who knows what other rollouts we will have,

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
dankenigsberg: +1 for keeping only two strategies. If an when Network introduces an in-place hotplugging for certain devices, an additional feature gate can be added.

dankenigsberg (approved):
/lgtm

Thanks. I like the simplicity and usability of this proposal.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
vladikr: That's an interesting topic. 

At this point, the hot plug of CPU and memory is only done via live migration, since pods are immutable.
However, in the future, we are planning to eventually use the [In place pod update resources](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources) feature, instead or to accompany live migration.

However, how hot plug is performed can be decided on the backend, instead of having a tunable for this.

I wonder how can a user decide which of these options is better. Why not make this decision on the backend?

enp0s3:
@vladikr @fabiand Hey, just want to clarify something - the current proposal scope is for `vm.spec.template` or for the whole `vm.spec`?

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
vladikr: that was just a typo

enp0s3:
/lgtm

What I meant in my question is whether we are going to support VM rollout strategy for fields under `vm.spec.` besides the `template`, like changing the `instancetype` or `preference` etc.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,205 @@
...
+
+Changes to `spec.template.spec.domain.cpu.maxSockets` would require a restart of the VM. This would also set the `RestartRequired` condition.
+
+CPU hot plug is triggered by changing `spec.template.spec.domain.cpu.sockets`. This change will be propagated to the VMI object as long as its value is lower than the value of `MaxSockets` on the last VM startup.
+Increasing the value of the `spec.template.spec.domain.cpu.sockets` beyond the `MaxSockets` will require a restart to take effect.
+
+### Memory Hot Plug
+
+The logic of memory hot plug will be similar to CPU hot plug, with the exception of the fields names.
+`spec.template.spec.domain.memory.maxGuest` will become mutable and will remain optional.
```
stu-gott, line 82: We have the opportunity to rename this field as part of this. I suggest something like maxGuestRAM or maxGuestMemory.

fabiand:
@enp0s3 we try to capture this in 
https://github.com/kubevirt/community/pull/242/files#diff-0100c665f25ed29091dfe21c7d3009896807e68f3cbe79e9e249cedab2797d13R43-R45

with this note we are saying: The strategy applies to all fields which are live updateable. It does not matter if they are in `spec.` or `sepc.template` as long as they are live updateable.

fabiand:
/hold

/lgtm
/approve

I'm feeling comfortable to approve as I was involved in many discussions to see how this can converge.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
EdDev: I understand the direction, but the generalization of this proposal and the actual progress of things are somehow colliding.
- The proposal attempts to generalize the VM rollout strategy, using only two examples: CPU & Memory.
   It does not mention storage or networking, nor a plan on how to consolidate them into the same strategy.
- Networking hotplug is not formally GA, but it will be very difficult to break it at the moment based on its (perhaps premature) adoption. I think this needs to be further discussed to detect a possible solution in the future.

IMO, if there is a time to plan for the harmonization, it is now. This proposal de-facto defines the minimum needed for it to occur.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
EdDev: I also favor the two options from the user spec API.
I do think the condition added to the status needs to specify what is needed exactly: A reset of the domain/guest or a live-migration. From the UX perspective, the user may need to take action either way.

If you intend to automate migration as a result of detecting something, then it is possible but it needs to be explicit defined in the design. I am not sure if this is what we really want at the moment, because it has other implications, including complexity, that we do not know if it is a needed or not.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
acardace: I'm not against the change but being inside `memory` maxGuestMemory would be redundant.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
enp0s3: @EdDev Hi, I think that generalization of this proposal should be in a dedicated design proposal, something like "dynamic VM configurtation". In that proposal we can define graduation phases, and connect the subjects like hotplug harmonization to these phases.

fabiand: @EdDev there are two main points in this proposal

1. be declarative and permit eventual consistency, incl defining an error condition
2. have a minimal api

To me this is not mem or cpu specific, and I'm also not - at a first glance - seeing how this collides with network.

Ain't these two points for a GA'ed network hotplug desirable as well?

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: > Well, this will cause a complication either way.

I'm missing the "why"? :)

> I would prefer for everything to do its best to reach "sync".

I'm not seeing anything which would prevent to enhance/optimize this further in future.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: @vladikr I think the question is if the current (effective) number of cpu/memory is reported in the status?

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
EdDev: > Ain't these two points for a GA'ed network hotplug desirable as well?

The goals are, the solution was more aggressive with networking.
There we are opinionated and avoid the staging option.
At the time, we thought it makes sense. I still think it makes sense now, but it is hard to argue with a desire to have minimal control as suggested in this proposal with the strategy. What I am saying is that the defaults collide.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
EdDev: > > Well, this will cause a complication either way.
>
> I'm missing the "why"? :)

From the user perspective: It matters which field he changed first, that will determine if the whole future changes will take effect immediately or not
From the implementation: On any change detected, there is a need to check the status and examine the condition. Therefore, not only the fields that changes determines if the change should role out to the VMI, but also its condition status. There are corner cases with this approach, as the status may or may not be up to date, dependent on who sets the condition state and when it is read.
Perhaps this is much simpler then what I imagine.

> > I would prefer for everything to do its best to reach "sync".
>
> I'm not seeing anything which would prevent to enhance/optimize this further in future.

Correct, but I am also not seeing a problem to do the best possible to sync changes based on the available information. The fact that some prev change requires reset is not a reason a new change cannot be applied. In the end, we do have the desired state in the spec, if there are dependencies they can be checked against the desired and current.

Anyway, as it can be improved, I guess it is not critical at the moment.
It will be critical if you consider totally different areas being added to this rollout logic, e.g. storage, virt infra or network being dependent on one another in this regard.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
vladikr: @fabiand hmm, the original question was about `actual **maximum** resources`.
To your question, the current CPU topology and current memory are already being reported in the status.

EdDev: Too many names, but to be clear: actual == current == effective
The **maximum** can still be expressed based on the information available from the domain or pod, no?

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,207 @@
...
+
+1. `LiveUpdate` means all changes to the VM object will be eventually applied to the VMI.
+    
+2. `Stage` would mean the current familiar behavior remains - all changes to the VM object will be staged and applied after a restart.
+
+In both cases, a condition called `RestartRequired` will be added to the VM to reflect the need for a restart and the reason why.
+
+The default value of the `vmRolloutStrategy` field if not specified will be `Stage`.
+
+The `vmRolloutStrategy` field will be usable only if the `VMLiveUpdateFeatures` feature gate is enabled.
```
AlonaKaplan, line 58: Why do we need both feature gate and `vmRolloutStrategy` field?
The fg should be enough. If it is enabled we will `LiveUpdate` if disabled we will `Stage` (the old behavior).

acardace: @EdDev gave a pretty good answer to this already https://github.com/kubevirt/community/pull/242#discussion_r1418917622

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,141 @@
...
+As a user, I want to be notified about changes that require a restart of the VM.
+
+## Repos
+
+kubevirt/kubevirt
+
+
+# Design
+
+A new field called vmRolloutStrategy will be introduced in the KubeVirt CR. This field will control the behavior of the VM objects. Two possible values will be available:
```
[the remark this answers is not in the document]
AlonaKaplan: Is the "stage" strategy there because we think it is a valid and useful use case or we keep it just because the live update is not graduated and we are afraid there are some automation scripts depend on the old behavior?
If the answer is the second option, then I think it perfectly fits a feature gate.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,207 @@
...
+
+The `vmRolloutStrategy` field will be usable only if the `VMLiveUpdateFeatures` feature gate is enabled.
+
+
+## VM API
+
+The VM level `liveUpdateFeatures` field will be removed.
+VMs using the `liveUpdateFeatures` will be rejected by the API.
+
+If `RestartRequired` condition is present, any additional changes to the VM spec - to live-updatedable or non-liveupdateable fields - will be staged. Thus the VM has to be restarted.
```
EdDev, line 66: I would suggest this phrasing:

```suggestion
If `RestartRequired` condition is present, additional changes to the VM spec - to live-updatedable or non-liveupdateable fields - may be staged.
It is up to the implementation to determine if a live-update is possible while other changes have been staged.
```

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,207 @@
...
+
+A new field called `vmRolloutStrategy` will be introduced in the KubeVirt CR. This field will control the behavior of the VM objects. Two possible values will be available:
+
+1. `LiveUpdate` means all changes to the VM object will be eventually applied to the VMI.
+    
+2. `Stage` would mean the current familiar behavior remains - all changes to the VM object will be staged and applied after a restart.
+
+In both cases, a condition called `RestartRequired` will be added to the VM to reflect the need for a restart and the reason why.
+
+The default value of the `vmRolloutStrategy` field if not specified will be `Stage`.
```
EdDev, line 56: The network hotplug can align to this configuration field and its default.
I am not that happy about the default being "Stage", but it does make sense to sync all to the same.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,207 @@
...
+    
+-   Maintain backwards compatibility for cluster owners that prefer keeping staged changes until next VM reboot.
+    
+
+## Non Goals
+
+-   Live uppdate for **all** fields of the Virtual Machine API. Only select fields can be propagated to the guest. 
+-   Make LiveUpdate the default. We defer this decision.
+-   Support upgrades of VMs that are using the `liveUpdateFeatures`.
+-   This design does consider the hot-plugging of network interfaces and additional storage.
```
EdDev, line 23: The proposal is generic in all aspects until this non-goal.
Networking and storage hotplug exist, I think it will be beneficial to consider them as part of this proposal. The other alternative is to change this proposal naming and goals to be specific to two live changes (CPU & Memory).

I would prefer to keep it generic and pave the way for the harmonization of hotplug now.

mtessun: Ack, on keeping it generic - the API should be consistent regardless of the device that is hotplugged.

In addition this could be like in virsh, where you can define if it is for the definition, Live or both. So with that you could have a "non-persistent" (or ephemeral) hotplug as well.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,207 @@
...
+
+A VM owner - with (at a minimum) create, update, and patch access for VM objects within a namespace.
+
+
+## User Stories
+
+As a VM owner I want changes that I applied to the VM object to propagate to the VMI if possible.
+
+As a VM owner, I want to be notified about changes that require a restart of the VM.
+
```
mtessun, line 35: There is another User Stories (the most important one) - As a VM User I want my VM be able to cope with changes to my workload without the need of rebooting it. This applies to all kinds of devices - network, disk, CPU, Memory, etc.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,207 @@
...
+> **Note:** Below we are only discussing changes to fields that support live updates.
+                  References to `all` fields only refers to fields that can be live-updated. 
+
+## KubeVirt CR
+
+A new field called `vmRolloutStrategy` will be introduced in the KubeVirt CR. This field will control the behavior of the VM objects. Two possible values will be available:
+
+1. `LiveUpdate` means all changes to the VM object will be eventually applied to the VMI.
+    
+2. `Stage` would mean the current familiar behavior remains - all changes to the VM object will be staged and applied after a restart.
```
mtessun, line 52: I think having a state that is only applied Live but never to the VM object would add the flexibility for non persitent changes. Not sure this is needed, but sometimes you only want a device until reboot (e.g. a software distribution network).

So I would suggest 3 stages:
1. Live Update
2. Live Update+Stage
3. Stage only

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,207 @@
...
+In both cases, a condition called `RestartRequired` will be added to the VM to reflect the need for a restart and the reason why.
+
+The default value of the `vmRolloutStrategy` field if not specified will be `Stage`.
+
+The `vmRolloutStrategy` field will be usable only if the `VMLiveUpdateFeatures` feature gate is enabled.
+
+
+## VM API
+
+The VM level `liveUpdateFeatures` field will be removed.
```
mtessun, line 63: What happens if VMs are already using this (due to using some TP features) - would we silently remove that/ignore it? Or would we issue a warning that the VM definition needs to be changed?

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
mtessun: From a user perspective I agree with Ed.
Maybe I did a change that requires a restart (say add memory for now) and then I am adding a NIC, because I need to deploy some software over that network. Why would I not get this NIC hotplugged?

I think this will cause quite some confusion.
If the change cannot be applied, I think it would be better to issue a Warning/and Event telling that the hot-change failed.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,207 @@
...
+
+### CPU Hot Plug
+Since the LiveUpdateFeatures API will be deprecated the `spec.template.spec.domain.cpu.maxSockets`field of the VM object will become mutable and stays optional.
+If it is not set by the user then the maxSockets value will continue to be internally computed according the cluster-wide `maxHotPlugRatio`, as it is done today.
+
+Changes to `spec.template.spec.domain.cpu.maxSockets` would require a restart of the VM. This would also set the `RestartRequired` condition.
+
+CPU hot plug is triggered by changing `spec.template.spec.domain.cpu.sockets`. This change will be propagated to the VMI object as long as its value is lower than the value of `MaxSockets` on the last VM startup.
+Increasing the value of the `spec.template.spec.domain.cpu.sockets` beyond the `MaxSockets` will require a restart to take effect.
+
```
mtessun, line 78: Long discussion sumary here:
In theory Sockets are not needed for hotplug - the # of vCPUs would do. As this has quite some implications, going with sockets seems to be fine, but that also means that in case of a topology like 32 Cores/2Threads only 64! additional CPUs can be added (or removed) - which is often not wanted.

Ideally we will get this to as granular as a single vCPU  (as qemu fills up the topology if there is not a complete socket popuated on its own).

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,207 @@
...
+### Memory Hot Plug
+
+The logic of memory hot plug will be similar to CPU hot plug, with the exception of the fields names.
```
mtessun, line 81: I believe we don't have (and don't want) a memory topology (a.k.a. vDIMMs). As such I would rephrase the above - as it is not similar to CPU unless we change it to be #vCPUs and not #Sockets.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
mtessun: For the max Values - I would suggest to also check the physical node - as you can never go beyond the physical resources for CPUs and Memory for a single VM (at least you shouldn't). So having a maximum that is the current physical limit makes sense from my pov.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: Fair point, I think this can be tackled in a follow up as needed.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: The struct would be removed from the API definition, VMs with this struct will either be silently accepted (and the struct dropped) or rejected. @jean-edouard how will the system behave if we remove the api?

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: Yes, we would like to take this vCPU approach to a dedicated proposal.

vladikr: Qemu alone cannot play a role in determining the VM's topology as KubeVirt needs to manage it correctly to assign vCPUs to a cpuset it receives from k8s. 
This is required for VMs with dedicated CPUs of course. 
Also, the current VM API requires setting a topology and a numb of vCPUs.
KubeVirt can add a vCPUs count but it won't be done as part of this proposal.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
vladikr: Hotplug is implemented today via live migration. If the target pod won't be scheduled if it requests more resources than any of the nodes can offer.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
jean-edouard: This is a good question: what happens to existing k8s objects when an update to their API removes fields used by said objects?
I'll try and figure out a way to test this locally and let you know the outcome.
Either way though, this is just not a supported use-case.

jean-edouard: We appear to be lucky on this one. I deployed a cluster with KubeVirt 1.1.0 and a VM with some liveUpdateFeatures set.
Then I updated to a custom version with liveUpdateFeatures removed (but no code to deal with existing VMs).
I then looked at the VM yaml and found that the liveUpdateFeatures block was just gone.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
dankenigsberg: @EdDev do you have a constructive suggestion on how to make this proposal cover network, too? Would dropping this line be enough, or any other changes are suggested?

EdDev: > @EdDev do you have a constructive suggestion on how to make this proposal cover network, too? Would dropping this line be enough, or any other changes are suggested?

Yes, the suggestions have been added to the relevant threads.
https://github.com/kubevirt/community/pull/242/files#r1425171177

https://github.com/kubevirt/community/pull/242/files#r1425166735

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
EdDev: k8s is usually not removing fields, they are just left in place with a deprecation notice and a no-op.

See https://github.com/kubevirt/community/pull/251/files#diff-845f63068ea8ebf517e69fa5547ab723ec6279b6743bbc15b9a2de895c9bdbf0R241

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
vladikr: @EdDev If I read your replies correctly, you suggest that networking has a way forward and can be adjusted to this proposal.
The second comment is about a generic issue with the rollout dynamics, not specific to networking. 

Regarding storage, I wouldn't commit to it at this point. Storage hot plug is done today via a subresource. We need a separate design about how it can be implemented in a declarative manner.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
vladikr: > If you intend to automate migration as a result of detecting something, then it is possible but it needs to be explicit defined in the design. I am not sure if this is what we really want at the moment, because it has other implications, including complexity, that we do not know if it is a needed or not.

@EdDev Changes to the VM's CPU and Memory will trigger a live migration via the workload updater controller. This is how it is implemented today.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
EdDev: > @EdDev If I read your replies correctly, you suggest that networking has a way forward and can be adjusted to this proposal.

Yes, networking can confirm to the suggested strategy field when available.
We can update the network hotplug proposal/user-guide to reflect that before graduation.

> The second comment is about a generic issue with the rollout dynamics, not specific to networking.

It is indeed generic, but it effects networking. We want to make sure the proposal here is not limiting in the sense that network hotplug will depend on other changes done on the spec (e.g. CPU or Memory hotplug).
 
> Regarding storage, I wouldn't commit to it at this point. Storage hot plug is done today via a subresource. We need a separate design about how it can be implemented in a declarative manner.

I agree it needs a dedicated proposal for storage. We just need to leave this proposal in a state it can also be applied to storage in the future. This is why I think someone from the sig-storage should  also review this PR.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
EdDev: Ok, thanks for the clarification.
Networking hotplug is not automating this, and I think not all will really want to automate it.

As long as this is left to the implementation, I think we are good.
IMO it will help to express in the condition the need for migration for the implementations that do not automate it. But this can be added later if you do not see the benefit now.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: Ok. it would be great if network can align to this, and then we are good.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: There are a few concerns:
1. We don't see how one resource could reliably plug in new resource, if a different source has pending changes. this is due to the pod templating mechanism. Imagine new sockets are pending (thus they are in the spec already), if network now adds a new nic, then how do we prevent that the new socket (or rather the new cpu requests) end up in the new pod?
2. Because it is delcarative, are we sure that the declarative nature will be understandable to the user - what is applied and what not?
3. assume that we've got staged changes, and now a resource like net is pluggin in a new nic, but this fails, how do we clearly communicate these errors

concerns like this brought the authors to be more conservative about this point.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
jean-edouard: `RestartRequired` means the VM object is now out of sync with its VMI.
From that point on, the VM should not be allowed to influence the VMI until the disconnect is resolved (by rebooting).
Allowing live updates when `RestartRequired` is present requires very complex logic, which would be a lot more confusing to the user than just "Can't make new live changes until next restart, sorry".

For example, take a VM with 1 `sockets` and 4 `maxSockets` and **either**:
- Change `maxSockets` to 2, there's now a `restartRequired` on the VM. Now, do you allow updating `sockets` to 2? What about 3? Technically, `sockets` can be live updated to 3, but that would lead to an invalid VM spec...
- Change `maxSockets` to 6, now you can technically allow live-updating `sockets` up to 4, but you need to remember the old value of `maxSockets` (how?). If `sockets` is set to 5 that needs to be staged...

A partial solution to this would be to stage all CPU-related changes as soon as a CPU-related change triggers a `RestartRequired`, but how do we link the condition to a category of things? What about incompatibilities between sections? (for example, CPUs have a limit on how much memory they can handle, to go above you need more sockets)

For user's sake and implementation sake, I really think `RestartRequired` should be incompatible with live updates. As mentioned in previous comments, if in the future we find safe ways to be more fine-grained, we can always improve the logic!

vladikr: > For user's sake and implementation sake, I really think `RestartRequired` should be incompatible with live updates. As mentioned in previous comments, if in the future we find safe ways to be more fine-grained, we can always improve the logic!

+1 I agree.

I don't see how we can commit to being able to process conflicting changes on the VM spec in one iteration. The following proposal can undo this decision.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
xpivarc: Note: Storage has some problems concerning declarative hotplug because there is a dependency on the order of volumes.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: The thing is: We are running out of time.
These uncertainties are the reason why it was defined to not cover this point.

In order to unblock this whole ... disucssion .... I'd advocating for leaving this undefined.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
EdDev: It is an issue with the DB in the api-server, the ref I have there has all the discussion about it. k8s usually do not ever remove structs even if in Alpha, they just mark them as deprecated and they are in practice a no-op with a build and possibly runtime warning.

I think the topic itself is a bit unclear, I for sure do not understand fully their reasoning and therefore it is hard to be sure if ignoring it is possible or not.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: Thanks for the example.

I think that the networking side is coming from a side with simple examples:
1. modify sockets --> requires restart
2. add vnic and network, no conflict, no error, works

For my part I don't know how often this will work, but my point is that there are _some_ orthorgonal changes wich :migh_ often work to be applied without impacting the currently staged changes.

But maybe we can separate the discussion about the semantics of `RestartRequired` and the tradeoffs around it

If we seaparate it, then I'd propose to reword the sentence to: 

   If `RestartRequired` condition is present, the VM has to be restarted.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: We'll follow up on the vCPU discussion seaparately

dankenigsberg:
/lgtm

thanks, @vladikr .

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: The sentence is only focused on the api semantics, not on the ABI changes.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: Okay let me reprhase:
API support wise these fields will be removed.
The technical implementation of this has to be seen,

EdDev (approved):
Thanks!

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,206 @@
...
+          cores: 1
+          threads: 1
+status:
+  conditions:
+    - type: RestartRequired
+      value: true
+      reason: 'The number of sockets is too high and requires a VM restart in order to become effective".
+```
+
+## Update/Rollback Compatibility
```
xpivarc, line 204: note: for the future, it would be great to have a testing section. Approvers could ack/nack functional tests, unit tests, and such. This could help us correctly test things, note this is currently a problem recognized by many approvers.

xpivarc (commented):
SGTM, I do have an objection to mentioning the `liveupdates`, I think we have a proposal that should be updated and this can just be referred to as an alternative to it.

Also, it is not clear to me how and when the network and storage will be included. There is no mention that the expectation should be "In the future more fields will be live updated". I specifically mention this because we say we cover fields that are live updatable but at the same time we are leaving them out of the initial implementation. So mentioning the user expectations for now and having the implementation plan would help me manage my expectations.

jean-edouard:
> SGTM, I do have an objection to mentioning the `liveupdates`, I think we have a proposal that should be updated and this can just be referred to as an alternative to it.

@xpivarc I'm having a hard time understanding what you mean here, could you please be more specific? Thanks!

xpivarc:
> > SGTM, I do have an objection to mentioning the `liveupdates`, I think we have a proposal that should be updated and this can just be referred to as an alternative to it.
> 
> @xpivarc I'm having a hard time understanding what you mean here, could you please be more specific? Thanks!

We introduce `liveUpdateFeatures` in https://github.com/kubevirt/community/pull/216/files#diff-6e1b7afa2f9bab4f36ac1c613e5cd392828b29dc7d922b57a2ea1facebb52fd3R59-R61. This proposal is alternative to the added field. I think that the proposal should be updated with our learnings and that the removal of the field should be part of it.

fabiand:
> We introduce `liveUpdateFeatures` in https://github.com/kubevirt/community/pull/216/files#diff-6e1b7afa2f9bab4f36ac1c613e5cd392828b29dc7d922b57a2ea1facebb52fd3R59-R61. This proposal is alternative to the added field. I think that the proposal should be updated with our learnings and that the removal of the field should be part of it.

It's on an alternative to liveUpdates - it's the successor of it. It is an incremenetal evolution of the liveUpdate non-GAed feature in order to get to a sane and well defined user-facing API.

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Non Goals
+
+-   Live uppdate for **all** fields of the Virtual Machine API. Only select fields can be propagated to the guest. 
+-   Make LiveUpdate the default. We defer this decision.
+-   Support upgrades of VMs that are using the `liveUpdateFeatures`.
+
+## Definition of Users
+
+A VM owner - with (at a minimum) create, update, and patch access for VM objects within a namespace.
```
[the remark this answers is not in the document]
fabiand: IIRC then create, update, patch are part of the edit role, thus I'd simplify, but use the edit term

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,206 @@
...
+As a VM owner, I want to be notified about changes that require a restart of the VM.
+
+## Repos
+
+kubevirt/kubevirt
+
+
+# Design
+
+> **Note:** Below we are only discussing changes to fields that support live updates.
```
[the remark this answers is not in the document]
fabiand: reworded

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,206 @@
...
+
+If `RestartRequired` condition is present, the VM has to be restarted.
+
+## Changes to CPU/Memory hotplug workflow
+
+### CPU Hot Plug
+Since the LiveUpdateFeatures API will be deprecated the `spec.template.spec.domain.cpu.maxSockets`field of the VM object will become mutable and stays optional.
+If it is not set by the user then the maxSockets value will continue to be internally computed according the cluster-wide `maxHotPlugRatio`, as it is done today.
+
+Changes to `spec.template.spec.domain.cpu.maxSockets` would require a restart of the VM. This would also set the `RestartRequired` condition.
```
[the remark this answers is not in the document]
fabiand: fixed

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
fabiand: @xpivarc can you follow up with a PR to add this to the design template?

design-proposals/vm-rollout-strategy/vm-rollout-strategy.md
```diff
@@ -0,0 +1,206 @@
...
+
+CPU hot plug is triggered by changing `spec.template.spec.domain.cpu.sockets`. This change will be propagated to the VMI object as long as its value is lower than the value of `MaxSockets` on the last VM startup.
+Increasing the value of the `spec.template.spec.domain.cpu.sockets` beyond the `MaxSockets` will require a restart to take effect.
```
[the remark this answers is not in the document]
fabiand: Addressed

fabiand:
All comments have been addressed. No major points were identified to be missing.

/approve

PR #243, opened 2023-09-14
HTTP import with DIC - Design Proposal

ido106:
Signed-off-by: Ido Aharon <iaharon@redhat.com>

### DIFF
--- a/design-proposals/data-import-cron-http-import.md
+++ b/design-proposals/data-import-cron-http-import.md
@@ -6,24 +6,23 @@ The design will also demonstrate to the user how to operate with such import typ
 ## Motivation
 Currently, the DataImportCron allows the import of Registry Imports only. Recently, there was a demand to also allow HTTP import types.  
 The problem that arises from HTTP imports is that there is no convention between the different sources, so it's hard to know when the image is updated for each source in a generic way, which will make the polling process more difficult than standard registry sources.  
-One possible solution If the URL is static is to use a Get request with an If-Modified-Since header where we specify the date from which we want to check if there was a change. If there was a change since the specified date, the request will return with a status of 200OK, and then we know that the image has been updated since that date.  
-
-In the current situation we can enable HTTP imports by updating the DesiredDigest label.  
+The current approach is to support only sources that support the If-Modified-Since header.
+If there was a change since the specified date, the request will return with a status of 200OK, and then we know that the image has been updated since that date.
 
 ## Goals
 
-* Allow the user to perform http imports manually with dataimportcron
-* Create a poller that will cover as many cases as possible for automatic updating
+* Allow the user to perform http imports manually with dataimportcron.
+* Create a poller that will cover automatic updating with the help of If-Modified-Since header.
 
 ## Non Goals
 
-* The poller will probably not cover all import cases and sometimes the user will have to manually update the digest
+* The poller will probably not cover all import cases and sometimes the user will have to do manual update.
 
 ## User Stories
 
-* As a user, I would like to import images from an HTTP source using DataImportCron
-* As a user, I would like the poller automatically update the image when it is needed
-* As a user, I would like to manually trigger an HTTP import with DataImportCron
+* As a user, I would like to import images from an HTTP source using DataImportCron.
+* As a user, I would like the poller automatically update the image when the source is updated.
+* As a user, I would like to manually trigger an HTTP import with DataImportCron.
 
 ## Repos
 
@@ -41,7 +40,6 @@ In the current situation we can enable HTTP imports by updating the DesiredDiges
 
 # Design
 
-If the URL is static:
 * Make a GET request with the If-Modified-Since Header starting from the date stored in the AnnLastCronTime annotation
 * If the returned status is 200OK, perform import again
 * Update AnnLastCronTime to time.Now()
@@ -75,4 +73,18 @@ spec:
   garbageCollect: Outdated
   importsToKeep: 2
   managedDataSource: fedora
+```
+
+This OS image mirror example supports the If-Modified-Since header: https://mirrors.dc.clear.net.ar/ubuntu/ls-lR.gz
+```
+iaharon@home ~ $ curl -I https://mirrors.dc.clear.net.ar/ubuntu/ls-lR.gz
+HTTP/1.1 200 OK
+Server: nginx/1.14.0 (Ubuntu)
+Date: Thu, 02 May 2024 08:58:16 GMT
+Content-Type: application/octet-stream
+Content-Length: 26925864
+Last-Modified: Tue, 30 Apr 2024 13:22:27 GMT <<
+Connection: keep-alive
+ETag: "6630f093-19adb28"
+Accept-Ranges: bytes
 ```
--- a/design-proposals/data-import-cron-http-import.md
+++ b/design-proposals/data-import-cron-http-import.md
@@ -1,0 +1,78 @@
+
+# Overview
+This design provides an approach to allow DataImportCron to import HTTP sources.  
+The design will also demonstrate to the user how to operate with such import types.  
+
+## Motivation
+Currently, the DataImportCron allows the import of Registry Imports only. Recently, there was a demand to also allow HTTP import types.  
+The problem that arises from HTTP imports is that there is no convention between the different sources, so it's hard to know when the image is updated for each source in a generic way, which will make the polling process more difficult than standard registry sources.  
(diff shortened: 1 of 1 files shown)


design-proposals/data-import-cron-http-import.md
```diff
@@ -0,0 +1,78 @@
+
+# Overview
+This design provides an approach to allow DataImportCron to import HTTP sources.  
+The design will also demonstrate to the user how to operate with such import types.  
+
+## Motivation
+Currently, the DataImportCron allows the import of Registry Imports only. Recently, there was a demand to also allow HTTP import types.  
+The problem that arises from HTTP imports is that there is no convention between the different sources, so it's hard to know when the image is updated for each source in a generic way, which will make the polling process more difficult than standard registry sources.  
+One possible solution If the URL is static is to use a Get request with an If-Modified-Since header where we specify the date from which we want to check if there was a change. If there was a change since the specified date, the request will return with a status of 200OK, and then we know that the image has been updated since that date.
```
awels, line 9: So we can do the `If-Modified-Since` header on a schedule right? We know the time the job kicked off last so we pass that with the header. Do we know if all http servers support that header, I suspect not all of them do.

design-proposals/data-import-cron-http-import.md
```diff
@@ -0,0 +1,78 @@
...
+**Phase1** - Manual Import Update
+* DataImportCron controller to handle HTTP imports
+* DataImportCron webhook/validation to support HTTP source type
+
+**Phase2** - Poller
+* DataImportCron poller to support HTTP source polling
+
+# Design
+
+If the URL is static:
```
awels, line 44: I don't understand what you mean by `If the URL is static` by definition the URL passed to the Datavolume or DataImportCron is static, it cannot be modified. Is there some other property we are interested in? I don't get it.

design-proposals/data-import-cron-http-import.md
```diff
@@ -0,0 +1,78 @@
...
+* Make a GET request with the If-Modified-Since Header starting from the date stored in the AnnLastCronTime annotation
+* If the returned status is 200OK, perform import again
+* Update AnnLastCronTime to time.Now()
```
awels, line 47: Have you considered that some sources provide a separate file with the hash of the image? For instance the fedora [cloud images](https://download.fedoraproject.org/pub/fedora/linux/releases/38/Cloud/x86_64/images/Fedora-Cloud-Base-38-1.6.x86_64.raw.xz) have a [file](https://download.fedoraproject.org/pub/fedora/linux/releases/38/Cloud/x86_64/images/Fedora-Cloud-38-1.6-x86_64-CHECKSUM) that contains the checksum of the file. So we could download that file and verify the hash against what we have and decide if a new one exists that way.

design-proposals/data-import-cron-http-import.md
ido106: It may be that not all servers will support this header and then we will have to use another way.
Alex suggested that we can offload polling entirely to the user
@akalenyu

design-proposals/data-import-cron-http-import.md
ido106: Sometimes the image has the same URL and then we can use the way I mentioned, but sometimes the suffix changes when you upload a new version and we want to identify such a case, so we have to use another way.
When such case happens we want to update the `AnnSourceDesiredDigest` annotation to be the new suffix.

design-proposals/data-import-cron-http-import.md
ido106: Good point. The question is what percentage of sources support this and whether all sources provide the file in the same format?
For example https://cloud.debian.org/images/cloud/bookworm/daily/latest/SHA512SUMS looks different from the file you mentioned

design-proposals/data-import-cron-http-import.md
```diff
@@ -0,0 +1,78 @@
...
+kind: DataImportCron
+metadata:
+  name: fedora-image-import-cron
+  namespace: golden-images
+spec:
+  template:
+    spec:
+      source:
+        http:
+          url: "http://tinycorelinux.net/14.x/x86/release/Core-14.0.iso"
```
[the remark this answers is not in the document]
ido106: If the URL is static then it can help, if it changes so we need to find out the new suffix in addition to the hashcode to know the new URL. There could be other issues like the format of the checksum file.
This can help with Fedora case, for example `Fedora-Cloud-Base-38-1.6.x86_64.qcow2` supposed to stay the same.

Maybe we can use the name of the version that is in the same file you mentioned (e.g `CorePure64-14.0.iso`). The question is whether the name always changes when the hashcode changes or it is possible to push changes to an existing image even if it is not named differently.

design-proposals/data-import-cron-http-import.md
akalenyu: As far as I can tell these are the areas where this file structure diverges:
- Hash algorithm (MD5 in tinycore, SHA512 in debian, SHA256 in Fedora)
- File structure (`filename  hash` in most but fedora does `SHA256 (filename) = hash`)

design-proposals/data-import-cron-http-import.md
akalenyu: Yeah just something I thought about.. we could provide an example poller impl.
and require folks to supply a poller URL at the webhook level for HTTP sources

design-proposals/data-import-cron-http-import.md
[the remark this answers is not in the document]
akalenyu: If for some sources the filename `Fedora-Cloud-Base-38-1.6.x86_64` changes things get more
tricky for us, we might need some API on the DV level that is similar to sourceRef which derives
the new filename from the checksum file

design-proposals/data-import-cron-http-import.md
awels: Maybe limit the options of 'poller' type to a few known ones, including 'bypass' and 'if-modified-since'. We could expand that later if we identify more options. Maybe include the template idea from below, not sure.

design-proposals/data-import-cron-http-import.md
awels: Honestly, I don't think we should consider that particular scenario. From my perspective that is a completely different file. The URL that is passed in is always static and cannot change. There would be no mechanism that allows us to reliably know if a different file is newer than the original. I guess we could consider a 'template' url in combination with some version scheme, but that would be an enhancement IMO.

design-proposals/data-import-cron-http-import.md
awels: Maybe we can catalog these, if there is a limited number of permutations of these formats, maybe we can implement them, and only support those limited number of options. I understand there might be many, but if we capture the most common ones, that might be enough.

design-proposals/data-import-cron-http-import.md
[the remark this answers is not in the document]
awels: Maybe some templating type thing could be useful. But I think we should start out with just considering 'static' URLs, get that working properly, then expand to more complex use cases if what we have is not sufficiently powerful.

aburdenthehand:
@ido106 Do you still want to continue with this proposal? If so I'll remove the lifecycle labels. 
From what I can tell it still has some unresolved questions raised on it.

ido106:
@awels So as far as I understand, you suggest that at first we only handle instances that support the If-Modified-Since header, while assuming static links only

awels:
@ido106 welcome back, glad you are doing well. Yes, lets start simple and then enhance if we find that what we did is not sufficient.

ido106:
@awels Thanks !
Okay, I will edit the design proposal soon

ido106:
@arnongilboa what do you think?

design-proposals/data-import-cron-http-import.md
akalenyu: I still like offloading the polling logic to the user. But second thoughts about running it on their behalf since it means a user app will run in the CDI namespace. We could always just disable our poller and have them create their own cronjobs

ido106: Maybe we can do it with the checksum like @awels suggested, but there is no unique convention for the file structure

arnongilboa:
> @arnongilboa what do you think?

Supporting HTTP header If-Modified-Since smells like a good starting point.

jean-edouard:
@awels @akalenyu We should probably make a decision about this design proposal. Ignoring the details, do you agree with the general idea? Thanks

awels:
Yes the general idea is sound, the problem is the details. In particular how do you detect changed images when the http server doesn't provide the needed details. But the feature definitely something we want.

PR #249, opened 2023-11-09
Passt network interface API and implementation replacement

ormergi:
Following the emerging [Kubevirt Network Binding Plugin API](https://docs.google.com/document/d/13fLaqZuM8El_CT5XzxPv-1epKsB1-Spv_jtXveEkkdQ/edit#heading=h.z27rudfpwy44) and introduction of Passt network binding plugin https://github.com/kubevirt/kubevirt/pull/10425 https://github.com/kubevirt/kubevirt/pull/10621 
its now possible to remove Passt network binding implementation from Kubevirt core.

This PR introduces a plan for replacing Kubevirt's  core Passt binding implementation with the Passt network binding plugin.

### DIFF
--- a/design-proposals/passt-impl-iface-api-replacment.md
+++ b/design-proposals/passt-impl-iface-api-replacment.md
@@ -1,0 +1,183 @@
+## Overview
+Following the emerging [Kubevirt Network Binding Plugin API](https://docs.google.com/document/d/13fLaqZuM8El_CT5XzxPv-1epKsB1-Spv_jtXveEkkdQ/edit#heading=h.z27rudfpwy44)
+and introduction of Passt network binding plugin [2](https://github.com/kubevirt/kubevirt/pull/10425) [3](https://github.com/kubevirt/kubevirt/pull/10621), 
+it's now possible to remove Passt network binding implementation from Kubevirt core.
+
+This proposal outlines the replacement plan for the existing Passt implementation and API.
+
+# Motivation
+Removing Passt network binding implementation from Kubevirt core reduces the maintenance burden.
+Having Passt binding code outside Kubevirt's core may ease contribution process as there is no need to be familiar 
+with [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt) core code.
+
+## Goals
+- Reduce maintenance of Passt network binding in Kubevirt core.
+  
+## Non Goals
+
+## User Stories
+
+## Repos
+[kubevirt/kubevirt](https://github.com/kubevirt/kubevirt):
+- [Passt binding plugin hook sidecar](https://github.com/kubevirt/kubevirt/tree/main/cmd/sidecars/network-passt-binding)
+- [Passt binding plugin CNI](https://github.com/kubevirt/kubevirt/tree/main/cmd/cniplugins/passt-binding)
+
+# Design
+Replace Kubevirt core Passt network binding implementation and interface API with Passt binding plugin.
+Users should be aware of the coming changes and be prepared.
+
+The proposed plan for replacing Kubevirt core Passt binding implementation and API is as follows:
+
+1. Announce the coming changes for Passt binding described in this section at kubevirt-dev mailing list.
+   This will give users some time to prepare for changes in Kubevirt next release v1.2.0. 
+
+2. Kubevirt next release v1.2.0 should have the following changes:
+   1. Raise warning when Passt feature gate is enabled saying its deprecated.
+   2. Raise warning when Passt interface API is being used saying its deprecated and Passt binding plugin should be used instead.
+   3. Mark relevant filed as deprecated.
+
+   VM that use Passt interface API will work regularly.
+ 
+3. Kubevirt next release v1.3.0 should have the following changes:
+   1. [Passt interface API](http://kubevirt.io/api-reference/main/definitions.html#_v1_interfacepasst) 
+      usage will be blocked by the admission webhook.
+      Creating new VMs that use Passt interface API will be blocked.
+      Existing VMs that use Passt interface API won't manage to re-start.
+    
+      Users who like to use Passt binding will have to follow the network binding plugin API:
+      1. Install Passt binding CNI (make sure the binary available on all nodes).
+      2. Create and register Passt binding network-attachment-definition.
+      3. Register Passt binding sidecar image.
+         
+   2. Remove Passt binding implementation from Kubevirt core.
+   
+   3. Deprecate Passt feature gate.
+      Set Passt feature gate to always false as a discontinued feature, as it will be replaced by a network binding plugin.
+
+> **Note:**
+> Passt binary remain part of virt-launcher compute container, as it executed by libvirt it should be aligned with its version.
+> Decoupling Passt binary from the compute container may require additional changes to enable libvirt process
+> access passt binary.
+
+## API Examples
+### VM with Passt interface using the retiring Passt interface API:
+1. Enable Passt feature:
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+metadata:
+  name: kubevirt
+  namespace: kubevirt
+spec:
+  configuration:
+    developerConfiguration:
+      featureGates:
+      - Passt
+  ...
+```
+
+2. Specify Passt interface in the VM spec template:
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: myvm
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          interfaces:
+          - name: blue
+            passt: {}
+      networks:
+      - name: blue
+        pod: {}
+  ...
+```
+
+### VM with Passt interface using network binding API:
+
+1. Passt network binding plugin network-attachment-definition should be created, see the following example:
+```yaml
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: kubevirt-passt-binding
+spec:
+  config: '{
+            "cniVersion": "0.3.1",
+            "type": "kubevirt-passt-binding"
+  }'
+```
+
+2. Register Passt network binding network-attachment-definition and sidecar image, by specifying them in Kubevirt configuration, and enable network binding plugin feature:
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+metadata:
+  name: kubevirt
+  namespace: kubevirt
+spec:
+  configuration:
+    developerConfiguration:
+      featureGates:
+      - NetworkBindingPlugins
+    network:
+      Binding:
+        passt:
+          sidecarImage: quay.io/kubevirt/network-passt-binding:v1.0.0
+          networkAttachmentDefinition: kubevirt-passt-binding
+  ...
+```
+
+3. Specify the registered Passt network binding plugin in VM spec template:
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: myvm
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          interfaces:
+          - name: blue
+            binding:
+              name: passt
+      networks:
+      - name: blue
+        pod: {}
+  ...
+```
+
+
+## Scalability
+
+## Update/Rollback Compatibility
+In Kubevirt v1.3.0: 
+- Existing running VMs that use Passt interface API will keep running with no disruption.
+- Stopped VMs or new ones, will not manage to start and require using the network binding plugin API.
+> **Note:** Passt doesnt support migration.
+
+## Functional Testing Approach
+Kubevirt's Passt network e2e tests shall spin up VMs with Passt interface using network binding plugin API.
+
+## Implementation Phases
+1. Kubevirt v1.2.0: 
+   1. Raise warnings when Passt feature is enabled and Passt interface API is used.
+   2. Declare relevant filed deprecated.
+
+2. Kubevirt v1.3.0:
+   1. Set admission webhook to block VM creation when Passt interface is being used.
+   2. Remove Passt binding implementation from Kubevirt core
+   3. Deprecate Passt feature gate, i.e.: always enabled, no-op.
+
+# Open questions
+## 1. Can we set the webhook to raise warnings at Kubevirt v1.1.1?
+   Kubevirt core Passt binding implementation and Passt network binding plugin already live along each other.
+   We need to have Kubevirt raise warnings when Passt feature gate is enabled and when Passt interface API is being used.
+
+   By moving these changes to v1.1.1, changes targeted to v1.3.0 can be done at v1.2.0,
+   and eventually shorten the time to complete Passt implementation replacement.


design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,73 @@
...
+
+Kubevirt next release v1.2.0 should have the following changes:
+- Passt feature gate shall set as deprecated
+  When enabled, a warning shall be raised saying the feature gate is deprecated, will be removed in next release, 
+  and network binding plugin API shall be used instead.
+
+- VM whose spec specifies passt interface (`spec.domain.devices.interface.[].passt`) will be created with 
+  Passt network binding hook sidecar and CNI, in a way Kubevirt core implementation is not executed.
+
+  This will require users to set the following Passt binding plugin prequisites in advance:
```
AlonaKaplan, line 47: If the user has to add the sidecar and the NAD to the kubevirt CR anyway, I don't see the point of this intermediate state. I would skip phase1.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,73 @@
...
+## Phase 2
+Kubevirt follow-up release v1.3.0 should have the following changes:
+- Passt feature gate shall be removed.
+  Passt networking would still be feature-gated by the network binding plugin feature-gate. 
+
+- [Passt interface API](http://kubevirt.io/api-reference/main/definitions.html#_v1_interfacepasst) 
+  usage shall be blocked by the admission webhook with a message saying binding plugin API should be
+  used instead.
+
+- At this stage Passt binding can be removed from Kubevirt core.
```
AlonaKaplan, line 62: I could be removed in phase1 (the core passt code is not used there).

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: We can make this requirement less harsh by having additional phase that allow using both Passt interface API and network binding plugin API:
When Passt interface API is used:
- Kubevirt will set default image for passt hook sidecar image if non is registered (similar to slirp)
- Kubevirt core implementation will be executed only to configure the pod network namespace (what passt CNI does).
Alternatively, the user could still use the network binding API by registering Passt hook sidecar image and NAD and install Passt binding CNI.

This comes with maintenance cost in the period Passt interface API is supported.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,66 @@
...
+      Passt network binding hook sidecar and CNI, in a way Kubevirt core implementation is not executed.
+
+      This will require users to set the following Passt binding plugin prequisites in advance:
+      1.  Install Passt binding CNI plugin in cluster nodes
+      2. Create and register Passt binding hook sidecar image and network-attachment-definition, similar to 
+         [Kubevirt Network Binding Plugin - bridge binding example](https://docs.google.com/document/d/13fLaqZuM8El_CT5XzxPv-1epKsB1-Spv_jtXveEkkdQ/edit#heading=h.z27rudfpwy44).
+
+      If one or more is missing, newly created VMs with Passt interface won't be able to start, existing ones will fail migrate.
+
+   2. [Passt interface API](http://kubevirt.io/api-reference/main/definitions.html#_v1_interfacepasst)
```
AlonaKaplan, line 48: I don't understand how point 1 and 2 can live together? They have two possible solution to the same API.
I would keeo just point 2.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,66 @@
+## Overview
+Following the emerging [Kubevirt Network Binding Plugin API](https://docs.google.com/document/d/13fLaqZuM8El_CT5XzxPv-1epKsB1-Spv_jtXveEkkdQ/edit#heading=h.z27rudfpwy44)
+and introduction of Passt network binding plugin [2](https://github.com/kubevirt/kubevirt/pull/10425) [3](https://github.com/kubevirt/kubevirt/pull/10621), 
+it's now possible to remove Passt network binding implementation from Kubevirt core.
+
+This doc outlines deprecation plan for Passt network binding, interface API and  feature gate.
```
EdDev, line 6: "This proposal outlines the replacement plan for the existing passt implementation & API."

The "deprecation" is an implementation step, it is preferred not to set it as the subject of this proposal.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,66 @@
...
+
+# Motivation
+Removing Passt network binding implementation from Kubevirt core and use Passt network binding plugin instead,
```
EdDev, line 9: Can you please start with the reason (maintenance) and mention the "how" (replace) in the "Design" section?

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,66 @@
...
+# Motivation
+Removing Passt network binding implementation from Kubevirt core and use Passt network binding plugin instead, 
+enable reduce the maintenance burden.
+Having Passt binding code outside Kubevirt's core may ease contribution process as there is no need to be familiar 
+with [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt) core code.
+
+## Goals
+- Deprecate Passt network interface API.
+- Remove Passt feature gate.
+- Remove Passt network binding from Kubevirt core.
```
EdDev, line 17: These seems to be implementation details, not goals.
As I see it, the goal is to reduce maintenance from the core project for the passt (currently experimental) network binding.

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Sorry for the mess, fixed.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,61 @@
...
+
+## User Stories
+
+## Repos
+[kubevirt/kubevirt](https://github.com/kubevirt/kubevirt):
+- [Passt binding plugin hook sidecar](https://github.com/kubevirt/kubevirt/tree/main/cmd/sidecars/network-passt-binding)
+- [Passt binding plugin CNI](https://github.com/kubevirt/kubevirt/tree/main/cmd/cniplugins/passt-binding)
+
+# Design
+Deprecating Passt feature-gate and interface API and replacing it with Passt binding plugin
```
EdDev, line 29: fg are not deprecated, feaures are deprecated and that has implications of the fg itself.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,61 @@
...
+require some changes users should be aware of and prepared for.
+
+The proposed deprecation plan for Passt binding interface API and feature-gate is as follows:
+
+1. Announce passt feature-gate and interface API deprecation in kubevirt-dev mailing list.
+   This will give users some time to prepare for changes in Kubevirt next release v1.2.0.
```
EdDev, line 35: Same, the existing passt binding is deprecated, not the FG.
It needs to include the timeline and recommended steps to take to manullay upgrade to the new passt API.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,61 @@
...
+
+The proposed deprecation plan for Passt binding interface API and feature-gate is as follows:
+
+1. Announce passt feature-gate and interface API deprecation in kubevirt-dev mailing list.
+   This will give users some time to prepare for changes in Kubevirt next release v1.2.0. 
+
+2. Kubevirt next release v1.2.0 should have the following changes: 
+   1. [Passt interface API](http://kubevirt.io/api-reference/main/definitions.html#_v1_interfacepasst) 
+      usage shall be blocked by the admission webhook with a message saying binding plugin API should be
+      used instead.
```
EdDev, line 40: That is not possible or adviced due to potential upgrade breakage of existing users who work with this binding in its old form.
Kubevirt must assure the users can upgrade without breaking anything (e.g. when the VM is migrated or restarted, it will fail to come up due to this suggested behavior).

It is also unlikely to be possible to do such changes for the next release, but that is open for debate I guess.
For me, it makes sense to give opportunity for users to change their configuration for at least one version. Considering that they may hear about this only when v1.2 is released.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,61 @@
...
+      
+      To user Passt binding the user will have to follow the network binding plugin API:
+      1. Install Passt binding CNI.
```
EdDev, line 43: Deploy?
Consider specifying that it needs to be done on each node in the cluster.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,61 @@
...
+      2. Create and register Passt binding network-attachment-definition.
+      3. Register Passt binding hook sidecar image.
```
EdDev, line 45: We do not call it "hook" on the API.

design-proposals/passt-feature-gate-iface-api-deprecation.md
EdDev: What needs to be done first is to add deprecation warnings on fields in the code and as warnings/logs in the runtime.

Then, in the next version the implementation can be removed, leaving the API fields and warnings/logs. The API fields will then be a no-op.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,61 @@
...
+         
+   2. Remove Passt binding implementation from Kubevirt core.
+   
+   3. Remove Passt feature gate.
+      Passt networking would still be feature-gated by the network binding plugin feature-gate.
```
EdDev, line 50: There will be no FG for passt, this sentence is confusing.
Perhaps it will be better to remove this sentence.

EdDev (changes requested):
Thanks.

I am missing in the proposal the steps a user needs to take to replace his existing VM/VMI spec configuration from the old API of passt to the new one.

design-proposals/passt-feature-gate-iface-api-deprecation.md
EdDev: The suggestion as I saw it was:
> usage shall be blocked by the admission webhook with a message

So I am saying we cannot block as that will cause existing VM specs to be broken completely.

> Then in v1.2 we should remove the old implementation, right?

Remove the implementation without removing the API, that will remain and will warn/log it is a noop.

> I don't think it is a good idea to run a VM with passt binding that has nothing behind it and not returning an error? What binding this VM will receive? The user will think all is good and the VM is running but behind the scenes it will receive a binding the user didn't request?

IMO we have no choice, we cannot break existing users, even if they use an alpha or beta feature.
We need to allow their VM/s to operate as close as possible to what they wanted.
I thought about falling back to the "default" or to masquerade binding (which is the closest to passt).

The user should get events about his problematic configuration.
The alternative of suddenly blocking his VM from even running is worse.

But I am open for other opinions in this regard.

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Done

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Done

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Done

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Done

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Done

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Declaring a feature as alpha, including alpha fields in graduated APIs, and not being able to revert it, make these deceleration pointless and out life harder, not to mention the API is getting ugly.
We cannot foresee the future and predict what changes will turned out to be irrelevant, we should have a way to do such changes.

Passt its not graduated feature, and as one users who decided using it anyways should know it may break.
As long that we provide a way and enough time for users to prepare for it I thing its fine.

> IMO we have no choice, we cannot break existing users, even if they use an alpha or beta feature.
We need to allow their VM/s to operate as close as possible to what they wanted.
I thought about falling back to the "default" or to masquerade binding (which is the closest to passt).

It can work for passt binding, but its odd, the VM is set with different settings then the desired spec.

Can we call it a temporary state and remove the field from the API entirely at v1.3?

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: In general this what installing a CNI means.

I dont mind being explicit about it and mention cluster nodes, done.

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Done

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Done

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: In general this what installing a CNI means.

I dont mind being explicit about it and mention cluster nodes, done.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,62 @@
...
+- [Passt binding plugin hook sidecar](https://github.com/kubevirt/kubevirt/tree/main/cmd/sidecars/network-passt-binding)
+- [Passt binding plugin CNI](https://github.com/kubevirt/kubevirt/tree/main/cmd/cniplugins/passt-binding)
+
+# Design
+Replace Kubevirt core Passt network binding implementation and interface API with Passt binding plugin.
+Users should be aware of the coming changes and be prepared.
+
+The proposed plan for replacing Kubevirt core Passt binding implementation and API is as follows:
+
+1. Announce the coming changes for Passt binding in kubevirt-dev mailing list.
```
EdDev, line 31: The "changes" are not known yet at this stage, either add a short description of what they are (e.g. deprecation and removal of passt core binding feature) or a ref to where it is defined.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,62 @@
...
+   This will give users some time to prepare for changes in Kubevirt next release v1.2.0. 
+
+2. Kubevirt next release v1.1.1 should have the following changes:
```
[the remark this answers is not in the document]
EdDev: v1.1.1 is a maintenance, fixing bugs and not necesserly consumed by users. Therefore, the formal release we can do this is v1.2 IMO.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,62 @@
...
+   Raise warning when Passt feature gate is enabled saying its deprecated
```
EdDev, line 35: Missing `.` at the end.
Perhaps it is worth using a list here.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,62 @@
...
+   Raise warning when Passt interface API is being used saying its deprecated and Passt binding plugin should be used instead.
+ 
+3. Kubevirt next release v1.2.0 should have the following changes:
```
EdDev, line 38: It should be v1.3.0 IMO.
The users have time to adjust during v1.2. Until then may not know about the prev stage (or warnings).

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,62 @@
...
+Users should be aware of the coming changes and be prepared.
+
+The proposed plan for replacing Kubevirt core Passt binding implementation and API is as follows:
+
+1. Announce the coming changes for Passt binding in kubevirt-dev mailing list.
+   This will give users some time to prepare for changes in Kubevirt next release v1.2.0. 
+
+2. Kubevirt next release v1.1.1 should have the following changes:
+   Raise warning when Passt feature gate is enabled saying its deprecated
+   Raise warning when Passt interface API is being used saying its deprecated and Passt binding plugin should be used instead.
```
EdDev, line 36: The 3rd change: Mark the fields as deprecated, so we will also see it in code.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,62 @@
...
+ 
+3. Kubevirt next release v1.2.0 should have the following changes:
+   1. [Passt interface API](http://kubevirt.io/api-reference/main/definitions.html#_v1_interfacepasst) 
+      set to be no-op.
+      When specified, the default network binding will be set instead (bridge/masquerade).
```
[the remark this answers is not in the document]
EdDev: I admit this is a hard one.

On one hand, it will be nicer not to break existing workloads.
On the other hand, it is indeed problematic to fallback to something that is not the desired spec. Although, the system itself is fine with having a diff between the desired and the actual, as long as it is announced/exposed.

The first option, to block creation, is the easiest to implement.
It is also the easiest to fix if we get feedback that this is causing too much pain.
I guess we can start with it and see if there is other feedback about it.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,62 @@
...
+      The warning will also say the subject interface is set with default network binding. 
+    
+      Users who like to use Passt binding will have to follow the network binding plugin API:
+      1. Install Passt binding CNI (make sure the binary available on all nodes).
+      2. Create and register Passt binding network-attachment-definition.
+      3. Register Passt binding sidecar image.
+         
+   2. Remove Passt binding implementation from Kubevirt core.
```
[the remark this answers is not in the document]
EdDev: Step 3 includes the removal of the core implementation, making the FG disabled with no option to enable it and the blocking the creation of a VM/VMI that uses the core passt API.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,62 @@
...
+   
+   3. Remove Passt feature gate.
+
+## API Examples
```
EdDev, line 53: You should add here what existing today and how it should change to.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Scalability
+
+## Update/Rollback Compatibility
```
EdDev, line 57: It should explicitly say there is none.

A backward compatability section should be set and it should mention what happens to existing running VM/s and when things will break.
E.g. starting a stopped VM with (old) passt defined will not come up in v1.3. Migrating old VMs with old passt while running in v1.3 will fail. Existing running VMs will continue working. etc

EdDev (changes requested):

design-proposals/passt-feature-gate-iface-api-deprecation.md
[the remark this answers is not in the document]
AlonaKaplan: oh, ok. I got it wrong.

design-proposals/passt-feature-gate-iface-api-deprecation.md
AlonaKaplan: passt doesn't support migration anyway. I agree with the rest.

design-proposals/passt-feature-gate-iface-api-deprecation.md
EdDev: > Declaring a feature as alpha, including alpha fields in graduated APIs, and not being able to revert it, make these deceleration pointless and out life harder, not to mention the API is getting ugly.
> We cannot foresee the future and predict what changes will turned out to be irrelevant, we should have a way to do such changes.

This is how k8s have deprecated such fields and it has to do with their persistence in the DB.
In any case, such things are beyond the scope of this specific proposal and can be raised in the sig-api meetings.

> It can work for passt binding, but its odd, the VM is set with different settings then the desired spec.

it is indeed odd, but it does solve the problem of performing an agressive act that breaks existing VMs.
Anyway, as was commented in another thread, we can start with the breaking option and see the feedback.

> Can we call it a temporary state and remove the field from the API entirely at v1.3?

Fields cannot be removed, only the implementation can be removed.

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Done

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,161 @@
...
+  ...
+```
+
+
+## Scalability
+
+## Update/Rollback Compatibility
+In Kubevirt v1.3.0: 
+- Existing running VMs that use Passt interface API will keep running with no disruption.
+- Stopped VMs or new ones, will not manage to start and require using the network binding API.
```
AlonaKaplan, line 155: using the network binding plugin API.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,161 @@
...
+metadata:
+  name: kubevirt-passt-binding
+spec:
+  config: '{
+            "cniVersion": "0.3.1",
+            "type": "kubevirt-passt-binding"
+  }'
+```
+
+2. Register Passt network binding network-attachment-definition and sidecar image, by specifying them in Kubevirt configuration:
```
EdDev, line 108: And enabling the the `NetworkBindingPlugins` FG.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,161 @@
...
+# Design
+Replace Kubevirt core Passt network binding implementation and interface API with Passt binding plugin.
+Users should be aware of the coming changes and be prepared.
+
+The proposed plan for replacing Kubevirt core Passt binding implementation and API is as follows:
+
+1. Announce the coming changes for Passt binding described in this section at kubevirt-dev mailing list.
+   This will give users some time to prepare for changes in Kubevirt next release v1.2.0. 
+
+2. Kubevirt next release v1.2.0 should have the following changes:
```
AlonaKaplan, line 34: I think point 2 can be done in v1.1.1 and point 3 in v1.2.
We have passt binding plugin living alongside passt old API in 1.1, we just forgot the release note and the warning. We can "fix" it is 1.1.1.
Let's add an open question about it.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+
+## Scalability
+
+## Update/Rollback Compatibility
+In Kubevirt v1.3.0: 
+- Existing running VMs that use Passt interface API will keep running with no disruption.
+- Stopped VMs or new ones, will not manage to start and require using the network binding API.
+- Passt doesnt support migration.
```
EdDev, line 156: This is confusing, no need to add it.
If you think it is important, add it as a "note".

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Functional Testing Approach
+Kubevirt's Passt network e2e tests shall spin up VMs with Passt interface using network binding plugin API.
+
+## Implementation Phases
```
EdDev, line 161: Please ref to the section above for the plan, or just repeat it here in a very short form.

design-proposals/passt-feature-gate-iface-api-deprecation.md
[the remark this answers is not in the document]
ormergi: Added to open questions section

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,176 @@
...
+    
+      Users who like to use Passt binding will have to follow the network binding plugin API:
+      1. Install Passt binding CNI (make sure the binary available on all nodes).
+      2. Create and register Passt binding network-attachment-definition.
+      3. Register Passt binding sidecar image.
+         
+   2. Remove Passt binding implementation from Kubevirt core.
+   
+   3. Deprecate Passt feature gate.
+      Passt feature gate is set as no-op by being enabled by default, following Kubeivrt's [deprecation doc](https://github.com/kubevirt/kubevirt/blob/main/docs/deprecation.md).
```
EdDev, line 55: This will not work.
FG needs to be set hard-coded to disabled and not be editable anymore.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,176 @@
...
+
+2. Kubevirt v1.3.0:
+   1. Set admission webhook to block VM creation when Passt interface is being used.
+   2. Remove Passt binding implementation from Kubevirt core
+   3. Deprecate Passt feature gate, i.e.: always enabled, no-op.
+
+# Open questions
+## 1. Can we set the webhook to raise warnings at Kubevirt v1.1.1?
+   Kubevirt core Passt binding implementation and Passt network binding plugin already live along each other.
+   We need to do some finlization and set the webhook to raise warnings when Passt feature gate is enabled
```
EdDev, line 175: Why the warning needs to be done from the webhook?

Anyway, I would expect to see here a reason why it is better to do it on v1.1.1.

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Since passt implementation is going to be removed Passt FG will have no meaning, whether its enabled or not, isnt it?

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: > Why the warning needs to be done from the webhook?

This is implementation detail I will remove it, but in general it allows providing better feedback to the user as it will be raised when ` kubectl create/apply` is used.

> Anyway, I would expect to see here a reason why it is better to do it on v1.1.1.

Done

design-proposals/passt-feature-gate-iface-api-deprecation.md
EdDev: No, as this may have effect on old vm/s. But the definition is clear, when a feature is removed, it meams its fg is false and cannot be enabled anymore.

design-proposals/passt-feature-gate-iface-api-deprecation.md
EdDev: > This is implementation detail I will remove it, but in general it allows providing better feedback to the user as it will be raised when ` kubectl create/apply` is used.

I disagree.

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,178 @@
...
+
+2. Kubevirt v1.3.0:
+   1. Set admission webhook to block VM creation when Passt interface is being used.
+   2. Remove Passt binding implementation from Kubevirt core
+   3. Deprecate Passt feature gate, i.e.: always enabled, no-op.
+
+# Open questions
+## 1. Can we set the webhook to raise warnings at Kubevirt v1.1.1?
+   Kubevirt core Passt binding implementation and Passt network binding plugin already live along each other.
+   We need to do some finalization and set raise warnings when Passt feature gate is enabled and when Passt interface API is being used.
```
AlonaKaplan, line 175: 1. What finalization?
2. set raise?

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: How would it affect old VMs?
Running VMs will have no disruption, regardless of the FG state.

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Fixed

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: BTW when enabling deprecated FG you get warning from the webhook following `kubectl apply/patch ..`

design-proposals/passt-feature-gate-iface-api-deprecation.md
EdDev: I know of scenarios where the fg is passed to the virt-launcher. It may not effect passt feature, but it may effect other features.
In general, if a feature is discontinued, my interpretation is that its fg is false.
If the feature graduated to ga, it is true.

I can even inagine cases where the code is not changed at all, only the fg is being locked to one of the options.

ormergi: > I know of scenarios where the fg is passed to the virt-launcher. It may not effect passt feature, but it may effect other features.

Sorry I dont fully understand what you mean.
There are two FG that get passed to virt-launcher at syncVMI routine,non of them is Passt FG.
How deprecating Passt FG, in other words makes it always enabled,  is it going to affect other features?

> In general, if a feature is discontinued, my interpretation is that its fg is false.
If the feature graduated to ga, it is true.

The [deprecation doc](https://github.com/kubevirt/kubevirt/blob/main/docs/deprecation.md) lack in regard to discontinued features, there is on going discussion at https://github.com/kubevirt/kubevirt/issues/7745 though.

Passt FG refereed to only at admission webhook when it condition Passt FG to be enabled when using Passt interface API.
As soon as Passt interface API usage is always blocked, Passt FG has no meaning whether its enabled or not.

Is you insist of the semantics, where Passt is discontinued feature and shall be always set to false, we may need to extend https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/feature-gates.go#L96-L119
to take in account such features.
Althogtgh I dont think its worth it because Passt FG will have no affect on the code.

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: @EdDev Dissagree with what? with raise warning at webhook level?

design-proposals/passt-feature-gate-iface-api-deprecation.md
```diff
@@ -0,0 +1,178 @@
...
+
+2. Kubevirt v1.3.0:
+   1. Set admission webhook to block VM creation when Passt interface is being used.
+   2. Remove Passt binding implementation from Kubevirt core
+   3. Deprecate Passt feature gate, i.e.: always enabled, no-op.
+
+# Open questions
+## 1. Can we set the webhook to raise warnings at Kubevirt v1.1.1?
+   Kubevirt core Passt binding implementation and Passt network binding plugin already live along each other.
+   We need to do some have Kubevirt raise warnings when Passt feature gate is enabled and when Passt interface API is being used.
```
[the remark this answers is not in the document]
ormergi: Sorry for the mess, Fixed.

design-proposals/passt-feature-gate-iface-api-deprecation.md
ormergi: Done

EdDev (approved):
Thanks!

vasiliy-ul (commented):
Hey, there are quite some conversations in this PR. Sorry if it has been discussed already. I am wondering: currently, virt-launcher container image includes the passt binary. Do you have any plan to decouple it and perhaps run in a separate container? Or will it remain part of the 'core' KubeVirt?

ormergi:
> Hey, there are quite some conversations in this PR. Sorry if it has been discussed already. I am wondering: currently, virt-launcher container image includes the passt binary. Do you have any plan to decouple it and perhaps run in a separate container? Or will it remain part of the 'core' KubeVirt?

Hi @vasiliy-ul, passt binary will remain part of Kubevirt core (virt-launcher compute container).
Passt is executed by libvirt and it better be aligned with libvirt version.
Thus we prefer to no invest in decouple them at the moment.

EdDev (approved):

PR #252, opened 2023-12-12
Adding a proposal for existing hooks sidecar

victortoso:
Adding a proposal based on @EdDev recommendation [here](https://github.com/kubevirt/kubevirt/pull/10842#issuecomment-1850673681.

The goal is to define a clear path for API graduation of the hooks. I'm not sure how much technical detail I should add in the proposal, so feel free to let me know what I have missing.


design-proposals/sidecar-hooks.md
```diff
@@ -0,0 +1,54 @@
...
+
+The gRPC server runs in the Sidecar. The Sidecars will use the module to establish the communication
+with virt-launcher and exchange capabilities to define which Hook version is being used and what are
+the hooks that are implemented.
+
+Virt-launcher will be entitled with managing the hooks. It'll verify and connect to all the Sidecars
+that are running by checking the pre-defined folder where the unix sockets for gRPC communication
+will be created.
+
+Virt-launcher should apply the requested changeds to the VMI.
```
lhmzhou, line 44: nit: requested changes*

design-proposals/sidecar-hooks.md
```diff
@@ -0,0 +1,54 @@
...
+A way to accomodate use cases that we might not want to bring and support in KubeVirt itself or to
+give room to test and develop features while they are still not mature enough.
+
+## Goals
+
+- Provide and maintain a set of gRPC APIs for users to do modifications to the VMI
+
+## Non Goals
+
+- Maintain a wide variety of user's applications in KubeVirt codebase
```
lhmzhou, line 20: Maybe expand on the non-goals: What types of user applications are considered "wide variety," and why is it not a goal to maintain them in the KubeVirt codebase?

design-proposals/sidecar-hooks.md
```diff
@@ -0,0 +1,54 @@
...
+
+Virt-launcher will be entitled with managing the hooks. It'll verify and connect to all the Sidecars
+that are running by checking the pre-defined folder where the unix sockets for gRPC communication
+will be created.
+
+Virt-launcher should apply the requested changeds to the VMI.
+
+## Hooks API promotion
+
+Should follow the same rules as stated in [API Graduation Guidelines][]
```
lhmzhou, line 48: nit: missing link to `API Graduation Guidelines` on Line 48

design-proposals/sidecar-hooks.md
victortoso: Thanks, fixed

design-proposals/sidecar-hooks.md
victortoso: I've used relative path line 54 (last line). It seems to have worked when looking from my personal branch
https://github.com/victortoso/kubevirt-community/blob/hooks-formal-proposal/design-proposals/sidecar-hooks.md

design-proposals/sidecar-hooks.md
victortoso: That's a good point. I think the important factor is core functionalities for KubeVirt that can be helped with sidecars. We do provide and maintain Today sidecars that configure slirp and passt, named "network pluggins".

I think it is reasonable that we improve hooks API as needed for core functionalities like that, but not necessarily to all cases.

If we do have a lot of sidecars with custom features that users want to have in a single place, we should simply create another repository for it. It does not need to be in KubeVirt repo, nor maintained by KubeVirt developers.

design-proposals/sidecar-hooks.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Non Goals
+
+- Provide and maintain a wide variety of sidecars that are not part of KubeVirt's core
+  functionalities.
+
+## User Stories
+
+- As a VM owner, I want to apply custom configurations to QEMU command line
+  - For debugability, by adding support or tweak [to run with debug tools][]
```
alicefr, line 27:
```suggestion
  - For debuggability, by adding support or tweak [to run with debug tools][]
```

design-proposals/sidecar-hooks.md
```diff
@@ -0,0 +1,56 @@
...
+- Provide and maintain sidecars that are important for core functionalities
+
+## Non Goals
+
+- Provide and maintain a wide variety of sidecars that are not part of KubeVirt's core
+  functionalities.
+
+## User Stories
+
+- As a VM owner, I want to apply custom configurations to QEMU command line
```
alicefr, line 26: It isn't really only for QEMU commandline. This is a single option. More in general is for providing custom XML portions

design-proposals/sidecar-hooks.md
```diff
@@ -0,0 +1,56 @@
...
+VMI will be deployed.
+
+## Motivation
+
+A way to accommodate use cases that we might not want to bring and support in KubeVirt itself or to
+give room to test and develop features while they are still not mature enough.
+
+## Goals
+
+- Provide and maintain a set of gRPC APIs for users to do modifications to the VMI
```
EdDev, line 16: I think these are too specific, possibly even touching the implementation.

I guess that the main goal is to define a stable integration point for custom programs, that are looking to extend Kubevirt VM functionality.

design-proposals/sidecar-hooks.md
```diff
@@ -0,0 +1,56 @@
...
+- Provide and maintain sidecars that are important for core functionalities
```
EdDev, line 17: I am unsure if this merits a special goal. Both external and internal sidecars are expected to get the same assurance and stability for the integration point.

design-proposals/sidecar-hooks.md
```diff
@@ -0,0 +1,56 @@
...
+
+- Provide and maintain a wide variety of sidecars that are not part of KubeVirt's core
+  functionalities.
+
+## User Stories
+
+- As a VM owner, I want to apply custom configurations to QEMU command line
+  - For debugability, by adding support or tweak [to run with debug tools][]
+  - To test unsupported features or apply changes that [could workaround bugs][]
+- As a KubeVirt developer, I want to provide alternative solutions that might not be ready to core
```
EdDev, line 29: We want to provide a modular, plug-able solution, to Kubevirt without overloading the core components. 

It is not necessary a matter of readiness.

design-proposals/sidecar-hooks.md
```diff
@@ -0,0 +1,56 @@
...
+
+The gRPC server runs in the Sidecar. The Sidecars will use the module to establish the communication
+with virt-launcher and exchange capabilities to define which Hook version is being used and what are
+the hooks that are implemented.
+
+Virt-launcher will be entitled with managing the hooks. It'll verify and connect to all the Sidecars
+that are running by checking the pre-defined folder where the unix sockets for gRPC communication
+will be created.
+
+Virt-launcher should apply the requested changes to the VMI.
```
EdDev, line 46: This is unclear.
I think virt-launcher is suppose to proxy domain creation and changes through the hook point before applying it to the domain manager (libvirt).

design-proposals/sidecar-hooks.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Hooks API promotion
```
EdDev, line 48: I think we are missing the API definition with all its details.
There is a need to express all existing versions and describe what each functionality does.

There is also a need to explain what are the implications and considerations of a sidecar author regarding backward and forward compatibility. Perhaps starting with the virt-launcher side, which relates to the Kubevirt support of hooks and what are the expectations from the server side.

It will be useful to elaborate on what happens when Kubevirt upgrades occur and what happens when the sidecar is upgraded.
Migration also needs to be taken into account, to explain the implications.

EdDev (commented):
Thank you for starting this effort, we are lacking intent and design thoughts on the sidecar hooks.

Recent work on sidecars & hooks exposed challenges on how to evolve the protocol API, keep old sidecars working and a clear view forward about the API stability so it can be trusted.

alicefr:
@victortoso I think we are missing what happen when we upgrade and migrate virt-launcher pods with sidecars. I'm thinking to the case when we upgrade virt-launcher with a deprecated hooks API. Do we have a mechanism to change the sidecar image on migration? Like, I want to upgrade my sidecar at the same time I upgrade KubeVirt

victortoso:
>  Do we have a mechanism to change the sidecar image on migration?

@alicefr there is none. The image is defined on annotation and that's it. That's a great use case.

The problem is that, this and other suggestions are improvements that should likely be implemented for the beta version. Should we define this now in this PR? Perhaps I can add a section of Future improvements? Or we keep it simple with only alpha related APIs / limitations and we update this when we agree on all improvements we want for Beta.

design-proposals/sidecar-hooks.md
victortoso: Thanks, I was in doubt about how deep I should go with the details. I'll go a little bit deeper.

design-proposals/sidecar-hooks.md
victortoso: Fixed, thanks.

design-proposals/sidecar-hooks.md
victortoso: Right, what you said is true but should be covered by the first bullet. What the second bullet introduces is that, we will provide some sidecars which add functionalities to KubeVirt and those sidecars we will be maintaining.

We might want to take out the second bullet out or not. Not sure where to draw the line in this document.

design-proposals/sidecar-hooks.md
victortoso: Hum, yes. In the end, it is an interface to make changes. I've changed it to the following:

```diff
-- Provide and maintain a set of gRPC APIs for users to do modifications to the VMI
+- Provide and maintain an interface for custom applications to do modifications to the VMI
```

design-proposals/sidecar-hooks.md
victortoso: Ah, do you disagree with it? I thought that `network binding plugins` was an example of this. Of course, I can be wrong.

design-proposals/sidecar-hooks.md
victortoso: Perhaps I can do a better wording here. What happens is that virt-launcher has an XML ready to be applied, but provides this to the sidecar over hook API. The sidecar apply changes to it. Virt-launcher applies the result XML to libvirt.

Changed to:
```diff
-Virt-launcher should apply the requested changes to the VMI.
+Virt-launcher will apply the result data provided by the sidecars and then create the VMI.
```

victortoso:
/remove-lifecycle stale

I'll get back to this. Its just that there are related ongoing discussions/events around hook sidecars that would merit being added here too.

go-bai:
@victortoso I think that when multiple hook sidecars have the same hook point, we can sort those hook points based on priority to determine the execution order. https://github.com/kubevirt/kubevirt/pull/11941

EdDev:
GA features already depend on the API between the virt-handler and virt-launcher, it is important to provide confidence that the protocol is stable and supported. Its extension was not smooth, and that needs attention.
We should revive this one, it is a technical dept that should be sorted out.

/remove-lifecycle rotten

EdDev:
> > GA features already depend on the API between the virt-handler and virt-launcher, it is important to provide confidence that the protocol is stable and supported.
> 
> The fact that GA features are relying on this does not imply stability already? I mean, I'm sure there are improvements that can be made...

It implies commitment and a strong dependency.
I would prefer to see a stable protocol for Beta. Once in Beta, it is not common or advised to ever break backward comparability. Current methods to extend the protocol/API break it unless we keep all previous structures.

> > Its extension was not smooth, and that needs attention.
> 
> Can you clarify? Just to be sure I'm not missing anything.

When something new is added, we are forced to keep all previous structures in order to keep compatibility.
We need something similar to grpc, where the client and server can negotiate the common minimum.

PR #245, opened 2023-09-25
design-proposal: volume migration

alicefr:
This proposal is a follow-up from the https://github.com/kubevirt/community/pull/240 and only targets moving the storage to another PVC. The VM live migration with local storage is out-of-scope.

In comparison to the previous proposal, this introduces a new CRD for representing the Volume Migration.

PR #260, opened 2024-02-22
design proposal: volume update strategy

alicefr:
Third attempt and version for the volume migration

After several discussions, the API and solution diverged quite a lot from the original idea and CRD proposed in https://github.com/kubevirt/community/pull/245. Therefore, I decided to close the previous proposal and open a new PR.

This version introduces a new update strategy for volume under the VM spec and tries to follow the same flow as the cpu and memory hotplug described in the [cpu-hotplug proposal](https://github.com/kubevirt/community/blob/main/design-proposals/cpu-hotplug.md)

I think the most controversial part of this proposal is the data volumes restriction. In the previous proposal, it was the controller that was getting rid of the datavolume parts. Now, we want that only the user touches the VM spec, then the DV declarations need to be removed manually. Please, consider that we plan to build an overlaying tool that should simplify the use of this API.

### DIFF
--- a/design-proposals/volume-update-strategy.md
+++ b/design-proposals/volume-update-strategy.md
@@ -1,0 +1,277 @@
+# Overview
+This design proposal extends KubeVirt APIs to enable the storage migration between PVCs.
+
+## Motivation
+In some circumstances, the guest owner may want to have the option of switching the storage provider or class for a group of volumes used by an active VM. Currently, the VM must be turned off and recreated, causing service disruptions and downtime.
+
+
+## Goals
+Design and develop an API to live update and migrate PVCs.
+  * The volume migration needs to be declarative and [GitOps](https://kubevirt.io/user-guide/operations/gitops/) compatible
+  * Volume migration implies an update of the volumes used by the VM. Hence, we want to trigger the storage migration on a VM update.
+  * The VM spec updates needs to be triggered by the user and not to be implicitly updated by the KubeVirt controller
+
+ ## Non  Goals
+*  This feature is not directly aimed at live migrating the VM using RWO volumes. Users will therefore continue to run into the restriction that the VM cannot be live-migrated if it uses RWO volumes.
+*  This design doesn't cover migration between storage classes, but rather implements basic APIs that can be used by overlying tooling.
+*  This design doesn't include the creation of the destination PVCs, it expects the destination PVCs to exist.
+
+ ## Definition of Users
+This feature is for all KubeVirt users who want to move the PVCs assigned to a VM without workload interruptions.
+
+## User Stories
+
+  1. As a VM owner, the storage class of volumes used by a running VM is deprecated, and the volumes need to be moved to another storage class.
+  2. As a VM owner, I want to use higher performance storage class that is now available in the cluster.
+  3. As a VM owner, I see my non-resizable storage device filling up and I want to move my VM to a larger storage device without interruptions in order to give it more storage.
+  4. As an infra admin, I need to replace the storage array due to a hardware refresh and I want to do this without stopping the VMs in order to make this infra operation transparent to VM owners.
+
+## Repos
+
+  - [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+
+This proposal introduces a new field in the VM spec to represent the volume update strategy called `updateVolumesStrategy`. This value is a unique field per VM, it can takes the values:
+ * `replacement` when the update of the volume aims to substitute the volume with another upon a VM restart
+ * `migration` when the update of the volume wants to migrate a volume to another one performing a full-copy of the content
+
+If no value is specified, the default behavior is `replacement`.
+
+Possibly, in the future, we could add a strategy for hotplugged volumes when the update wants to immediately add a new volume to a running VM.
+
+## API examples
+
+#### VM Spec API update
+
+Original VM spec
+```yaml
+  spec
+  volumes:
+  - name: vol1
+    persistentVolumeClaim:
+     claimName: src-pvc1
+  - name: vol2
+    persistentVolumeClaim:
+     claimName: src-pvc2
+  - name: vol3
+     persistentVolumeClaim:
+    claimName: src-pvc3
+```
+
+Update VM spec:
+```yaml
+spec:
+  updateVolumesStrategy: migration
+  volumes:
+  - name: vol1
+    persistentVolumeClaim:
+     claimName: dst-pvc1
+  - name: vol2
+    persistentVolumeClaim:
+     claimName: dst-pvc2
+  - name: vol3
+    persistentVolumeClaim:
+      claimName: src-pvc3
+```
+
+In this, example the user is able to update and migrate the first and second volume while the third one remains untouched.
+
+## Flow
+
+The volume migration update strategy follow the same [update sequence](https://github.com/kubevirt/community/blob/main/design-proposals/cpu-hotplug.md#phase-2-triggering-a-live-migration-when-the-number-of-sockets-changes-to-adjust-the-virt-launcher-cpu-resources) proposed for memory and cpu hotplug.
+
+Sequence of events if the `updateVolumesStrategy` is `migration`:
+
+* the user updates the volumes on a vm.spec.template.spec.domain.volumes
+* the VM controller determines the volumes which needs to be updated
+* the VM controller validates if the volumes can be migrated
+* the VM controller updates the volumes accordingly in the active vmi.Spec
+* the VMI controller writes the required volumes to migrate in the vmi.Status and set the update condition in the VMI status
+
+Example of the VMI status:
+```yaml
+  status:
+    conditions:
+    - lastProbeTime: null
+      lastTransitionTime: "2024-04-09T07:51:30Z"
+      message: migrate volumes
+      status: "True"
+      type: VolumesChange
+[..]
+    migratedVolumes:
+    - destinationPVCInfo:
+        claimName: dest-pvc-1
+        volumeMode: Block
+      sourcePVCInfo:
+        claimName: src-pvc-1
+        volumeMode: Filesystem
+      volumeName: volume1
+    - destinationPVCInfo:
+        claimName: dest-pvc-2
+        volumeMode: Block
+      sourcePVCInfo:
+        claimName: src-pvc-2
+        volumeMode: Filesystem
+      volumeName: volume2
+```
+
+* the VMI workload-update controller detects the update condition `VolumesChange` and triggers the VMI migration. When it creates the VMI migration object it also set the label `kubevirt.io/volume-update-migration: vm-name`.
+* the Migration controller constructs a new target pod with the desired volumes
+* virt-launcher updates the volumes to migrate in the migration request for LibVirt
+* the VMI live migrates
+* the VMI status is updated to reflect that the volume update action have completed and the update condition is removed
+
+This feature depends on the `vmRolloutStrategy` which needs to be set to `LiveUpdate`. On the long term, this should become the default behavior.
+
+The label `kubevirt.io/volume-update-in-progress` can be used to filter and identify the VM migrations triggered by a volume update.
+
+The status of the volume update can be monitored by using the label and getting the corresponding VMI migration:
+```console
+$ kubectl get virtualmachineinstancemigrations -l kubevirt.io/volume-update-in-progress: vmi` --watch=true
+NAME                           PHASE       VMI
+kubevirt-workload-update-abcd  Running     vmi
+```
+
+The volume migration can be aborted by changing the destination migrated volumes back to the previous ones if the migration is still in progress.
+If there is a volume update in progress and the volumes don't correspond to the old set, then the update is ignored and the migration is continued. This avoids to create a confusing state for the volumes of the VM during the migration.
+
+Example of the VMI status:
+```yaml
+  status:
+    conditions:
+    - lastProbeTime: null
+      lastTransitionTime: "2024-04-09T08:00:53Z"
+      status: "False"
+      type: VolumesChange
+    - lastProbeTime: null
+      lastTransitionTime: "2024-04-09T08:00:53Z"
+      message: volume migration aborted because migrated volumes have changed
+      status: "True"
+      type: ChangeAbortion
+```
+
+If the volume migration has already completed then this is simply considered a new volume update.
+
+## Design details and motivation
+
+### Motivation on the approaches
+
+There a multiple ways to perform the storage migration using libvirt. It is possible to simply copy the storage to the new destination and then swap the original disk with the new one. This approach is the most efficient as the overhead is caused only by the storage copy.
+
+A different approach is to simultaneously move the storage and live-migrate the guest. When the destination storage is not available on the host where the VM is executing, this solution is required. For this case, costs are greater since there is storage as well as live VM migration overhead.
+
+As KubeVirt is designed today and due to Kubernetes design principles of volume pod immutability, it isn't possible to hotplug and unplug new volumes while the pod is still running. KubeVirt already supports volume hotplug by attaching a new volume to a dummy pod and then bind mounting the storage in the `virt-launcher` pod. Sadly, this method fails to address the issue of removing the old PVCs that were expressly declared in the virt-launcher pod.
+Declaring all volumes hotpluggable could be a potential remedy, but we prefer to avoid it because it has been shown to be fragile for some storage providers and opaque to Kubernetes.
+
+The only workable and generic solution for this restriction appears to be to live migrate the VM with storage. Live migration involves the creation of a new virt-launcher pod by KubeVirt that has the destination PVCs attached.
+
+This migrating to a new storage provider is an exceptional operation, and very resource intensive. If in the future, the k8s community will consider the volume mutability in the pod spec, then the lighter approach could be taken into consideration.
+
+## Limitations
+
+### Unsupported volumes
+
+Certain types of volumes and disks aren't suitable to be migrated or the feature isn't supported yet. Those migrated volumes will be marked as rejected in the VMI status
+
+The current types of rejected volumes are:
+  * Hotpluggable volumes, this case will be considered in a later iteration.
+  * Filesystem volumes, currently virtiofs doesn't support live-migration.
+  * Shareable volumes, if a disk is shared between multiple VMs, it cannot be safely copied.
+  * LUNs disks, originally the disk might support SCSI protocol but the destination PVC class does not. This case will be considered in a later iteration.
+
+### Datavolumes
+
+DataVolumes are used to import the image during VM provisioning, but they may essentially be thought of as PVCs for the rest of the VM life cycle. With this API, we don't support live updates for the migration strategy to datavolumes. Therefore, the users can use this update strategy only when they replace datavolumes with PVCs.
+
+If the VM makes use of datavolume templates, then the template associated with the source volume needs to be removed by the user as well.
+
+Example using datavolumes:
+
+Original VM definition:
+```yaml
+---
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: vm-dv
+spec:
+  dataVolumeTemplates:
+  - metadata:
+      name: src-pvc
+    spec:
+      pvc:
+        accessModes:
+        - ReadWriteOnce
+        volumeMode: Filesystem
+        resources:
+          requests:
+            storage: 2Gi
+        storageClassName: local
+      source:
+        registry:
+          url: docker://registry:5000/kubevirt/alpine-container-disk-demo:devel
+[..]
+    spec:
+      domain:
+        devices:
+          disks:
+          - disk:
+              bus: virtio
+            name: datavolumedisk1
+
+      volumes:
+      - dataVolume:
+          name: src-pvc
+        name: datavolumedisk
+```
+
+Valid updated VM spec:
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: vm-dv
+spec:
+[..]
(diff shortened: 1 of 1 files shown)


alicefr:
/cc @mhenriks @awels @acardace @vladikr 

Hopefully, this time will be the good one :)

design-proposals/volume-update-strategy.md
```diff
@@ -0,0 +1,216 @@
...
+* the user updates the volumes on a vm.spec.template.spec.domain.volumes
+* the VM controller determines the volumes which needs to be updated 
+* the VM controller validates the volumes if they can be migrated
+* the VM controller updates the volumes accordingly in the active vmi.Spec
+* the VM controller writes the required volumes to migrate in the vmi.Status and set the update condition in the VMI status
+* the VMI workload-update controller detects the update condition and triggers the VMI migration
+* the Migration controller constructs a new target pod with the desired volumes
+* virt-launcher updates the volumes to migrate in the migration request for LibVirt
+* the VMI live migrates
+* the VM status is updated to reflect that the volume update action have completed
```
acardace, line 95: here maybe you want to mention that the update condition will be removed.

design-proposals/volume-update-strategy.md
```diff
@@ -0,0 +1,216 @@
+# Overview
```
acardace, line 1: You should mention somewhere that this feature relies on the `vmRolloutStrategy` field set to `LiveUpdate`, which also needs the `VMLiveUpdateFeatures` feature gate to be enabled.

fabiand (commented):
This looks like a ... tangible path forward. nice to see it converged on these api semantics,

design-proposals/volume-update-strategy.md
```diff
@@ -0,0 +1,216 @@
...
+
+      volumes:
+      - persistentVolumeClaim:
+          name: dst-pvc
+        name: datavolumedisk
+```
+
+## Scalability
+
+This proposal shouldn't have any impact on scalability.
```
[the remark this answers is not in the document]
alicefr: Thanks @fabiand, i'll add a note to that

design-proposals/volume-update-strategy.md
alicefr: Here, I'm a bit unsure how to proceed. We introduce already the new field `updateVolumesStrategy`. Do we want also to force the user to set the `vmRolloutStrategy`? I would also use a separate feature gate

alicefr:
@awels @mhenriks could you please check the part related to the datavolumes and tell me if you agree with the limitations

design-proposals/volume-update-strategy.md
```diff
@@ -0,0 +1,218 @@
...
+  * Hotpluggable volumes, this case will be considered in a later iteration.
+  * Filesystem volumes, currently virtiofs doesn't support live-migration.
+  * Shareable volumes, if a disk is shared between multiple VMs, it cannot be safely copied.
+  * LUNs disks, originally the disk might support SCSI protocol but the destination PVC class does not. This case will be considered in a later iteration.
+  
+### Datavolumes
+
+DataVolumes are used to import the image during VM provisioning, but they may essentially be thought of as PVCs for the rest of the VM life cycle. With this API, we don't support live updates for the migration strategy to datavolumes. Therefore, the users can use this update strategy only when they replace datavolumes with PVCs.
+
+If the VM makes use of datavolume templates, then the template associated with the source volume needs to be removed by the user as well.
```
mhenriks, line 130: Should we note that removing the template does not delete the datavolume?  Also if they want to delete the datavolume should wait until migration complete?

mhenriks:
Are you planning to how migration status is reported?

design-proposals/volume-update-strategy.md
alicefr: I can add it

fabiand:
Great question by @mhenriks - Yes. How do we track progress?

alicefr:
> Are you planning to show how migration status is reported?

The status is the status of the VM migration. I don't think we show today the percentage of the completion in the VMI migration object, but we could add it.

fabiand:
> The status is the status of the VM migration.

How do we know which migration is related to a spec change?

> I don't think we show today the percentage of the completion in the VMI migration object, but we could add it.

Maybe we do not want to solve this today (we could not solve it in the past 3 years or so ;) )

alicefr:
> > The status is the status of the VM migration.
> 
> How do we know which migration is related to a spec change?

I could put somehow this information in the VMI migration objct, like a label. However, I don't think we have this possibility for cpu and memory hotplug either. @acardace am I correct?

design-proposals/volume-update-strategy.md
```diff
@@ -0,0 +1,218 @@
...
+
+The volume migration update strategy follow the same [update sequence](https://github.com/kubevirt/community/blob/main/design-proposals/cpu-hotplug.md#phase-2-triggering-a-live-migration-when-the-number-of-sockets-changes-to-adjust-the-virt-launcher-cpu-resources) proposed for memory and cpu hotplug.
+
+Sequence of events if the `updateVolumesStrategy` is migration:
+
+* the user updates the volumes on a vm.spec.template.spec.domain.volumes
+* the VM controller determines the volumes which needs to be updated 
+* the VM controller validates if the volumes can be migrated
+* the VM controller updates the volumes accordingly in the active vmi.Spec
+* the VM controller writes the required volumes to migrate in the vmi.Status and set the update condition in the VMI status
```
fabiand, line 90: Please define the condition which will be used.

Please - where approriate - also specify what the condition will state in case of progress, success (well, gone I guess), and failure

alicefr:
> Re-reading this proposal to some extent, so:
>
> * vol changes, LM is triggered --> here we really migh want to set an annotation to express the "cause". this is something that we should use in other areas as well (virtctl triggered --> cause virtctl, eviction triggered --> cause eviction, …)

Can we target this for a second iteration? I could add it to the implementation phase. We might want to spent more thoughts to the other cases where we trigger the migration like updates and node eviction.
I won't use an annotation but here rather a condition in the status of the VMI migration object

> * I suppose that if the LM fails, then this will be expressed in the condition? This design should define the condition which is being used.

I can add a condition on the VM that express if the volume update was successful. However, here, there are 2 types condition:
* The first is on the VMI for signal that it needs to update the volumes with the migration. This is what is described in the proposal and it is similar to [`HotVCPUChange` and `HotMemoryChange`](https://github.com/kubevirt/kubevirt/blob/main/staging/src/kubevirt.io/api/core/v1/types.go#L560-L563)
* The second should the result of the update. AFAIU, for cpu and memory hotplug we don't describe the status of the migration but rather we leave the first condition on the VMI until we succeeded the update. @acardace, please, correct me here. This part isn't included in the proposal. Do we want also an additional condition that describes the update result?

design-proposals/volume-update-strategy.md
alicefr: Please, see my comment https://github.com/kubevirt/community/pull/260#issuecomment-1983111884

alicefr:
@fabiand @mhenriks I was suggesting [here](https://github.com/kubevirt/community/pull/260#issuecomment-1982897920) to use a label because it is something we can filter. With the new proposal with the update strategy it isn't straightforward to identify which migration is associated to the volume update.

design-proposals/volume-update-strategy.md
alicefr: Clarify this offline and added on the proposal

design-proposals/volume-update-strategy.md
alicefr: Note here, it is the VMI controller not the VM

design-proposals/volume-update-strategy.md
```diff
@@ -0,0 +1,218 @@
...
+* the VMI workload-update controller detects the update condition and triggers the VMI migration
+* the Migration controller constructs a new target pod with the desired volumes
+* virt-launcher updates the volumes to migrate in the migration request for LibVirt
+* the VMI live migrates
+* the VM status is updated to reflect that the volume update action have completed and the update condition is removed
```
alicefr, line 95: This is the VMI not the VM controller

design-proposals/volume-update-strategy.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Goals
+Design and develop an API to live update and migrate PVCs.
+  * The volume migration needs to be declarative and [GitOps](https://kubevirt.io/user-guide/operations/gitops/) compatible
+  * Volume migration implies an update of the volumes used by the VM. Hence, we want to trigger the storage migration on a VM update.
+  * The VM spec updates needs to be triggered by the user and not to be implicitly updated by the KubeVirt controller
+
+ ## Non  Goals
+*  This feature is not directly aimed at live migrating the VM using RWO volumes. Hence, the API covers and refers only to the volume update.
+*  This design doesn't cover migration between storage classes, but rather implements basic APIs that can be used by overlying tooling.
```
mhenriks, line 16: I would maybe leave this out?  Migrating between storageclasses is one of the main goals even if it is not a detail that is important to the implementation

design-proposals/volume-update-strategy.md
```diff
@@ -0,0 +1,218 @@
...
+
+
+## Goals
+Design and develop an API to live update and migrate PVCs.
+  * The volume migration needs to be declarative and [GitOps](https://kubevirt.io/user-guide/operations/gitops/) compatible
+  * Volume migration implies an update of the volumes used by the VM. Hence, we want to trigger the storage migration on a VM update.
+  * The VM spec updates needs to be triggered by the user and not to be implicitly updated by the KubeVirt controller
+
+ ## Non  Goals
+*  This feature is not directly aimed at live migrating the VM using RWO volumes. Hence, the API covers and refers only to the volume update.
```
mhenriks, line 15: This is actually the first I'm hearing of this limitation.  But I guess I'm not surprised.  What will happen if user creates a RWO volume by mistake?

mhenriks (commented):
I think it would be helpful for this doc to provide an example of how a user can monitor the status of volume migration i.e. find the matching migration object and interpret the status

design-proposals/volume-update-strategy.md
alicefr: @mhenriks maybe the sentence isn't clear. I don't think this isn't something new. What I wanted to say is that this feature doesn't enable general VM live-migration with RWO volumes. If your VM uses RWO volumes and the user want to live migrate (standard one, with the VMI migration creation object) then they still get an error message that the volume needs to use RWX mode.
This feature only cover the case when you want to change PVC not just migrating the VM

design-proposals/volume-update-strategy.md
alicefr: Here, I wanted to mentioned that this API doesn't deal with storage classes. We started this work by tring to design an API directly using storage classes and then remove it. I'd like to mention to the user that they don't need to expect any storage class notion with this feature. Does this make sense?

design-proposals/volume-update-strategy.md
alicefr: I tried to be more explicit here, please, check if now the sentence is more clear

alicefr:
> I think it would be helpful for this doc to provide an example of how a user can monitor the status of volume migration i.e. find the matching migration object and interpret the status

@mhenriks I added an example how we can use a label to get the corresponding VMI migration. The information regarding the update and storage migration can be inferred from the status of the VMI migration object,

alicefr:
@fabiand @mhenriks @acardace are you satisfied with the current state of the design? I'd like to start implementing it.

acardace:
/approve

I am, thanks Alice for the hard work and the patience especially.

acardace:
/lgtm

forgot I don't have approval rights here.

fabiand:
woot. :

design-proposals/volume-update-strategy.md
```diff
@@ -0,0 +1,229 @@
...
+* the VM controller determines the volumes which needs to be updated 
+* the VM controller validates if the volumes can be migrated
+* the VM controller updates the volumes accordingly in the active vmi.Spec
+* the VMI controller writes the required volumes to migrate in the vmi.Status and set the update condition in the VMI status
+* the VMI workload-update controller detects the update condition `VolumesChange` and triggers the VMI migration. When it creates the VMI migration object it also set the label `kubevirt.io/volume-update-migration: vm-name`.
+* the Migration controller constructs a new target pod with the desired volumes
+* virt-launcher updates the volumes to migrate in the migration request for LibVirt
+* the VMI live migrates
+* the VMI status is updated to reflect that the volume update action have completed and the update condition is removed
+
```
mhenriks, line 96: Question @aglitke brought up offline:  What happens if `updateVolumesStrategy` is changed from `migration` while a migration is in progress?

alicefr: This correspond to 2 VM spec versions. The answer depends is if the migrated volume is also update when the `updateVolumesStrategy` is changed.

If the migrated volumes aren't changed, then the VM will still migrate and it will have a condition that the change will be effective on restarts.

If the update concerns also one of the migrated volumes, the VM is still migrating and the volumes will be copied and the VM will be marked that it requires a restart for the new update. 
However, we might encounter a conflicts if the VM is shutdown while it is still migrating. I guess in this case, we can interpret it as an abortion an simply use the latest version of the VM and don't perform the volume migration.
The original PVC won't be deleted, so there isn't any loss of data and the user can always revert the VM definition.

Does this sound reasonable to you?

mhenriks: I'm not sure I understand.  Consider the following scenario:

1.  User sets `updateVolumesStrategy` to `migration`.  Removes volume A and adds volume B to the list of volumes.
2. Internally we start migrating volume A to Volume B
3. While migration is still in progress, the user edits the VM to change `updateVolumesStrategy` to `replace` and adds volume C

I assume migration from A to B continue?  C will be added when the VM reboots?

What if volume B is also removed is step 3?  Will the migration be aborted?

vladikr: > I'm not sure I understand. Consider the following scenario:
> 
> 1. User sets `updateVolumesStrategy` to `migration`.  Removes volume A and adds volume B to the list of volumes.
> 2. Internally we start migrating volume A to Volume B
> 3. While migration is still in progress, the user edits the VM to change `updateVolumesStrategy` to `replace` and adds volume C
> 
> I assume migration from A to B continue? C will be added when the VM reboots?

That's what I would expect too.

> 
> What if volume B is also removed is step 3? Will the migration be aborted?

Do we already have a way to cancel an ongoing volume migration?

alicefr: > I'm not sure I understand. Consider the following scenario:
> 
> 1. User sets `updateVolumesStrategy` to `migration`.  Removes volume A and adds volume B to the list of volumes.
> 2. Internally we start migrating volume A to Volume B
> 3. While migration is still in progress, the user edits the VM to change `updateVolumesStrategy` to `replace` and adds volume C
> 
> I assume migration from A to B continue? C will be added when the VM reboots?

Yes, the migration still continues and yes C will be added when the VM reboots
> 
> What if volume B is also removed is step 3? Will the migration be aborted?

The migration should be aborted only if we reboot the VM during the migration, otherwise, this should continue.

Is it  clearer?

alicefr: @mhenriks @vladikr @acardace after I started to implement the abortion part, I realized that allow any volume change while the VM is migrating can lead to a very confusing state. Therefore, I decided that we allow a volume change while we are migrating only if we revert the volumes to the previous one. Basically, changing the destination volumes back to the source ones. 
For cpu and memory hotplug, we return an error if the VMI is still migrating. Hence, this follow more or less the same pattern with the difference that the migration can be aborted if we return to the previous volume set.
Does this sound reasonable to you?

acardace: @alicefr yes, it sounds reasonable.

mhenriks: > we allow a volume change while we are migrating only if we revert the volumes to the previous one.

Say a VM has two volume A and B.  We start migrating A to C.  While that is in progress we decide to migrate B to D.  I assume that change will be blocked?

alicefr: Yes, while the update is in progress you can only change C -> A while C ->D will be rejected

alicefr: Changing C->A will abort the migration.

alicefr:
@vladikr @fabiand this proposal has already 2 lgtm. Would you like to have another look and eventually approve it?

design-proposals/volume-update-strategy.md
```diff
@@ -0,0 +1,234 @@
...
+This feature depends on the `vmRolloutStrategy` which needs to be set to `LiveUpdate`. On the long term, this should become the default behavior.
+
+The label `kubevirt.io/volume-update` can be used to filter and identify the VM migrations triggered by a volume update.
```
fabiand, line 99: `kubevirt.io/volume-update`

I'm not sure baout the semantics, can the label be more clear, i.e.
`kubevirt.io/consequence-of-volume-update`
or
`kubevirt.io/volume-update-in-progress`
or
`kubevirt.io/due-to-volume-update`

design-proposals/volume-update-strategy.md
```diff
@@ -0,0 +1,234 @@
...
+The volume migration update strategy follow the same [update sequence](https://github.com/kubevirt/community/blob/main/design-proposals/cpu-hotplug.md#phase-2-triggering-a-live-migration-when-the-number-of-sockets-changes-to-adjust-the-virt-launcher-cpu-resources) proposed for memory and cpu hotplug.
+
+Sequence of events if the `updateVolumesStrategy` is migration:
+
+* the user updates the volumes on a vm.spec.template.spec.domain.volumes
+* the VM controller determines the volumes which needs to be updated
+* the VM controller validates if the volumes can be migrated
+* the VM controller updates the volumes accordingly in the active vmi.Spec
+* the VMI controller writes the required volumes to migrate in the vmi.Status and set the update condition in the VMI status
+* the VMI workload-update controller detects the update condition `VolumesChange` and triggers the VMI migration. When it creates the VMI migration object it also set the label `kubevirt.io/volume-update-migration: vm-name`.
```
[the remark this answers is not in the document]
vladikr: @fabiand , it's not changed yet, this is in progress. 
Also, I prefer to keep it as suggested to be consistent with :
```
    VirtualMachineInstanceVCPUChange = "HotVCPUChange"
    // Indicates that the VMI is hot(un)plugging memory
    VirtualMachineInstanceMemoryChange = "HotMemoryChange"

```

vladikr:
It looks good to me already.
Let's wait for Fabian's comments to be resolved and we can get this in.

design-proposals/volume-update-strategy.md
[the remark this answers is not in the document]
fabiand: Thanks for clariying the semantics.
+1 on alignment with existing.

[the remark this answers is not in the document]
alicefr: @fabiand yes the naming is trying to be consistent with the cpu and memory change condition as Vladik pointed out

design-proposals/volume-update-strategy.md
alicefr: ok, I would go for `kubevirt.io/volume-update-in-progress`

alicefr:
I tried to addressed the last comments, PTAL, thanks!

design-proposals/volume-update-strategy.md
```diff
@@ -0,0 +1,234 @@
...
+
+The volume migration update strategy follow the same [update sequence](https://github.com/kubevirt/community/blob/main/design-proposals/cpu-hotplug.md#phase-2-triggering-a-live-migration-when-the-number-of-sockets-changes-to-adjust-the-virt-launcher-cpu-resources) proposed for memory and cpu hotplug.
+
+Sequence of events if the `updateVolumesStrategy` is migration:
+
+* the user updates the volumes on a vm.spec.template.spec.domain.volumes
+* the VM controller determines the volumes which needs to be updated
+* the VM controller validates if the volumes can be migrated
+* the VM controller updates the volumes accordingly in the active vmi.Spec
+* the VMI controller writes the required volumes to migrate in the vmi.Status and set the update condition in the VMI status
```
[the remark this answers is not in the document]
fabiand: Thanks, just one last request: Please add an example where we have a few disks which now require an update. I simply wonder how many disks are mentioned in the condition, I do wonder if this needs to be parseable.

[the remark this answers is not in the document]
alicefr: @fabiand do you mean in the condition? Right now, I'm not mentioning any volumes in the condition, the information of the change is reported under migrated `migratedVolumes` and this is simply a list. So, it would look like:
```yaml
  status:
    conditions:
    - lastProbeTime: null
      lastTransitionTime: "2024-04-09T07:51:30Z"
      message: migrate volumes
      status: "True"
      type: VolumesChange
[..]
    migratedVolumes:
    - destinationPVCInfo:
        claimName: dest-pvc-1
        volumeMode: Block
      sourcePVCInfo:
        claimName: src-pvc-1
        volumeMode: Filesystem
    - destinationPVCInfo:
        claimName: dest-pvc-2
        volumeMode: Block
      sourcePVCInfo:
        claimName: src-pvc-2
        volumeMode: Filesystem
``` 

Do you want me to add this example or do you mean something else?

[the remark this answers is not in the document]
fabiand: Perfect. And yes, I was missing this example.

Witht his I'm good.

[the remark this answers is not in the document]
alicefr: Ok. I added the second volume in the example, so it is clearer

fabiand:
Thanks lgtm!

vladikr:
/approve
Looks good to me as well. 
Thank you @alicefr !

PR #264, opened 2024-03-02
Proposal: KubeVirtBMC

starbops:

Propose [KubeBMC](https://github.com/starbops/kubebmc), enabling a way to do out-of-band management against VMs using IPMI/Redfish just like what we do to bare-metal servers, to the KubeVirt ecosystem.

Related issue: kubevirt/kubevirt#9108

**Checklist**

**Release note**:

```release-note
NONE
```

### DIFF
--- a/design-proposals/kubevirtbmc.md
+++ b/design-proposals/kubevirtbmc.md
@@ -1,0 +1,154 @@
+# Overview
+
+This proposal introduces [KubeVirtBMC](https://github.com/starbops/kubebmc), a BMC emulator for Kubernetes that facilitates the deployment of software/applications/platforms such as OpenShift and OpenStack - whose installers typically require communication with bare-metal out-of-band management protocols like IPMI and Redfish - in KubeVirt VMs for development, testing, and debugging purposes, similar to the functionality provided by [VirtualBMC](https://opendev.org/openstack/virtualbmc) and [sushy-emulator](https://opendev.org/openstack/sushy-tools) but within a Kubernetes context.
+
+## Motivation
+
+Developers require the ability to deploy applications or systems in local virtual environments like bare-metal ones. Existing solutions involve libvirt domains or QEMU VMs with BMC emulators, which are not directly compatible with KubeVirt due to its internal use of [`virtqemud`](https://libvirt.org/manpages/virtqemud.html), necessitating a Kubernetes-native solution. This was initially raised in the GitHub issue: [Expose a virtual BMC service per virtual machine, for IPMI and/or Redfish actions · Issue #9108 · kubevirt/kubevirt](https://github.com/kubevirt/kubevirt/issues/9108).
+
+## Goals
+
+- Provide essential BMC functionalities such as (rebooting, changing boot devices, and mounting virtual media) via IPMI and Redfish protocols against VMs backed by KubeVirt
+- Expose service endpoints for clients to access either in-cluster or externally
+- Enable deployment tools, e.g. [Metal3](https://metal3.io/), [Tinkerbell](https://tinkerbell.org/), [Bifrost](https://opendev.org/openstack/bifrost/), etc., to install software on VMs without adapting to the virtual environment by modifying their source code
+
+## Non Goals
+
+- Providing BMC functionalities for bare-metal machines
+
+## Definition of Users
+
+- Users who want to do out-of-band management to the VMs within a KubeVirt environment directly using bare-metal-like methods, e.g., IPMI or Redfish
+- Users who want to install software like OpenShift or OpenStack on VMs locally within a KubeVirt environment using existing deployment tools
+
+## User Stories
+
+- As a developer, I want to deploy a multi-node OpenShift cluster using its native installer in KubeVirt VMs.
+- As a tester, I need to quickly spin up and destroy OpenStack clusters in a virtual environment for testing purposes.
+
+## Repos
+
+No existing repos under the KubeVirt organization will be impacted. KubeVirtBMC, proposed to be a new project under the KubeVirt organization, augments the KubeVirt ecosystem as an optional add-on providing an operator and the emulated BMC implementation.
+
+# Design
+
+KubeVirtBMC will act as a bridge between KubeVirt VMs and the BMC functionalities, using Kubernetes services to expose these capabilities. KubeVirtBMC consists of two components:
+
+- **virt-bmc-controller**: A typical Kubernetes controller reconciles the VirtualMachineBMC, VirtualMachine, and Service objects
+- **virt-bmc**: A BMC simulator for serving IPMI and Redfish endpoints and translating the incoming requests to native KubeVirt API calls
+
+By manually creating the VirtualMachineBMC object, the `virt-bmc-controller` will watch and reconcile it to complete the BMC scaffold. This includes a BMC emulator Pod, i.e., `virt-bmc`, and its Service object working together to provide the BMC functionalities to various clients over the network. In addition, KubeVirtBMC could be configured to auto-create VirtualMachineBMC objects every time a new VirtualMachine is created.
+
+Inside the BMC emulator, the IPMI/Redfish requests are translated to native KubeVirt API calls and sent to the API server. It's essentially a combination of IPMI/HTTP server and Kubernetes/KubeVirt client.
+
+The following graph illustrates the entire KubeVirtBMC workflow:
+
+```mermaid
+flowchart LR
+    controller["kubevirtbmc-controller"]
+    cr["kubevirtbmc CR"]
+    kbmc-pod["kbmc Pod"]
+    kbmc-svc["kbmc Service"]
+    controller-.->|watches|cr
+    cr-.->|owns|kbmc-svc
+    cr-.->|owns|kbmc-pod
+    client--->|IPMI|kbmc-svc
+    kbmc-svc-->kbmc-pod
+    kbmc-pod-->|HTTP|apiserver
+    apiserver-->|modifies|vm
+    vm-->|creates|vmi
+```
+
+## API Examples
+
+The proposed VirtualMachineBMC CRD looks like the following:
+
+```yaml
+kind: VirtualMachineBMC
+metadata:
+  name: openshift-cluster-node-0
+spec:
+  # A Secret object reference for the IPMI/Redfish authentication.
+  # It should contain a pair of username and password.
+  authSecret:
+    namespace: default
+    name: bmc-cred
+  # The VM this emulated BMC manages
+  virtualMachine:
+    namespace: default
+    name: openshift-cluster-node-0
+```
+
+The condition of the created resources (Pod and Service) will be reflected under the `.status` section:
+
+```yaml
+status:
+  serviceIP: 10.53.23.74
+  conditions:
+  - lastUpdateTime: "2024-03-01T08:48:23Z"
+    status: "True"
+    type: BMCEmulatorReady
+  - lastUpdateTime: "2024-03-01T08:48:23Z"
+    type: EndpointReady
+    status: "True"
+```
+
+## Scalability
+
+KubeVirtBMC is designed to scale within Kubernetes environments, and the number of emulated BMC Pods and Services should increase as the number of VM increases. Hence, the overhead each Pod introduced to the entire system should also be considered.
+
+## Update/Rollback Compatibility
+
+The proposed feature does not introduce any breaking changes to KubeVirt itself, so the functionality of versions before this feature should not be impacted upon update/rollback.
+
+## Functional Testing Approach
+
+Utilizes Kubernetes testing frameworks to validate functionality across different deployment scenarios.
+
+Functional tests will simply:
+
+- Create a VM
+- Ensure the corresponding `kbmc` Pod and Serivce objects are created
+- Create a testing Pod with any container image that includes the `ipmitool`/`curl` binaries
+- Ensure the BMC emulator works as expected by sending IPMI/Redfish requests within the testing Pod to the BMC endpoint of the VM
+
+# Implementation Phases
+
+## Phase 1: Transfer the KubeBMC project to the KubeVirt organization
+
+The [KubeBMC project](https://github.com/starbops/kubebmc) already has a basic set of IPMI functionalities implemented:
+
+- Power on/off/reset
+- Power status retrieval
+- Boot device configuration (PXE or disk)
+
+However, some chores need to be done before transferring. For example, the CRD name should be changed from KubeBMC to a more specific one like VirtualMachineBMC, and the API group of the VirtualMachineBMC CRD needs to be replaced with a suitable name conforming to KubeVirt's convention. Also, the code needs to be protected by robust unit tests and integration test plans.
+
+## Phase 2: Align the implementation with the proposed design
+
+The current CRD is slightly different from the one proposed above:
+
+```yaml
+kind: KubeBMC
+metadata:
+  name: openstack-ubuntu-compute
+  namespace: kubebmc-system
+spec:
+  password: password
+  username: admin
+  vmName: compute
+  vmNamespace: openstack-ubuntu
+status:
+  ready: true
+  serviceIP: 10.53.1.159
+```
+
+Some work must be done to get align with the proposed design, including schema change and reconciling loop alteration.
+
+## Phase 3: Support Redfish
+
+KubeVirtBMC will support not only IPMI but also expose functionalities through the Redfish protocol. Additionally, it will enable virtual media operations for VMs, providing a new option for boot device selection.
+
+## Phase 4: Support events
+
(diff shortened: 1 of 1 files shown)


design-proposals/kubebmc.md
```diff
@@ -0,0 +1,156 @@
...
+- Users who want to install software like OpenShift or OpenStack on VMs locally within a KubeVirt environment using existing deployment tools
+
+## User Stories
+
+- As a developer, I want to deploy a multi-node OpenShift cluster using its native installer in KubeVirt VMs.
+- As a tester, I need to quickly spin up and destroy OpenStack clusters in a virtual environment for testing purposes.
+
+## Repos
+
+No existing repos under the KubeVirt organization will be impacted. KubeBMC augments the KubeVirt ecosystem as an add-on.
```
0xFelix, line 31:
```suggestion
No existing repos under the KubeVirt organization will be impacted. KubeBMC augments the KubeVirt ecosystem as an optional add-on providing an operator and the BMC implementation.
```

design-proposals/kubebmc.md
```diff
@@ -0,0 +1,156 @@
...
+
+# Design
+
+KubeBMC will act as a bridge between KubeVirt VMs and the BMC functionalities, using Kubernetes services to expose these capabilities. KubeBMC consists of two components:
+
+- **kubebmc-controller**: A typical Kubernetes controller reconciles the KubeBMC, VirtualMachine, and Service objects
+- **kbmc**: A BMC simulator for serving IPMI and Redfish endpoints and translating the incoming requests to native Kubernetes API calls
```
0xFelix, line 38:
```suggestion
- **kbmc**: A BMC simulator for serving IPMI and Redfish endpoints and translating the incoming requests to native KubeVirt API calls
```

design-proposals/kubebmc.md
```diff
@@ -0,0 +1,156 @@
...
+
+Each VirtualMachine will have its own KubeBMC object created by `kubebmc-controller` during the VM creation. The controller will further process the KubeBMC object to complete the BMC scaffold. This includes a BMC emulator Pod, i.e., `kbmc`, and its Service object working together to provide the BMC functionalities to various clients over the network.
+
+Inside the BMC emulator, the IPMI/Redfish requests are translated to native Kubernetes API calls and sent to the API server. It's essentially a combination of IPMI/HTTP server and Kubernetes client.
```
0xFelix, line 42:
```suggestion
Inside the BMC emulator, the IPMI/Redfish requests are translated to native KubeVirt API calls and sent to the API server. It's essentially a combination of IPMI/HTTP server and Kubernetes/KubeVirt client.
```

design-proposals/kubebmc.md
```diff
@@ -0,0 +1,156 @@
...
+    cr-.->|owns|kbmc-svc
+    cr-.->|owns|kbmc-pod
+    client--->|IPMI|kbmc-svc
+    kbmc-svc-->kbmc-pod
+    kbmc-pod-->|HTTP|apiserver
+    apiserver-->|modifies|vm
+    vm-->|creates|vmi
+```
+
+It's also possible for VMs being created to opt out of such a functionality. This could be done by annotating the VirtualMachine object with a specific key-value pair.
```
0xFelix, line 62: I would prefer to make this opt-in instead of opt-out. Can the automatic creation of BMCs be controlled e.g. by a filter set on the controllers CRD? For a start we could keep creation of BMC CRDs a manual task?

design-proposals/kubebmc.md
```diff
@@ -0,0 +1,156 @@
...
+    status: "True"
+    type: BMCEmulatorReady
+  - lastUpdateTime: "2024-03-01T08:48:23Z"
+    type: EndpointReady
+    status: "True"
+```
+
+## Scalability
+
+KubeBMC is designed to scale within Kubernetes environments, and the number of emulated BMC Pods and Services should increase as the number of VM increases. Hence, the overhead each Pod introduced to the entire system should also be considered.
```
0xFelix, line 100: I wonder if we could make this a sidecar to the VM?

design-proposals/kubebmc.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Phase 1: Transfer the KubeBMC project to the KubeVirt organization
+
+The [KubeBMC project](https://github.com/starbops/kubebmc) already has a basic set of IPMI functionalities implemented:
+
+- Power on/off/reset
+- Power status retrieval
+- Boot device configuration (PXE or disk)
+
+However, some chores need to be done before transferring. For example, the API group of the KubeBMC CRD needs to be replaced with a suitable name conforming to KubeVirt's convention. Also, the code needs to be protected by robust unit tests and integration test plans.
```
0xFelix, line 127: I would also consider renaming the CRDs. E.g. `VirtualMachineBMC`?

design-proposals/kubebmc.md
```diff
@@ -0,0 +1,156 @@
+# Overview
+
+This proposal introduces [KubeBMC](https://github.com/starbops/kubebmc), a BMC emulator for Kubernetes that facilitates the deployment of software/applications/platforms such as OpenShift and OpenStack - whose installers typically require communication with bare-metal out-of-band management protocols like IPMI and Redfish - in KubeVirt VMs for development, testing, and debugging purposes, similar to the functionality provided by [VirtualBMC](https://opendev.org/openstack/virtualbmc) and [sushy-emulator](https://opendev.org/openstack/sushy-tools) but within a Kubernetes context.
+
+## Motivation
+
+Developers require the ability to deploy OpenShift and OpenStack clusters on local virtual environments. Existing solutions involve libvirt domains or QEMU VMs with BMC emulators, which are not directly compatible with KubeVirt due to its internal use of [`virtqemud`](https://libvirt.org/manpages/virtqemud.html), necessitating a Kubernetes-native solution. This was initially raised in the GitHub issue: [Expose a virtual BMC service per virtual machine, for IPMI and/or Redfish actions · Issue #9108 · kubevirt/kubevirt](https://github.com/kubevirt/kubevirt/issues/9108).
```
0xFelix, line 7: Is it the only use case?

0xFelix (commented):
Many thanks for following up on this! I added some comments.

design-proposals/kubebmc.md
starbops: It could be done like that. There's a VM reconciler ensuring the 1:1 mapping between VirtualMachine and KubeBMC objects. By the "filter set" you mentioned above, did you mean that a one-to-many relationship is preferable?

design-proposals/kubebmc.md
starbops: Yes. Then the emulated BMC could operate the VM by directly communicating with `virtqemud` via the Unix socket. No communication with `kube-apiserver` is needed anymore. However, the entire architecture will be different from the current design of KubeBMC and more coupled with KubeVirt itself because users might need to determine if they want a virtualized BMC sidecar container created alongside the VM at the beginning.

That's my initial thoughts about it. Maybe it's not that complicated. Please correct me if there's anything wrong. Thank you.

design-proposals/kubebmc.md
starbops: Good idea! It makes more sense because many people get confused when they hear KubeBMC. "Is it an emulated BMC for the entire Kubernetes cluster?" something like that :P

design-proposals/kubebmc.md
starbops: AFAIK relevant projects like Virtual BMC and sushy-tools were created to test bare-metal provision-related tools/libraries. I would say it's the primary use case.

design-proposals/kubebmc.md
0xFelix: I meant a filter with which you can select the VMs a BMC is created for, e.g. a label selector. The filter could be configured in the CRD/CR of the BMC controller.

design-proposals/kubebmc.md
0xFelix: Thinking about it again, this might not be so easy. Because a sidecar can only be attached to a VMI (a running VM), while the BMC also needs to be available when the VM is stopped. (how do you start it otherwise?)

Also, you are right, it is more decoupled right now and should probably stay like that to keep the VM "lean".

design-proposals/kubebmc.md
0xFelix: Exactly, I think this way it would better fit into the naming scheme of existing KubeVirt CRDs.

design-proposals/kubebmc.md
0xFelix: Just thinking, if we could add some more. But it's fine for now!

design-proposals/kubebmc.md
starbops: We could, but the current implementation is a 1:1 mapping with a namespaced name reference between the BMC and VM. If a filter-like label selector is used instead, it implies a BMC could map to more than one VM. We need to multiplex them using different ports. That's a different approach.

Back to the original topic, do you think adding an annotation to the VM CRD to opt-in is a good idea?

design-proposals/kubebmc.md
starbops: Nice catch! If we go with the sidecar design, powering on an already shutdown VM via BMC wouldn't be possible. The BMC serves as an out-of-band manner for users/programs to manage the host itself; that's how it was initially targeted.

design-proposals/kubebmc.md
lentzi90: > For a start we could keep creation of BMC CRDs a manual task?

This sounds most reasonable to me to start with. No "magic" annotations, less complexity and surprises at the cost of slightly more work for the user creating the custom resources.

design-proposals/kubebmc.md
lentzi90: Maybe we can generalize it a bit (e.g. we want the ability so simulate baremetal environments in general, not just OpenStack and OpenShift) or add some more examples. I saw Metal3 mentioned already, it could be added here also. We would like to use it to simulate baremetal kubernetes cluster through CAPI/CAPM3.

lentzi90:
Hi, Metal3.io maintainer here :wave: 
I'm very interested in this as it might let us run our e2e tests on KubeVirt VMs. Currently we rely on home baked bash scripts and playbooks for setting up libvirt VMs + sushy-tools and similar.
What can I do to help this move forward?

For us working on metal3, I would say that redfish is much more important than ipmi. Virtualmedia would also be very important. Since KubeBMC could simplify our e2e tests greatly, I think we can also help out with development and maintenance to some degree :slightly_smiling_face:

design-proposals/kubebmc.md
0xFelix: Did I understand correctly, that in your proposal `kubebmc-controller` is supposed to create a BMC CRD for every VM? I was thinking we should make `kubebmc-controller` configurable by adding a label selector to its configuration, so the label selector controlls for which VMs a BMC CRD will be created. I would choose labels, since Kubernetes intends labels to be used to identify objects. 

As suggested and backed by @lentzi90 we could implement that at a later stage though.

0xFelix:
Hi @lentzi90,

next step would be to finalize this proposal. If there are no further objections and the proposal is accepted, we could start by setting up a repository under the KubeVirt org and migrating the original project to its new home. Basically we need to take the first two steps as lined out in the proposal.

I will bring this up with the other maintainers to see if there are any further thoughts or concerns.

design-proposals/kubebmc.md
```diff
@@ -0,0 +1,156 @@
+# Overview
+
+This proposal introduces [KubeBMC](https://github.com/starbops/kubebmc), a BMC emulator for Kubernetes that facilitates the deployment of software/applications/platforms such as OpenShift and OpenStack - whose installers typically require communication with bare-metal out-of-band management protocols like IPMI and Redfish - in KubeVirt VMs for development, testing, and debugging purposes, similar to the functionality provided by [VirtualBMC](https://opendev.org/openstack/virtualbmc) and [sushy-emulator](https://opendev.org/openstack/sushy-tools) but within a Kubernetes context.
```
lyarwood, line 3: There's a discussion about the CRD naming below but I wonder if the project itself needs to be something a little less generic to k8s and more specific to KubeVirt to avoid people getting the wrong idea here? `KubeVirtBMC`?

design-proposals/kubebmc.md
```diff
@@ -0,0 +1,156 @@
...
+The proposed KubeBMC CRD looks like the following:
+
+```yaml
+kind: KubeBMC
+metadata:
+  name: openshift-cluster-node-0
+spec:
+  # A Secret object reference for the IPMI/Redfish authentication.
+  # It should contain a pair of username and password.
+  authSecretRef:
```
lyarwood, line 75: Do we need `Ref` in these attribute names?

design-proposals/kubebmc.md
lyarwood: +1 to default to manual CR creation to begin with and adding label selectors etc later.

design-proposals/kubebmc.md
```diff
@@ -0,0 +1,156 @@
...
+- Users who want to install software like OpenShift or OpenStack on VMs locally within a KubeVirt environment using existing deployment tools
+
+## User Stories
+
+- As a developer, I want to deploy a multi-node OpenShift cluster using its native installer in KubeVirt VMs.
+- As a tester, I need to quickly spin up and destroy OpenStack clusters in a virtual environment for testing purposes.
+
+## Repos
+
+No existing repos under the KubeVirt organization will be impacted. KubeBMC augments the KubeVirt ecosystem as an optional add-on providing an operator and the BMC implementation.
```
lyarwood, line 31: I was under the impression you wanted to house the project under the KubeVirt org?

0xFelix: IIUC this tries to express that the project can live under the KubeVirt umbrella but will not impact other repositories. It is purely optional, you can choose to deploy it on your KubeVirt cluster but you don't have to.

design-proposals/kubebmc.md
0xFelix: A good idea, similar to the renaming of the CRDs, the project itself should also be identifiable as part of KubeVirt.

starbops: `KubeVirtBMC` makes sense. Thanks for the suggestion.

design-proposals/kubebmc.md
starbops: I forgot to mention Metal3 here :doh:
Yeah I think in this paragraph it's better to have a general description.

design-proposals/kubebmc.md
starbops: > Did I understand correctly, that in your proposal kubebmc-controller is supposed to create a BMC CRD for every VM?

Yes, the `kubebmc-controller` in the current implementation will create a BMC CRD (and thus a pair of Pod and Service) for each VM, so it's a one-to-one mapping. If we change the specifier from a namespaced name to a label selector, does that mean a one-to-many mapping is preferable?

design-proposals/kubebmc.md
starbops: We could get rid of it :)

design-proposals/kubevirtbmc.md
```diff
@@ -0,0 +1,156 @@
...
+## Repos
+
+No existing repos under the KubeVirt organization will be impacted. KubeVirtBMC, proposed to be a new project under the KubeVirt organization, augments the KubeVirt ecosystem as an optional add-on providing an operator and the emulated BMC implementation.
+
+# Design
+
+KubeVirtBMC will act as a bridge between KubeVirt VMs and the BMC functionalities, using Kubernetes services to expose these capabilities. KubeVirtBMC consists of two components:
+
+- **vmbmc-controller**: A typical Kubernetes controller reconciles the VirtualMachineBMC, VirtualMachine, and Service objects
+- **vmbmc**: A BMC simulator for serving IPMI and Redfish endpoints and translating the incoming requests to native KubeVirt API calls
```
0xFelix, line 38: Other controllers / Pods in KubeVirt have a named prefixed with `virt-`. Maybe it makes sense to prefix these names too? E.g. `virt-bmc-controller` and `virt-bmc`?

lyarwood: Wouldn't that suggest they were somehow core to KubeVirt and maybe managed by virt-operator?

0xFelix: True, but that argument could apply to the CRD names too?

starbops: The CRD has now been renamed `VirtualMachineBMC`. `KubeVirtBMC` is the project name. I'm okay with renaming the components if that follows the convention of the KubeVirt ecosystem.

vladikr:
/approve

Thanks @0xFelix! A separate repo with its own set of functional tests is a way to go.

PR #281, opened 2024-03-31
network binding plugin: DownwardAPI for device info

AlonaKaplan:
Design proposal to extend the network binding plugin entry in the kubevirt CR with an option for a downward API.
When specified, a downwardAPI from the virt-launcher pod to the binding sidecar will be added.
Currently the only supported downward API is "device-info", exposing the a map between the logical network name to its network status annotation device info.

Fixes #

**Checklist**

**Release note**:

```release-note

```

### DIFF
--- a/design-proposals/network-binding-plugin/network-binding-plugin.md
+++ b/design-proposals/network-binding-plugin/network-binding-plugin.md
@@ -188,6 +188,7 @@ the binding is referenced by name with optional additional arguments.
         - sidecarImage
         - domainAttachment (a standard domain definition that exists in
           the core . E.g `tap`, `sriov`).
+        - downwardAPI (E.g `device-info`)
 - VMI Specification CR:
     - Define a new binding subtree under spec-domain-devices-interfaces.
       Each interface has a binding subtree that contains a required name
@@ -256,6 +257,76 @@ VMI status and in one case to pass the data (processed) to the
 `virt-launcher`.
 
 - DownwardAPI
+
+##### Device Info
+
+In some cases the multus status annotation
+contains [device-info](https://github.com/k8snetworkplumbingwg/device-info-spec/blob/main/SPEC.md).
+This info may be valuable for the binding plugin sidecar. For example
+`SRIOV` and `VDPA` plugins need this info.
+
+As the multus status annotation is not there yet on `virt-launcher` pod
+creation, the suggested solution is to use downwardAPI volume and mount
+it to the sidecar.
+The downwardAPI volume will track a new `virt-launcher` pod
+annotation `kubevirt.io/network-info` that will contain the `network-info`
+of the networks that have `downwardAPI`.
+Very similar to the
+existing [SRIOV downwardAPI](https://github.com/kubevirt/kubevirt/pull/8226)
+solution which is mapping the networkName to the `pci-address` of the
+device and mounting the downwardAPI volume to the `virt-launcher` pod.
+
+**Note**: Multus provides device-info information since version v3.7.
+If an older version is used, the downward API is not supported,
+the annotation will have empty values.
+
+Alternatives:
+
+- Instead of configuring in the kubevirt CR whether deviceInfo should be
+  mounted to the sidecar or not, we can pass via the downward API the device
+  info of all the interfaces that have it.
+  The downside in this approach is that a DownwardAPI volume will be added to
+  the pod even when it is not actually needed.
+- Passing `PCIDEVICE_<RESOURCE_NAME>_NET_INFO` env var from the
+  `virt-launcher` to the sidecar by adding it as an annotation to the
+  vmi object passed to the onDefineDomain method. The problem with this
+  approach is that it will reintroduce the
+  same [bug](https://github.com/kubevirt/kubevirt/issues/7444)
+  that the SRIOV downward API solved.
+- Sharing
+[CNIDeviceInfo](https://github.com/k8snetworkplumbingwg/device-info-spec/blob/main/SPEC.md#5-new-cni-plugin-arguments)
+file with the sidecar. It is a very similar approach to the
+downwardAPI option, the disadvantage here is that the user would have
+to specify the file path in the `NetworkAttachmentDefinition`.
+- Instead of using an annotation and DownwardAPI, extend the VMI interface status
+API to contain the interface device info.
+
+  Pros:
+  - DownwardAPI volume is not required, the VMI is already passed to the virt-launcher
+  compute and sidecar containers.
+
+  Cons:
+  - The VMI CRD should be changed.
+    The device infromation is exposed to users as a contract,
+    while the pod annotation with the DownwardAPI is an implementation detail with no such contract.
+  - It should be easy to move from the pod annotation to VMI Status in the future,
+  but we cannot go the opposite way.
+  - Keeping pod specific information in the VMI status may be raceful over migration.
+  - Users that have read permissions on the VMI will be able to see the device info.
+- Using DownwardAPI to pass the whole multus status annotation.
+
+   Pros:
+   - Additional annotation is not required.
+
+   Cons:
+   - The multus status annotation contains networking information about all the pod interfaces.
+  It may be unsecure to pass this info to the sidecar.
+   - Extending the network info to contain information from other sources besides the multus status
+  annotation will be impossible.
+
+- Additional alternatives to a similar issue are listed in the SRIOV
+mapping
+design [doc](https://github.com/kubevirt/community/blob/main/design-proposals/sriov-interfaces-mapping.md).
 
 #### Run services in the virt-launcher pod
 
@@ -941,3 +1012,154 @@ Libvirt configuration snippet:
     <rom enabled='no'/>
 </interface>
 ```
+
+## Appendix F: Concrete example for vDPA binding
+
+Present a concrete example of the actions needed at each integration
+point, considering the option suggested in the
+[design section](#design-details).
+
+### Kubevirt CR Specification
+
+The vDPA binding is registered to the Kubevirt CR.
+The plugin is then referenced by name from the VM Specification.
+
+> **Note**: The vDPA binding does not need to perform changes in the
+> pod network, therefore, there is no need for a
+> network-attachment-definition.
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+metadata:
+  name: kubevirt
+  namespace: kubevirt
+spec:
+  configuration:
+    network:
+      binding:
+        vdpa:
+          sidecarImage: quay.io/kubevirt/network-vdpa-binding
+          downwardAPI: device-info
+```
+
+### VM Specification
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: vmi-1-bridge
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          interfaces:
+            - name: blue
+              mac: 12:34:56:78:9a:bc
+              binding:
+                name: vdpa
+      networks:
+        - name: blue
+          multus:
+            networkName: vdpa-network
+```
+
+### NetworkAttachmentDefinition Specification
+
+Not required for vDPA binding
+
+> **Note**: Defining a NetworkAttachmentDefinition to connect the pod to the node vDPA interface is required,
+> this NetworkAttachmentDefinition is specified in the VM spec.
+> No additional NetworkAttachmentDefinition is required, because there is no need to configure the pod network
+> stack for vDPA.
+
+### Webhook Admitter
+
+- The core KubeVirt admitter assures that only one method of binding
+  definition is possible: Either through the Interface API or Network Binding Plugin API.
+- No addmiter is provided as part of the vDPA binding.
+
+### Pod Definition
+
+The VMI controller updates the pod manifest per the VMI specification.
+
+- From the VMI spec, the following snippet is detected:
+
+```yaml
+name: blue
+binding:
+  name: vdpa
+```
+
+- As a result, the pod is updated with the following annotations & spec:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: virt-launcher-123
+  annotations:
+    kubevirt.io/network-info: '{
+      "interfaces": [{
+        "network": "blue"
+        "device-info": {
+          "type": "vdpa",
+          "version": "1.1.0",
+          "vdpa": {
+            "parent-device": "vdpa:0000:65:00.3",
+            "driver": "vhost",
+            "path": "/dev/vhost-vdpa-1",
+            "pci-address": "0000:65:00.3"
+          }
+        }
+      }]
+    }'
+  spec:
+    containers:
+      - name: compute
+
+      - name: hook-sidecar-1
+        volumeMounts:
+          - name: network-info
+            mountPath: /etc/pod-info
+    volumes:
+      - name: network-info-annotation
+        downwardAPI:
+          items:
+            - path: network-info
+              fieldRef:
+                fieldPath: metadata.annotations[kubevirt.io/network-info]
+```
+
+### Configure Pod network namespace
+
+Not required for vDPA binding
+
+### Reacting to CNI results (Multus status annotation)
+
+Device info from the multus status annotation will be propagated to the
+sidecar using downwardAPI. The data will be used to compose the libvirt
+domain xml interface.
+
+### Run services in the virt-launcher pod
+
+Not required for vDPA binding
+
+### Domain definition & creation
+
+A sidecar will be introduced to add the vDPA interfaces to the libvirt
+domain.
+
+Libvirt configuration snippet:
+
+```xml
+<interface type='vdpa'>
+    <driver name='vfio'/>
+    <source dev='/dev/vhost-vdpa-1 network=default'/>
+      <alias name='ua-blue'/>
+      <address type='pci' domain='0x0000' bus='0x65' slot='0x00' function='0x3'/>
+</interface>
+```
+
+- The domain itself is created by the `virt-launcher`.


design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -257,6 +258,48 @@ VMI status and in one case to pass the data (processed) to the
...
+existing [SRIOV downwardAPI](https://github.com/kubevirt/kubevirt/pull/8226)
+solution which is mapping the networkName to the `pci-address` of the
+device and mounting the downwardAPI volume to the `virt-launcher` pod.
+
+> **Note**: Old multus versions may not report the device-info.
+
+Alternatives:
+
+- Instead of configuring in the kubevirt CR whether deviceInfo should be
+  mounted to the sidecar or not, we can always mount it in case it exists
```
ormergi, line 285: > we can always mount it in case it exists in the status

This is not clear could please rephrase it to be more explicit?

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -257,6 +258,48 @@ VMI status and in one case to pass the data (processed) to the
...
+  in the status. The downside in this approach is that a DownwardAPI
+  volume will be added to the pod even when it is not actually needed.
+- Passing `PCIDEVICE_<RESOURCE_NAME>_NET_INFO` env var from the
+  `virt-launcher` to the sidecar by adding it as an annotation to the
+  vmi object passed to the onDefineDomain method. The problem with this
+  approach is that it will reintroduce the
+  same [bug](https://github.com/kubevirt/kubevirt/issues/7444) SRIOV
```
ormergi, line 292:
```suggestion
same [bug](https://github.com/kubevirt/kubevirt/issues/7444)
```

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -941,3 +984,151 @@ Libvirt configuration snippet:
     <rom enabled='no'/>
 </interface>
 ```
+
+## Appendix F: Concrete example for vDPA binding
+
+Present a concrete example of the actions needed at each integration
+point, considering the option suggested in the
+[design section](#design-details).
```
ormergi, line 992: leftover?

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -941,3 +984,151 @@ Libvirt configuration snippet:
...
+```
+
+### NetworkAttachmentDefinition Specification
+
+Not required for vDPA binding
+
+> **Note**: Just to clarify, a NAD is required for the vDPA CNI and
+> device plugin, but it is not in the scope of this doc. An extra NAD
+> for
+> the network binding plugin is not required.
```
ormergi, line 1047:
```suggestion:
Defining a net-attach-def to connect the pod to the node vDPA interface is required, the that is spesifed in the VM spec.
No additional net-attach-def required, because there is no need to configure the pod network stack for vDPA.
```

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -941,3 +984,151 @@ Libvirt configuration snippet:
...
+### Webhook Admitter
+
+- The core KubeVirt admitter assures that only one method of binding
+  definition is possible: Either the old style method or the new one.
```
ormergi, line 1051:
```suggestion
  definition is possible: Either through the Interface API or Network Binding Plugin API.
```

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -941,3 +984,151 @@ Libvirt configuration snippet:
...
+### Pod Definition
+
+The VMI controller updates the pod manifest per the VMI specification.
+
+- From the VMI spec, the following snippet is detected:
+
+```yaml
+name: blue
+binding:
+  name: vdpa
```
ormergi, line 1063: It would be easier to understand if the at least the interface stanza is mentioned:
VMI spec.domain.devices.interfaces:
```
interfaces:
- name: blue
  binding:
    name: vdpa
```

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -941,3 +984,151 @@ Libvirt configuration snippet:
...
+            "driver": "vhost",
+            "path": "/dev/vhost-vdpa-1",
+            "pci-address": "0000:65:00.3"
+          }
+        }
+    }'
+  spec:
+    containers:
+      - name: compute
+
```
ormergi, line 1090: nit:
```suggestion
...
```

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -941,3 +984,151 @@ Libvirt configuration snippet:
...
+      - name: hook-sidecar-1
+        volumeMounts:
+          - name: network-device-info-map
+            mountPath: /etc/pod-info
```
ormergi, line 1094: How about mounting the device info at its own directory? maybe `/etc/net-device-info`?

Note that the current SR-IOV mapping annotation is mounted to `/etc/pod-info`

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -941,3 +984,151 @@ Libvirt configuration snippet:
...
+          }
+        }
+    }'
+  spec:
+    containers:
+      - name: compute
+
+      - name: hook-sidecar-1
+        volumeMounts:
+          - name: network-device-info-map
```
ormergi, line 1093: Please consider removing the `map` prefix, it doesnt add value in the context of volumeMount name.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -257,6 +258,48 @@ VMI status and in one case to pass the data (processed) to the
...
+In some cases the multus status annotation
+contains [device-info](https://github.com/k8snetworkplumbingwg/device-info-spec/blob/main/SPEC.md).
+This info may be valuable for the binding plugin sidecar. For example
+`SRIOV` and `VDPA` plugins need this info.
+
+As the multus status annotation is not there yet on `virt-launcher` pod
+creation, the suggested solution is to use downwardAPI volume and mount
+it to the sidecar.
+The downwardAPI volume will track a new `virt-launcher` pod
+annotation `kubevirt.io/network-device-info-map` that will contain a
```
ormergi, line 272: Please consider removing map prefix, IMO it doesn't add value.
Also if tomorrow we decide to change the annotation value format (to not be map) the name would no longer be valid.

ormergi (commented):
First look

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -257,6 +258,48 @@ VMI status and in one case to pass the data (processed) to the
...
+mapping between the logical network name to the `device-info` of this
+network.
+Very similar to the
+existing [SRIOV downwardAPI](https://github.com/kubevirt/kubevirt/pull/8226)
+solution which is mapping the networkName to the `pci-address` of the
+device and mounting the downwardAPI volume to the `virt-launcher` pod.
+
+> **Note**: Old multus versions may not report the device-info.
```
EdDev, line 280: Please specify the version and what will happen if it is missing.

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev, line 272: I would prefer a generalization to avoid the need to create multiple values that track different annotations. We already have a good example with the existing SR-IOV one.

It should be harmes to generalize it as long we have a proper structure and a general name.
The only disadvantage is the embedded json, but we have it now anyway.

Here is a potential structure:
Annotation name: `kubevirt.io/virt-launcher-data`
Annotation value format:
```json
{
  "network": {
    "interfaces": [
      "name": "net1"
      "networkLogicalName": <logical name>
      "device-info": {...}
    ]
  }
}
```

In any case, please explicitly define the format here.

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev: What leftover do you see here?

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev: This the format used on all other examples, I do not think it is that important to break the convention.
A different PR can suggest changing the convention to all if you would like to invest in this.

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev: It will make the MD interpretation break and look weird, I do not recommend adding it.

EdDev (commented):
Thank you.

My main feedback here is to consider a generalization of the information, to avoid creating many mounting points for each need. Other needs to pass information from the pod resource to the application running in the pod can ride on this same annotation & mount point.

design-proposals/network-binding-plugin/network-binding-plugin.md
ormergi: lines 990-992, it doesn't say anything, just describing what it should.

design-proposals/network-binding-plugin/network-binding-plugin.md
ormergi: Note that its inside "```yaml```" block, it shouldnt break.

design-proposals/network-binding-plugin/network-binding-plugin.md
AlonaKaplan: @EdDev 
How about?
```
type VirtLauncherData struct {
	networkDeviceInfoMap map[string]*networkv1.DeviceInfo
}
```

design-proposals/network-binding-plugin/network-binding-plugin.md
ormergi: I recall it was Multus that stated to reflect PCI info since version v3.7
https://github.com/kubevirt/community/blob/main/design-proposals/sriov-interfaces-mapping.md#example-1 
See the note in this section

design-proposals/network-binding-plugin/network-binding-plugin.md
AlonaKaplan: done.

design-proposals/network-binding-plugin/network-binding-plugin.md
AlonaKaplan: It is intentionally under the same dir. The path is different - /etc/pod-info/device-info

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev:
```go
type VirtLauncherDownwardData struct {
	Network *Network `json:"network,omitempty"`
}
type Network struct {
	Interfaces []Interface `json:"interfaces,omitempty"`
}

type Interface struct {
	Name               string `json:"name"`
	NetworkLogicalName string `json:"networkLogicalName"`
	DeviceInfo         string `json:"deviceInfo,omitempty"`
}
```

AlonaKaplan: I agree, I prefer to keep it narrow as well.
If in the future there will be a need to extend it, it should be easy, the volume will have to point to a new annotation.
The only downside is that we will have to keep the old annotation for a while to keep backward compatibility for old launchers that are already running.

AlonaKaplan: The fact it is a map is part of the API. I cannot change it to be something else without breaking backward compatibility.

design-proposals/network-binding-plugin/network-binding-plugin.md
AlonaKaplan: Thanks Or, updated the doc.

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev: > Here we try to enable a way to pass data from virt-controller (who has access to virt-launcher pod annotation) to virt-launcher app. The suggestion to generalize the annotation may sounds appealing and future proof, it may seem like a mere annotation, but it actually become an internal API; communication channel between virt-controller and virt-launcher.

Indeed, it is a structured, stable and open for extension protocol.
It is not a path focused on one single problem that needs to be solved, it looks at the general need and defines it well.

> Currently we have similar channel, between virt-handler and launcher, using gRPC and protobuf (handler-launche-com) Unlike the `handler-launcher-com`, this annotation can easily be abused by the next dev and becoming a "go to workaround" when introducing an API change require more time and work in a tight time-frame. I am talking about cases when there is a need to pass data from virt-controller to virt-launcher and its possible using `handler-launcher-com ` / API change.

I do not accept this as an argument. Abusing a channel and protocol is always possible, it cannot be the reasoning for not establishing one.

With the same reasoning I can claim the usage of this DownwardAPI mechanism is an abuse, giving ideas to others to use the same for other means.

> Keep in mind that we struggle to keep track and apply proper review process to formal and stable APIs (hopefully it will get better with sig-api group). I am afraid this annotation will become hard to maintain and become another backward-compatibility thing to worry about.

This is true for any API change, no matter where it is placed.

> If in the future there will be a need to extend it, it should be easy, the volume will have to point to a new annotation.
> The only downside is that we will have to keep the old annotation for a while to keep backward compatibility for old launchers that are already running.

I do not think the notion of adding an API, even if internal, and admitting it may be replaced if an extension will be needed is a good direction. This implies revolution is better than evolution in a supposed-to-be stable project API.

I think the data communication channel should be formatted correctly to allow it to extend.
I disagree with narrowing its scope as argued above.
But as with the previous incident (SR-IOV DownwardAPI), I will not block the whole solution on this.

EdDev: I do not think the meaning of `map` is universally understood.
In the `Go` programming language a key-value structure is named `map`, in other languages it is called a `dict` (Python).
In plain English it describes an are of land/sea.

Even in programming we try to avoid expressing the type of a variable in the name. We usually prefer the name to represent what the data contains and not how it contains it.

I also think the `map` is redundant.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -941,3 +987,151 @@ Libvirt configuration snippet:
...
+
+### NetworkAttachmentDefinition Specification
+
+Not required for vDPA binding
+
+> **Note**: Defining a NetworkAttachmentDefinition to connect the pod to the node vDPA interface is required,
+> this NetworkAttachmentDefinition is specified in the VM spec.
+> No additional NetworkAttachmentDefinition is required, because there is no need to configure the pod network
+> stack for vDPA.
+### Webhook Admitter
```
EdDev, line 1051: nit: Newline is need above.

EdDev (commented):
Overall it looks ready. I think you should un-draft it.

I had only one naming comment that matters, but it is not a blocker.
I will wait for a response and then lgtm this.

Thanks.

design-proposals/network-binding-plugin/network-binding-plugin.md
AlonaKaplan: done.

design-proposals/network-binding-plugin/network-binding-plugin.md
ormergi: Also commented on the impl. PR 
I would store the new device-info annotation in different dir then the sriov-network-pci one.
IMO it will be easier to debug that way, also avoid potential overwrites, if tomorrow someone changes the downward API volumes render code.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -257,6 +258,50 @@ VMI status and in one case to pass the data (processed) to the
...
+In some cases the multus status annotation
+contains [device-info](https://github.com/k8snetworkplumbingwg/device-info-spec/blob/main/SPEC.md).
+This info may be valuable for the binding plugin sidecar. For example
+`SRIOV` and `VDPA` plugins need this info.
+
+As the multus status annotation is not there yet on `virt-launcher` pod
+creation, the suggested solution is to use downwardAPI volume and mount
+it to the sidecar.
+The downwardAPI volume will track a new `virt-launcher` pod
+annotation `kubevirt.io/network-info` that will contain the `network-info`
```
ormergi, line 272: Changing it to network-info immediately raise question "why passing network info and not any other info?"
I would keep the scope narrow as before and have the annotation provide device-info only.

It seems that other network info already passed to the virt-launcher app through other channels (VMI status/gRPC, guest-agent), the only missing one was the device-info that up until the SR-IOV mapping bug would be given by the device-plugin.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -257,6 +258,50 @@ VMI status and in one case to pass the data (processed) to the
...
+of the networks that have `downwardAPI`.
+Very similar to the
+existing [SRIOV downwardAPI](https://github.com/kubevirt/kubevirt/pull/8226)
+solution which is mapping the networkName to the `pci-address` of the
+device and mounting the downwardAPI volume to the `virt-launcher` pod.
+
+**Note**: Multus provides device-info information since version v3.7.
+If an older version is used, the downward API is not supported,
+the annotation will have empty values.
+
```
ormergi, line 282: I think it will be nice defining the data model of the introduced annotation value, including what info it provides (PCI address..) in which format (map, json..)

Also when this annotation and downward API volume are be added (for every VM or only for those who require plugin sidecar)

design-proposals/network-binding-plugin/network-binding-plugin.md
AlonaKaplan: The reason I changed it is that in the CR we define - `DownwardAPI: device-info`. Meaning the API is extendable and we may add other DownwardAPI type. I want the CR API to be align to the pod one.

design-proposals/network-binding-plugin/network-binding-plugin.md
AlonaKaplan: > I think it will be nice defining the data model of the introduced annotation value, including what info it provides (PCI address..) in which format (map, json..)

The vDPA example has it.

> Also when this annotation and downward API volume are be added (for every VM or only for those who require plugin sidecar)

It is written - `The downwardAPI volume will track a new `virt-launcher` pod
annotation `kubevirt.io/network-info` that will contain the `network-info`
of the networks that have `downwardAPI`.`
I find the exact details of when the volume is added are implementation details and shouldn't be part of a high level design.

EdDev (approved):
Thank you!

design-proposals/network-binding-plugin/network-binding-plugin.md
ormergi: >The vDPA example has it.

Right but it is just an example I think it should be well defined.
Anyhow I am not insisting on it the proposal looks good.

> I find the exact details of when the volume is added are implementation details and shouldn't be part of a high level design.

At first I thought it could become backward-comparability concern, but it shouldn't be an issue:
On Kubevirt upgrade, for existing pods, virt-controller will just add the new annotation / new annotation value.
The annotation value wont propagate to the launcher app until the VM is migrated, 
which is fine since it cause no disruption.
New pods will be created with new annotation  volume.

ormergi (approved):
LGTM

design-proposals/network-binding-plugin/network-binding-plugin.md
AlonaKaplan: It is an API, so yes, a change should be considered as an API change.

EdDev:
This design has been discussed in the `sid-api` meeting on the 2nd of April and some questions have been raised there, most answered directly (with a summary in the meeting notes).

@alaypatel07 ,  if there is any input you will like to add or have more questions, please let us know.
We will like to make sure nothing was missed and that we exhausted all other options to solve this need.

alaypatel07:
It looks like there is an API field already in VMI status to reflects the network information xref: https://github.com/kubevirt/api/blob/main/core/v1/types.go#L224

The information from the above mentioned fields collected and populated from different sources that is reflected from here: https://github.com/kubevirt/kubevirt/blob/c7d255d3345c80e8c6b054638187fdd730997b04/pkg/network/setup/netstat.go#L80-L87

Is there a possibility to add a [source](https://github.com/kubevirt/api/blob/main/core/v1/types.go#L661) called pod-annotation and populate the device-info from pod annotation to vmi status in virt-controller?

How will the affect migration? 

It looks like the migration controller is already patching VMI status, it is possible to clear the device info added in the interface status https://github.com/kubevirt/kubevirt/blob/3dbf033c27af8d39749e8360dba81d79478aa89e/pkg/virt-controller/watch/migration.go#L1292. Once the information is cleared, the virt-controller already has the ability to recognize that this is a migration from hotplugging feature, see here https://github.com/kubevirt/kubevirt/blob/de14b8127a54d8599ad3852feccad92d20c7f08b/pkg/virt-controller/watch/vm.go#L3196-L3195. A similar logic could be put into place where the info source is updated with new pod annotations.

with the above context the algorithm would be:

virt-controller changes:
1. check if the vmi is migrating, 
   1. if migrating and device-info is missing, add device-info to vmi.status.interfaces fields from new pod
   2. if not migrating and device-info is missing add info to vmi.status.interfaces fields from the pod
2. migration-controller changes:
   1. before creating the target pod, clear the vmi.status.interfaces fields with device-info fields
   2. <need to handle cases where migration is stuck>

side-car changes:
- get the device info information from vmi.status.interfaces

AlonaKaplan:
> It looks like there is an API field already in VMI status to reflects the network information xref: https://github.com/kubevirt/api/blob/main/core/v1/types.go#L224
> 
> The information from the above mentioned fields collected and populated from different sources that is reflected from here: https://github.com/kubevirt/kubevirt/blob/c7d255d3345c80e8c6b054638187fdd730997b04/pkg/network/setup/netstat.go#L80-L87
> 
> Is there a possibility to add a [source](https://github.com/kubevirt/api/blob/main/core/v1/types.go#L661) called pod-annotation and populate the device-info from pod annotation to vmi status in virt-controller?
> 

We already have an info-source for it, it is called `multus-status`.

> How will the affect migration?
> 
> It looks like the migration controller is already patching VMI status, it is possible to clear the device info added in the interface status https://github.com/kubevirt/kubevirt/blob/3dbf033c27af8d39749e8360dba81d79478aa89e/pkg/virt-controller/watch/migration.go#L1292. Once the information is cleared, the virt-controller already has the ability to recognize that this is a migration from hotplugging feature, see here https://github.com/kubevirt/kubevirt/blob/de14b8127a54d8599ad3852feccad92d20c7f08b/pkg/virt-controller/watch/vm.go#L3196-L3195. A similar logic could be put into place where the info source is updated with new pod annotations.
> 
> with the above context the algorithm would be:
> 
> virt-controller changes:
> 
>     1. check if the vmi is migrating,
>        
>        1. if migrating and device-info is missing, add device-info to vmi.status.interfaces fields from new pod
>        2. if not migrating and device-info is missing add info to vmi.status.interfaces fields from the pod
> 
>     2. migration-controller changes:
>        
>        1. before creating the target pod, clear the vmi.status.interfaces fields with device-info fields
>        2.
> 
> 
> side-car changes:
> 
>     * get the device info information from vmi.status.interfaces

@alaypatel07 Thanks for looking into it.
Managing device-info on the VMI status may be hard and raceful over migration/migration cancellation.
To avoid this race/complication I suggest that each status interface will contain a list of multus info. Each multus info will include the virt-launcher pod name.

```
type VirtualMachineInstanceNetworkInterface struct {
...
    []MultusInfo `json:"multusInfo,omitempty"`
}

type MultusInfo struct {
    podName string `json:"podName,omitempty"`
    DeviceInfo *v1.DeviceInfo `json:"deviceInfo,omitempty"`
}
```


virt-launcher compute container has an environment variable `POD_NAME` that it can use to find out the relevant device info from the status (we will have to expose this env var to the sidecar as well).

The problems I see with this approach over the downward API are -
1. We expose in the VMI status internal information that is not necessary reflects the VMI interface status.
2. From the security perspective - we expose the device-info to any user that has permissions to view the VMI. With the downwardAPI only users with `pod/exec` or pod viewing permissions can see the device info.

What do you think?

alaypatel07:
@AlonaKaplan can you please enumerate the kind of races will be presented with migration workflow?

The way I see it is, if the migration controller can clear the device information before the migration starts, the migration target will have the right device information.

In case of migration cancellation the migration controller will back the device information from target back into status. 

May be with above approach it will be safe to achieve synchronization during migration but please highlight the gap I may have missed.



Regarding security: we are just shifting the security model from reading pod annotations to reading vmi status.  Can the admin have the same safety measures through rbac for vmi status as for pod annotation?

AlonaKaplan:
> @AlonaKaplan can you please enumerate the kind of races will be presented with migration workflow?

I don't have an exact scenario, but the fact that we will have one reconcile loop that will be responsible to update the status with the devices info information from the current pod and another routine that will be responsible for cleaning this info upon some event (migration, migration abort) sounds raceful to me.
If we will decide to proceed with this approach despite of the security issues I can spend some time to think about a specific race that may happen/whether it is solvable.

> 
> The way I see it is, if the migration controller can clear the device information before the migration starts, the migration target will have the right device information.
> 
> In case of migration cancellation the migration controller will back the device information from target back into status.
> 



> May be with above approach it will be safe to achieve synchronization during migration but please highlight the gap I may have missed.
> 
> Regarding security: we are just shifting the security model from reading pod annotations to reading vmi status. Can the admin have the same safety measures through rbac for vmi status as for pod annotation?

The sidecar is not reading the pod annotation, a DownwardAPI volume is used to pass the pod annotation to a file inside the sidecar.
VM users that have permissions to see the pod status usually don't have permissions to see the pod annotations or exec into a pod container (and see the file).

EdDev:
> > @AlonaKaplan can you please enumerate the kind of races will be presented with migration workflow?
> 
> I don't have an exact scenario, but the fact that we will have one reconcile loop that will be responsible to update the status with the devices info information from the current pod and another routine that will be responsible for cleaning this info upon some event (migration, migration abort) sounds raceful to me. If we will decide to proceed with this approach despite of the security issues I can spend some time to think about a specific race that may happen/whether it is solvable.

A simple translation to the above can be: When the migration controller creates the target pod and at the same time resets the device-info information from the VMI status, the VMI controller 1 ns later patches the VMI status with current device-info update from the source pod.

Another example could be that the migration fails in the middle, and the source VM needs to be restored. usually no special action is needed, but in the case of SR-IOV and potentially vDPA, the devices need to be hotplug back at the domain level and we need the correct device-info information (not the one from the target). I do not think we can assure the information has been updated already by the VMI controller while the virt-handler performs the reconciliation on the VM.

We have the Migration, VM and VMI controller, then the two virt-handlers and two virt-launchers. All need to work in parallel and do their thing without overriding each other.
If we build the solution with strong assumptions on how these components reconcile, we create a strong coupling on the low level details, causing the system to become fragile and sensitive to changes. 

An attempt to react in an edge triggered manner is problematic in a distributed system like Kubernetes.
The fact that a change occurs at a specific time, is not a guaranty that it can be persisted when other clients reconcile the same data.

> Edit: we also don't know what each sidecar is doing with the device info and when. So I find it problematic the vmi status will contain the device info of the target pod as long as the source pod sidecars are still alive.

I agree.
The assumption that the information is processed or not, at the sidecar post domain creation, is dependent on the binding plugin implementation and the call to it from the virt-launcher.

> > The way I see it is, if the migration controller can clear the device information before the migration starts, the migration target will have the right device information.
> > In case of migration cancellation the migration controller will back the device information from target back into status.
> 
> > May be with above approach it will be safe to achieve synchronization during migration but please highlight the gap I may have missed.

I think there is an assumption above that the action of setting a value (or reset it) is observed synchronously all over the system components. The system is asynchronous in nature, it takes time for the data to be "seen" by other components and in many cause one can override the other when working on the same data.
Usually that is ok, because the system reconcile to a stable state, but I do not think this is the case here. You will basically loose the "event" of starting a migration (the event caused the reset of the data).

> > Regarding security: we are just shifting the security model from reading pod annotations to reading vmi status. Can the admin have the same safety measures through rbac for vmi status as for pod annotation?
> 
> The sidecar is not reading the pod annotation, a DownwardAPI volume is used to pass the pod annotation to a file inside the sidecar. VM users that have permissions to see the pod status usually don't have permissions to see the pod annotations or exec into a pod container (and see the file).

I would argue that regular VM users have only access to the VM CR in their namespace, they have no permission to see low-level components like Pods. Maybe the admin will provide users access to VMIs as well.
I guess we can assume VM access is provide to users, then maybe access to VMI is given and finally I would expect no access to Pods is provided.

But I am unsure how important is this security concern compared to the usage of DownwardAPI.

There is another API level point: Exposing the information through the VMI Status requires a VMI CRD change, exposing new fields in a stable API. This means that we will not be able to change the format in a breaking way once the network binding plugin is GA. I think that when using an annotation, and not documenting it as an external API, we have more freedom to change.

We have cons and pros on both approaches.
@AlonaKaplan , can we add this alternative in the design and perhaps compare their pros & cons?

EdDev:
> But I am unsure how important is this security concern compared to the usage of DownwardAPI.

I have received feedback about this "security" offline, which worth sharing:
It does not need to be considered a security problem, it is a layering problem.
The use of VMI Status exposes the information to users as a contract, while the pod annotation with the DownwardAPI is an implementation detail with no such contract.
One is formal, one is not.

I pretty much tend to agree that having a contract for information that is needed only for internal usage is less appealing. We have no potential client to the VMI Status that is interested in this information, it is used only by the application inside the pod, which we cannot refer to as a client (it is internal).

We are able to move from the pod annotation to VMI Status in the future, but we cannot go the opposite way.
Therefore, IMO, it will be better to proceed with the current suggestion.

EdDev (approved):
Thanks!

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -941,3 +1001,154 @@ Libvirt configuration snippet:
...
+      - name: hook-sidecar-1
+        volumeMounts:
+          - name: network-info
+            mountPath: /etc/pod-info
+    volumes:
+      - name: network-info
+        downwardAPI:
+          items:
+            - path: device-info
```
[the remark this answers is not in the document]
AlonaKaplan: Typo, the path should be `network-info`.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -941,3 +1001,154 @@ Libvirt configuration snippet:
...
+metadata:
+  name: kubevirt
+  namespace: kubevirt
+spec:
+  configuration:
+    network:
+      binding:
+        vdpa:
+          sidecarImage: quay.io/kubevirt/network-vdpa-binding
+          downwardAPI: device-info
```
alaypatel07, line 1031: implementation wise, should this be a list of strings? do you anticipate that there could be more than one metadata information that could be passed to sidecar network plugins?

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -941,3 +1001,154 @@ Libvirt configuration snippet:
...
+
+- As a result, the pod is updated with the following annotations & spec:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: virt-launcher-123
+  annotations:
+    kubevirt.io/network-info: '{
```
alaypatel07, line 1092: If the same information is available as a different annotation `k8s.v1.cni.cncf.io/network-status`, out of curiosity what problem does having a kubevirt specific annotation solve?

https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/how-to-use.md#launch-pod-with-text-annotation

design-proposals/network-binding-plugin/network-binding-plugin.md
AlonaKaplan: I think that in the future there might be other data that a plugin would require but I don't think that a combination of more than one type will be required. And even if it would be (unlikely), we can always introduce a new type that represents a combination.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -257,6 +258,65 @@ VMI status and in one case to pass the data (processed) to the
...
+to specify the file path in the `NetworkAttachmentDefinition`.
+- Instead of using an annotation and DownwardAPI, extend the VMI interface status
+API to contain the interface device info.
+
+  Pros:
+  - DownwardAPI volume is not required, the VMI is already passed to the virt-launcher
+  compute and sidecar containers.
+
+  Cons:
+  - The VMI CRD should be changed.
```
alaypatel07, line 309: I do-not agree with the first two points here, see here: https://github.com/kubevirt/community/pull/281#issuecomment-2049651323

design-proposals/network-binding-plugin/network-binding-plugin.md
alaypatel07: As an example in the future consider that the network sidecar needs  `kubevirt.io/foo` annotation along with `kubevirt.io/network-info` as metadata in sidecar. How would this API evolve?

design-proposals/network-binding-plugin/network-binding-plugin.md
AlonaKaplan: The network-status annotation contains information about all the pod interfaces, including IPs and other confidential data.The new annotation collects from the status only the device info of the interfaces that use a plug-in that supports it, the other interface and other info besides the device-infi is not passed.
We find this approach more secure and avoiding sidecar users to get the complete pod networking information.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -257,6 +258,65 @@ VMI status and in one case to pass the data (processed) to the
...
+    The device infromation is exposed to users as a contract,
+    while the pod annotation with the DownwardAPI is an implementation detail with no such contract.
+  - It should be easy to move from the pod annotation to VMI Status in the future,
+  but we cannot go the opposite way.
+  - Keeping pod specific information in the VMI status may be raceful over migration.
+  - Users that have read permissions on the VMI will be able to see the device info.
```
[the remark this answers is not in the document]
AlonaKaplan: User that has read permission on the VMI not necessary has read permissions on the pod, therefore, with the downward API option, this kind of user won't be able to see the device info.

design-proposals/network-binding-plugin/network-binding-plugin.md
AlonaKaplan: `downwardAPI: device-info-foo` or comma separated `downwardAPI: deviceinfo,foo`. Or we can introduce `downwardAPIs`. Although I don't think this kind of combination will ever be required. I cannot think of a plugin that will need information from several pod fields.  I don't even believe that anything besides deviceinfo will be required, but kept this option open.

design-proposals/network-binding-plugin/network-binding-plugin.md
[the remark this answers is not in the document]
alaypatel07: Similarly, could it be possible that some users dont have access to VMI read but have access to read pods? 

All I am saying is that depending on the cluster configuration, some could have read access and other could not. We cannot make an assumption that one solution is more secure than other because it is largely dependent on cluster configuration

design-proposals/network-binding-plugin/network-binding-plugin.md
AlonaKaplan: Done, added it to the alternatives sections.

design-proposals/network-binding-plugin/network-binding-plugin.md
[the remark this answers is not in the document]
AlonaKaplan: We can say that the suggested option doesn't give any user more sensitive information than it already have.
The device info was already exposed to the pod via an environment variable by the device plugin. The downward API is needed just to create the mapping between the device info to the network using it.
But the virt-launcher is already familiar with the networks and with the device info.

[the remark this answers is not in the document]
alaypatel07: That sounds like an accurate statement, +1

alaypatel07:
This design was discussed again in sig-api call and @EdDev was able to answer all of the questions raised there. The summary is here:

```
[alaypatel07] there are two things happening here
- Virt-controller adding an annotation [kubevirt.io/device-info](http://kubevirt.io/device-info) to virt-launcher pod
- Virt-controller adding the annotation as downward API to the pod for the sidecar
- Can we implement the sidecar solution generically? I.e. any sidecar that needs the annotation can mention in kubevirt spec and will be available as downward api volume mount
   - [edwardh] The need for this feature has not been expressed. If needed in the future the networking binding plugin will use the new api
[alaypatel07] why cant we use the multus annotation and have to introduce a new [kubevirt.io/device-info](http://kubevirt.io/device-info) annotation?
- [edwardh] It is so that the information that goes into network sidecar is controlled regardless of multus annotation. If the multus annotation breaks compatibility, network sidecar will not be affected by it
```

Given this, going to unhold the design. Thanks folks @AlonaKaplan and @EdDev for allowing this to be discussed in the sig-api call.

/unhold

design-proposals/network-binding-plugin/network-binding-plugin.md
ormergi: This reminds me of K8S Pod status API where the IPS  filed was added once IP was not sufficient.
IMO it is not nice to add the another filed just to support multiple options, I would go with sting slice or support coma separated filed in the future (not optimal).

AlonaKaplan:
Thanks @alaypatel07 

Raising 2 lgtms to approve.

/approve

EdDev (approved):

PR #283, opened 2024-04-04
design-proposals: Allow custom guest CPU topologies within instancetypes 

lyarwood:

At present users of the `instancetype.kubevirt.io/v1beta1` API and CRDs are able to define a CPU topology for a `VirtualMachine` through a combination of vCPUs provided by the `guest` attribute of an instance type and an optional topology provided by the `preferredCPUTopology` attribute of a preference.

This design proposal aims to extend the API to allow for an optional guest CPU topology to be provided by a referenced instance type instead of a generic number of vCPUs.

Fixes #

This PR also introduces a dedicated `instancetype.kubevirt.io` directory for design proposals related to the API and CRDs.

**Checklist**

**Release note**:

```release-note
NONE
```


design-proposals/instancetype.kubevirt.io/custom-cpu-topology.md
```diff
@@ -0,0 +1,124 @@
...
+  "sockets": 1,
+  "threads": 1
+}
+```
+
+This design proposal aims to extend the API to allow for an optional guest CPU topology to be provided by a referenced instance type instead of a generic number of vCPUs.
+
+## Motivation
+
+The initial design of instance types and preferences was heavily influenced by hyper scalers and other large scale IaaS projects. Such environments allow users to select a generic amount of vCPUs without specifying exactly how these are then mapped to the underlying guest. While a useful starting point KubeVirt is now attracting traditional virtualization users and use cases that demand both certainty around how these vCPUs are mapped to the guest while also demanding specific topologies not easily expressed by our current API.
```
fabiand, line 48: Where is our current instanceType support not sufficient?

I'm missing to see the use-case where a specific virtual (!) CPU topology/layout in the instanceType is helpful.
Also because we can not enforce/request this layout from the physical side (aka there is no scheduler support to request this topology).

Use-cases would help to understand why we want to add this complexity to instanceTypes.

lyarwood: > Where is our current instanceType support not sufficient?

We've had requests to accommodate topologies such as `spec.cpu.topology.{sockets:3,cores: 2,threads: 1}` that we can't easily express with the current generic vCPU amount in the instance type and preferredCPUTopology preference.

> I'm missing to see the use-case where a specific virtual (!) CPU topology/layout in the instanceType is helpful. Also because we can not enforce/request this layout from the physical side (aka there is no scheduler support to request this topology).
>
> Use-cases would help to understand why we want to add this complexity to instanceTypes.

RT and DPDK are the two use cases where this has come up. In both we've had downstream teams request to be able to create instance types with specific guest CPU topologies based on the host while using dedicated CPU placement etc.

I'm aware that doesn't guarantee anything but from an instance type API point of view providing an optional direct pass through of a complete guest CPU topology from the instance type to the VMI is valid IMHO.

fabiand: Is there an email thread or design by the RT folks in order to get the full picture of their requirements?

lyarwood: @fabiand so after some back and fourth with a few people it looks like the only thing really missing from the current API is an ability to spread vCPUs across sockets, cores *and* threads in order to mimic SMT within the guest. We shouldn't need to allow custom topologies for this and I think we can get away with extending the existing `preferSpread` `preferredCPUTopology`.

A very rough draft of this idea is available here https://github.com/kubevirt/kubevirt/pull/11729 and I'll also follow up with a design proposal. As a result I'm going to close out this PR.

/close

fabiand: And for the record: I am technically not against adding the a specific topology to instanceTypes.

However, the combination of `cpu.guest` and startegy is much more powerful, because it makes it simpler for higher level systems (such as UI and bindings) and eventually for the user to deal with.

A user does not need to understand the HOW and WHY if they "simply" want to increase the compute capacity.

PR #293, opened 2024-05-17
Design-proposal: KubeVirt DRA design proposal

rthallisey:
DRA ([Dynamic Resource Allocation](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/)) design proposal.

#### What this PR does / why we need it:
KubeVirt DRA integration will allow VM users fine-grained control of devices.  This is important for many use-cases:
- Full Device Advertisement 
In the current device-plugin model, GPUs can only be advertised in one mode (Passthrough, MIG, vGPU) for the full lifecycle of the device.  This model doesn’t work when an end-user has the ability to choose which type of device they need - PT, vGPU, or MIG.
Advertising full devices is a common use-case in ephemeral workloads like HPC, ML, Cloud Gaming, ect.., that require on-demand driver switches on the GPU, based on the type of workload requested.
- Pluggable CPU Management
Similar to the GPU, the CPU can be logically separated and handed out to workloads.  Conceptually the use-case would be the same as GPUs - allow a plugin to handle the specialized allocation of the CPU.  Similarly, need memory management as well.  We should be allowing full platform management for specialized vertical solutions.

DRA is not a device-plugin replacement - it solves more problems.

#### Release note:
```release-note
None
```

### DIFF
--- a/design-proposals/dra.md
+++ b/design-proposals/dra.md
@@ -1,0 +1,161 @@
+# Overview
+This proposal is about adding supporting DRA (dynamic resource allocation) in KubeVirt.
+DRA allows vendors fine-grained control of devices.  The device-plugin model will continue
+to exist in Kubernetes, but DRA will offer vendors more control over the device topology.
+
+## Motivation
+DRA adoption is important for KubeVirt so that vendors can expect the same
+control of their devices using Virtual Machines and Containers.
+
+## Goals
+- Align on how KubeVirt will consume external and in-tree DRA drivers
+- Align on what drivers KubeVirt will support in tree
+
+## Non Goals
+- Replace device-plugin support in KubeVirt
+
+## Definition of Users
+A user is a person that wants to attach a device to a VM
+
+## User Stories
+- As a user, I would like to use my GPU dra driver with KubeVirt
+- As a user, I would like to use KubeVirt's default driver
+
+## Repos
+kubevirt/kubevirt
+
+# Design
+
+## API Examples
+
+### VM API with PassThrough GPU
+
+```
+apiVersion: resource.k8s.io/v1alpha2
+kind: ResourceClass
+name: gpu.resource.nvidia.com
+driverName: gpu.resource.nvidia.com
+---
+apiVersion: rtx4090.gpu.resource.nvidia.com/v1
+kind: ClaimParameters
+name: rtx4090-claim-parameters
+spec:
+  driver: vfio
+---
+apiVersion: resource.k8s.io/v1alpha2
+kind: ResourceClaimTemplate
+metadata:
+  name: rtx4090-claim-template
+spec:
+  spec:
+    resourceClassName: gpu.resource.nvidia.com
+    parametersRef:
+      apiGroup: rtx4090.gpu.resource.nvidia.com/v1
+      kind: ClaimParameters
+      name: rtx4090-claim-parameters
+---
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  labels:
+    kubevirt.io/vm: vm-cirros
+  name: vm-cirros
+spec:
+  running: false
+  template:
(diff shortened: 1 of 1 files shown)


rthallisey:
cc @varunrsekar

design-proposals/dra.md
```diff
@@ -0,0 +1,161 @@
...
+  running: false
+  template:
+    metadata:
+      labels:
+        kubevirt.io/vm: vm-cirros
+    spec:
+      resourceClaims:
+      - name: a100-40C
+        source:
+          resourceClaimTemplateName: a100-40C-claim-template
```
varunrsekar, line 135: Specifically for GPUs, this is the spec I had in mind to keep in-sync with the existing architecture and to address an usecase such as:
> As a user, I would like to use the dra driver for certain devices and device-plugins for others.

```
---
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  labels:
    kubevirt.io/vm: vm-cirros
  name: vm-cirros
spec:
  running: false
  template:
    metadata:
      labels:
        kubevirt.io/vm: vm-cirros
    spec:
       domain:
          devices:
             gpus:
             - name: gpu0 
               resourceClaim:
                  name: a100-40C
                  source:
                     resourceClaimTemplateName: a100-40C-claim-template
–--
apiVersion: v1
kind: Pod
metadata:
  name: virt-launcher-cirros
spec:
  containers:
  - name: virt-launcher
    image: virt-launcher
    resources:
      claims:
      - name: a100-40C
  resourceClaims:
  - name: a100-40C
    source:
      resourceClaimTemplateName: a100-40C-claim-template
```

The idea is that, for a `gpu` device, if the `resourceClaim` field is set, then kubevirt will know its to be provisioned by a dra driver and applies it to the virt-launcher pod spec. If instead, the `deviceName` is set, then kubevirt will know it is to be provisioned by a device-plugin and will set `resources.requests[<deviceName>]`.

The same rationale can be applied to the other device types from https://pkg.go.dev/kubevirt.io/api/core/v1#DomainSpec that would need DRA integration

alicefr:
@rthallisey one thing that is difficult for me to model with DRA is how we pass the device information. Today, there is an implicit API/mechanism between the device plugin and kubevirt how the device information are passed through environmental variables. For example, kubevirt cannot work out-of-the-box with any device plugins. 
Here, we kind have the some problem. Do you have any ideas how we could formalize more the mechanism how to pass the device information?

design-proposals/dra.md
```diff
@@ -0,0 +1,161 @@
...
+  running: false
+  template:
+    metadata:
+      labels:
+        kubevirt.io/vm: vm-cirros
+    spec:
+      resourceClaims:
+      - name: rtx4090
+        source:
+          resourceClaimTemplateName: rtx4090-claim-template
```
alicefr, line 73: Do we want to reference a claim here or the claim template? Can the user set the spec of the template?

fabiand: We have to have both.

rthallisey:
@alicefr 
>@rthallisey one thing that is difficult for me to model with DRA is how we pass the device information. Today, there is an implicit API/mechanism between the device plugin and kubevirt how the device information are passed through environmental variables. For example, kubevirt cannot work out-of-the-box with any device plugins.
Here, we kind have the some problem. Do you have any ideas how we could formalize more the mechanism how to pass the device information?
>
@varunrsekar is discussing this with upstream DRA folks.  We're working on a path to formalize device info in vendor plugins.

aburdenthehand:
@rthallisey @varunrsekar Are we still going ahead with this?

rthallisey:
@aburdenthehand yes

rthallisey:
Discussion has moved to https://github.com/kubevirt/community/pull/331

PR #294, opened 2024-05-22
network-binding-plugin: add plugin for vhostuser interfaces.

bgaussen:

This design proposal aims at implementing a new network binding plugin to support vhostuser interfaces. This will allow fast userspace datapath when used with a userspace dataplane like OVS-DPDK or VPP.
This design proposal takes into consideration sockets sharing issues between kubevirt and dataplane pods.

**Checklist**

- [X] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present.
- [X] PR: The PR description is expressive enough and will help future contributors
- [X] Community: Announcement to [kubevirt-dev](https://groups.google.com/g/kubevirt-dev/) was [considered](https://groups.google.com/g/kubevirt-dev/c/3w_WStrJfZw/m/yWSBpDAKAQAJ)

**Release note**:

```release-note
NONE
```


design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,141 @@
...
+This diagram explains this mechanism.
+
+![kubevirt-plugin-extension](kubevirt-plugin-extension.png)
+
+The advantages of such an approach are:
+- is generic and could also be potentially reused by other device types
+- hides the KubeVirt implementation details. Currently, you need to know where the KubeVirt sockets are located in the virt-launcher filesystem. Potentially, if we change the directory path for the sockets, this would break the CNI plugin
+- can isolate the resources dedicated to that particular plugin
+
+The drawback side is that `virt-launcher` pod would need to be run with `privileged: true` in order to do the bind mount.
```
alicefr, line 107: I don't think this is true. The directory can be exposed by virt-handler. Hence, no need of privileged for virt-launcher. This would also be unfeasible since virt-launcher is untrusted

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Device plugin for `vhostuser sockets` resource
+
+Device plugins have the ability to add hostPath mounts to the pods when they request for managed resources.
+We could implement a vhostuser device plugin that would manage two kinds of resources:
+- dataplane: 1
+  This only resource is resquested by the userspace dataplane, and add a `/var/run/vhost_sockets` mount to the dataplane pod.
```
alicefr, line 114: typo: `resquested` -> `requested`

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,141 @@
...
+- vhostuser sockets: n
+  This as many resources as we want to handle, is requested by the `virt-launcher` pod using vhostuser plugin. This makes the device plugin create a per pod directory like `/var/run/vhost_sockets/<launcher-id>`, and mount it into the `virt-launcher` pod.
```
alicefr, line 116: Have you tried this? I still worry that the directory created by the dp will still have the original issue with the mount propagation HostToContainer

alicefr, line 116: Is the pod ID know by the CNI plugin or outside KubeVirt?

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,141 @@
...
+
+This solution allows both dataplane and vm pods to share the vhostuser sockets without requiring to be privileged, and without the need for the CNI to do some bind mounts and avoid the constraints on Multus mountPropagation option.
+
+We still have to care about directory and sockets permission (and SELinux categories?). 
+
+## API Examples
+(tangible API examples used for discussion)
```
alicefr, line 123: Please. still add here the examples, it helps with the reviews and when this proposal is merged

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
bgaussen: I don't see very well the interaction between virt-handler, virt-launcher and the exposed directories. Need to check that.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
bgaussen: The only bind mount needed is the one the DP will push through kubelet. As there will be no further bind mount inside the pods, and as it will not be necessary for the CNI to do it neither, there should be no issue with mountPropagation.

bgaussen: DP can push annotation to the pod with an `id` it will generate. So it's not really the virt-launcher pod id, as it does not exist at the time DP is called.
The CNI can read the pod annotation and use it to configure the vhostuser with the right path in the data plane.

alicefr: Still I'm not sure how kubelet mounts the directory inside the pod. It might be possible that it is with HostToContainer and the socket won't be visible in the host directory. I would really like to check it with real code.

bgaussen: I understand your doubts ;) So I tested the scenario with [generic-device-plugin](https://github.com/squat/generic-device-plugin), and the following resources definition:

```yaml
        - --domain
        - dataplane.io
        - --device
        - |
          name: dataplane
          groups:
            - count: 1
              paths:
                - path: /var/lib/sockets
                  mountPath: /var/lib/sockets
                  type: Mount
                  permissions: mrw
        - --device
        - |
          name: sockets
          groups:
            - paths:
                - path: /var/lib/sockets/pod*
                  mountPath: /var/lib/socket
                  type: Mount
                  permissions: mrw
```

When a pod's container requests a `datplane.io/socket`, it will get assigned one of the existing `/var/lib/sockets/podXXX`, that gets bind mounted to `/var/lib/socket` in the container. Any file created there by the container is available in host `/var/lib/sockets/podXXX`.

If we have a dataplane pod requesting `dataplane.io/dataplane`, the whole `/var/lib/sockets` directory is bind mounted to `/var/lib/sockets`, and the dataplane container can see the files created in `/var/lib/sockets/podXXX`.
Reverse is also true if dataplane container creates a file in  `/var/lib/sockets/podXXX` it is available to the earlier pod's container.

I tested with only unprivileged pods.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,142 @@
...
+    <source type='unix' path='/var/run/kubevirt/sockets/poda08a0fcbdea' mode='server'/>
+    <target dev='poda08a0fcbdea'/>
+    <model type='virtio-non-transitional'/>
+    <mac address='ca:fe:ca:fe:42:42'/>
+    <driver name='vhost' queues='8' rx_queue_size='1024' tx_queue_size='1024'/>
+    <alias name='ua-net1'/>
+</interface>
+```
+
+This design leverages the existing `sockets` emptyDir mounted in `/var/run/kubevirt/sockets`. This allows the CNI to bind mount the socket emptyDir (`/var/lib/kubelet/<pod uid>/volumes/kubernetes.io~empty-dir/sockets`) to a host directory available to the dataplane pod through a hostPath mount.
```
EdDev, line 76: I'm a bit confused with the changes this proposals has passed through.
Can you please keep the alternatives and the reasons they have been kept out in an appendix?
E.g. https://github.com/kubevirt/community/pull/294#discussion_r1621991963 discussed the DP performing the mounting.

Regarding this intro sentence:
- How exactly can a CNI perform a mount to the pod/container? At the time the CNI is executed, there is not filesystem namespace defined yet. The CNI is also unlikely to have access to the api-server, it is unsafe and limited to trusted binaries only (e.g. Multus).
- The diagram describes in step 4 that `qemu` creates the socket.
  First, it will be nice to see that diagram steps described in details in text as well.
  In general, libvirt handles interfaces in an unmanaged manner, i.e. all it needs is provided to it (e.g. tap device). This is done to allow libvirt to operate in an unprivileged/rootless mode.
  - Is the socket creation possible with the rootless priv?
  - Can the socket be created by something else and then just consumed by libvirt?

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,142 @@
...
+
+Here is a diagram showing sockets sharing mecanisms between `virt-launcher` pod, `userspace CNI` and the `dataplane` pod.
+
+![kubevirt-vhostuser-shared-sockets](kubevirt-vhostuser-shared-sockets.png)
+
+Sharing the `vhostuser` sockets between `virt-launcher` pods and the dataplane pod is something to be enhanced in order to limit usage of hostPath volumes and bind mounts.
+
+## Alternative designs
+
+Some alternative designs were discussed in ![kubevirt-dev mailing-list](https://groups.google.com/g/kubevirt-dev/c/3w_WStrJfZw/m/yWSBpDAKAQAJ).
```
EdDev, line 91: Please summarize them in an appendix so we can see the full picture in one place.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Expose a virt-launcher pod directory to binding plugin
+
+This requires to implement a new network binding plugin mechanism we could expose the content of a `virt-launcher` directory to an external plugin.
+The plugin registration in Kubervirt resource would define the target directory on the node where the directory should be exposed.
+
+This diagram explains this mechanism.
+
+![kubevirt-plugin-extension](kubevirt-plugin-extension.png)
```
EdDev, line 100: I do not understand what is suggested here.
In virt-launcher we have a plugin container (the sidecar) and the compute container in which libvirt runs.
We have virt-handler in which a privileged operations may be conducted on the virt-launcher and on the api-server.

- Do you want a new parameter to be added to the network binding plugin spec to share data of the plugin container with the node, the virt-handler or the compute container in the virt-launcher?
- I do not understand how this is related to the CNI plugin. I guess this is related to my misunderstanding on how a CNI plugin can mount anything (I asked it in a different thread).

EdDev (commented):
I revisited this a bit late, sorry.

There are several points which I do not fully understand yet.
I would be interested in having a meeting to discuss this in detail, hopefully clarifying a few points and allowing this to be pushed forward.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,142 @@
...
+            binding:
+              name: vhostuser
+            macAddress: ca:fe:ca:fe:42:42
+```
+
+`network-vhostuser-binding` translates the VMI definition into libvirt domain XML modifications:
+1. Creates a new interface with `type='vhostuser'`
+2. Set the MAC address if specified in the VMI spec
+3. If `networkInterfaceMultiqueue` is set to `true`, add the number of queues calculated after the number of cores of the VMI
+4. Add `memAccess='shared'` to all NUMA cells elements
```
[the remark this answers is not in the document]
bgaussen: Hi Ed,

You mean the hook is called one time for each device using the same binding plugin? If so yes we need to add some words about that.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
bgaussen: > Regarding this intro sentence:
> 
> * How exactly can a CNI perform a mount to the pod/container? At the time the CNI is executed, there is not filesystem namespace defined yet. The CNI is also unlikely to have access to the api-server, it is unsafe and limited to trusted binaries only (e.g. Multus).

Indeed CNI needs to access api-server. Multus 3 allowed that through its own kubeconfig.multus. With Multus 4 thick mode, it's  [no longer possible,](https://github.com/k8snetworkplumbingwg/multus-cni/issues/1258#issue-2233364757) and CNI requesting api-server access needs to handle their own credentials creation.

> * The diagram describes in step 4 that `qemu` creates the socket.
>   First, it will be nice to see that diagram steps described in details in text as well.
>   In general, libvirt handles interfaces in an unmanaged manner, i.e. all it needs is provided to it (e.g. tap device). This is done to allow libvirt to operate in an unprivileged/rootless mode.
>   
>   * Is the socket creation possible with the rootless priv?

Yes it is, as far the filesystem permissions AND SElinux allows it.

>   * Can the socket be created by something else and then just consumed by libvirt?

It can. vhostusers sockets are created in server mode, and consumed in client mode. Usually qemu is in server mode, and the data plane the client. By the way the other way is getting [deprecated in ovs-dpdk](https://docs.openvswitch.org/en/latest/topics/dpdk/vhost-user/#vhost-user-vs-vhost-user-client) for example.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
bgaussen: > I do not understand what is suggested here. In virt-launcher we have a plugin container (the sidecar) and the compute container in which libvirt runs. We have virt-handler in which a privileged operations may be conducted on the virt-launcher and on the api-server.
> 

I guess both virt-launcher and virt-handler needs to be clarified when it comes to network binding plugins.

> * Do you want a new parameter to be added to the network binding plugin spec to share data of the plugin container with the node, the virt-handler or the compute container in the virt-launcher?

In that case it'd be a new network binding plugin spec parameter.

> * I do not understand how this is related to the CNI plugin. I guess this is related to my misunderstanding on how a CNI plugin can mount anything (I asked it in a different thread).

Indeed, the CNI needs to know where the socket is located to configure the data plane with this socket. In current userspace CNI implementation, the CNI does a bind mount of the socket into data plane filesystem namespace.

toelke:
I only found this discussion now, so I am very late in offering our experience:

We are productively using a self-build vhost-user management:

1. An admission webhook patches all pods with label `kubevirt.io=virt-launcher` to contain a `hostPath` `volume` (`/run/vpp`, in our case) and corresponding `volumeMounts`
2. A sidecar that patches the xml in a way very similar to this proposal. We add a new PCI subtree for the virtual devices, but I know that we are breaking kubevirt expectations there. Please note that having `<reconnect enabled="yes"/>` is useful here.

In our case, the driver creating the socket also creates a `config.json` in the same folder that is read by the sidecar to communicate information to the sidecar.

I will join the meeting on Wednesday.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,141 @@
...
+```xml
+<cpu mode="host-model">
+        <topology sockets="2" cores="8" threads="1"></topology>
+        <numa>
+            <cell id="0" cpus="0-7" memory="2097152" unit="KiB" memAccess="shared"/>
+            <cell id="1" cpus="8-15" memory="2097152" unit="KiB" memAccess="shared"/>
+        </numa>
+</cpu>
+<interface type='vhostuser'>
+    <source type='unix' path='/var/run/kubevirt/sockets/poda08a0fcbdea' mode='server'/>
```
ormergi, line 67: I wonder, does the socket is created by the CNI inside virt-launcher pod?

If so, I think the socket path can be reflected on Multus network-status annotation under the subject element device-info, see https://github.com/k8snetworkplumbingwg/device-info-spec/blob/main/SPEC.md#315-vhost-user

The vhost-user device-info can be exposed using the network binding-plugin API's [downardAPI](https://github.com/kubevirt/community/blob/main/design-proposals/network-binding-plugin/network-binding-plugin.md#device-info).
The sidecar could then read the vhost-user socket from the mounted downwardAPI volume and spesify it in the domain xml.

Please see [vDPA example](https://github.com/kubevirt/community/blob/main/design-proposals/network-binding-plugin/network-binding-plugin.md#appendix-f-concrete-example-for-vdpa-binding) for reference how downardAPI can be used.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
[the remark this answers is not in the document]
ormergi: The `OnDefineDomain` hook can be called multiple times by Kubeivrt  regardless of devices count.
Thus the sidecar implementation for `OnDefineDomain` hook should be idempotent.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,141 @@
...
+The `vhostuser` secondary interfaces configuration in the dataplane is under the responsibility of Multus and the CNI such as `userspace CNI`.
+
+## Definition of Users
+Users of the feature are everyone that deploys a VM.
+
+## User Stories
+- As a user, I want to create a VM with one or serveral `vhostuser` interfaces attached to a userspace dataplane.
+- As a user, I want the `vhostuser` interface to be configured with a specific MAC address.
+- As a user, I want to enable multi-queue on the `vhostuser` interface
+- As a Network Binding Plugin developper, I want the shared socket path to be accessible to virt-launcher pod
```
ormergi, line 21: typo 
```suggestion
- As a Network Binding Plugin developer, I want the shared socket path to be accessible to virt-launcher pod 
```

ormergi (commented):
Thanks for the proposal

Please see my inline comments

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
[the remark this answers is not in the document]
EdDev: > You mean the hook is called one time for each device using the same binding plugin? If so yes we need to add some words about that.

Just assume that it can be called many times, you should not assume how many times.
Today, it is called twice in a normal flow, for the purpose of calculating some default that libvirt provides.
In that past it was called many more times (but we got it optimized since).

In a flow with failures, reconciliation may cause it to be called several times.

In the future, other hook points in the code flow may trigger it.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
EdDev: > the CNI does a bind mount of the socket into data plane filesystem namespace.

This part I do not understand.
AFAIU there is no socket yet, as that is created by libvirt when the domain is created.
Perhaps we can clarify this by listing all the steps that are actually done.

bgaussen: Sorry I meant "the CNI does a bind mount of directory where the socket will be created"...
You can refer to the diagram included in the DP.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
bgaussen: Thank you for the hints. Indeed the new network binding plugin downardAPI will be helpful!

bgaussen:
Hi all,

I just committed modifications to this design proposal PR to focus on the device plugin based solution to facilitate socket sharing, as discussed during last week meeting.

Regards,

Benoit.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,186 @@
...
+Device plugins have the ability to add mounts into the pods when they request for managed resources.
+
+This design proposal relies on a device plugin that would manage two kinds of resources, representing a host directory where sockets will rely (for example `/var/run/vhost_sockets`):
+- **dataplane**: `1`  
+  This only resource is requested by the userspace dataplane, and add a `/var/run/vhost_sockets` mount to the dataplane pod.
+- **vhostuser sockets**: `n`  
+  This as many resources as we want to handle and may represent the number of port of the dataplane vSwitch.  
+  It is requested through VM or VMI definition in resources request spec. In turn the `virt-launcher` pod will request the same resources.  
+  This makes the device plugin create a per pod directory like `/var/run/vhost_sockets/<pod-dp-id>`, and mount it into the `virt-launcher` pod as to well known location `/var/run/vhostuser`.  
+  The device plugin has to generate a `pod-dp-id` and push it as an annotation in `virt-launcher` pod. This will be used later by the CNI or any component that configures the vhostuser socket in the dataplane with right path.
```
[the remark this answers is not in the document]
bgaussen: You're right and we are trying to not rely on API server for the CNI to get annotations.
For example this [Mutlus PR](https://github.com/k8snetworkplumbingwg/multus-cni/pull/1204) could ease the task.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,186 @@
...
+- As a VM User, I want to be able to configure the `vhostuser` interface as transitional
+- As a Cluster Admin, I want to be able to enable `network-vhostuser-binding`
+- As a Network Binding Plugin Developer, I want the shared socket path to be accessible to `virt-launcher` pod 
+- As a Dataplane Developer, I want to access all `vhostuser` sockets of VM pods
+- As a CNI Developer, I want to know whet vhostuser sockets are located
+ 
+## Repos
+Kubevirt repo, and most specificaly [cmd/sidecars](https://github.com/kubevirt/kubevirt/tree/main/cmd/sidecars).
+
+## Design
```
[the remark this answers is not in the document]
bgaussen: The plugin is designed to support secondary networks.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Functional Testing Approach
+Create a VM with several `vhostuser` interfaces then:
+- check the generated domain XML contains all interfaces with appropriate configuration
+- check the vhostuser sockets are created in the expected directory of virt-launcher pod
+- check the vhostuser sockets are available to the dataplane pod
+- check the VM is running
+
+# Implementation Phases
+1. First implementation of the `network-vhostuser-binding` done
```
yuvalk, line 184: can you link to where that is?

bgaussen:
Hi all,

A little update: we're currently implementing the Device Plugin for sockets sharing. We're implementing the device-info spec to share device information between Device Plugin and CNI, and we'll rely on DownwardAPI support in Kubevirt 1.3 to get network-status annotation in the network binding plugin (as suggested by @ormergi, thx 😁).

I'll update the Design Proposal ASAP.

bgaussen:
Hi All,

We made great progress and implemented Device Plugin (DP), CNI and Network Binding Plugin (NBP):

- DP handles socketDirs where vhostuser sockets will be created, and injects the mount into the compute container requesting the socket device.
- DP is compliant with [device-info-spec ](https://github.com/k8snetworkplumbingwg/device-info-spec/blob/main/SPEC.md), as pointed by @ormergi, and creates device plugin devinfo files
- CNI implements also device-info-spec. That allows to get allocated device spec, including the socketDir path. It enables the CNI to push the spec, through multus, to `virt-launcher` pod annotations (that is available at the end in `kubevirt.io/network-info` annotation)
-  NBP retrieves the socketDir path from the annotation and DownwardAPI and creates the vhostuser interface in DomainXML with the right path to the socket

Example of this interface:

```xml
    <interface type='vhostuser'>
      <mac address='02:15:7e:00:00:05'/>
      <source type='unix' path='/var/lib/sockets/socketDir04/pod6c270ef2f25' mode='server'/>
      <target dev='pod6c270ef2f25'/>
      <model type='virtio-non-transitional'/>
      <driver name='vhost' queues='4' rx_queue_size='1024' tx_queue_size='1024'/>
      <alias name='ua-net1'/>
      <address type='pci' domain='0x0000' bus='0x02' slot='0x00' function='0x0'/>
    </interface>
```

where /var/lib/sockets/socketDir04 is a random directory managed by the DP.

So far so good... but...

When we live migrate the VM, a new destination pod is created, with a new socket request to DP. The socketDir will obviously not be the same as the one on source pod (ex: /var/lib/sockets/socketDir09)

But from what I understand the source domainXML is sent to the destination pod and qemu tries to start with the source socketDir, and fails:

```
compute {"component":"virt-launcher","kind":"","level":"error","msg":"Live migration failed.","name":"ubuntu-vm1","namespace":"tests","pos":"live-migration-source.go:845","reason":"error encountered during Migrate
ToURI3 libvirt api call: virError(Code=1, Domain=10, Message='internal error: process exited while connecting to monitor: 2024-09-06T13:12:18.279191Z qemu-kvm: -chardev socket,id=charua-net1,path=/var/lib/sockets/
socketDir04/pod6c270ef2f25,server=on: Failed to bind socket to /var/lib/sockets/socketDir04/pod6c270ef2f25: No such file or directory')","timestamp":"2024-09-06T13:12:18.334616Z","uid":"33c51f90-1843-48ed-851f-4cb
11d75af41"}
```

Do you have any idea on how to manage this situation? Is there any plan to implements something through NBP livemigrate methods that could alter the destination domainXML?

As a "temporary" workaround, we intend to use a mutating webhook to inject a shared empty-dir volume in `virt-launcher` pod spec, that will be shared between NBP and compute containers (mounted in `/var/lib/vhost-sockets`). NBP will use the info from the `kubevirt.io/network-info` annotation/downwardAPI to create a link in the shared volume (`/var/lib/vhost-sockets/net1 -> /var/lib/sockets/socketDir09`).
The vhostuser socket path in domainXML will be `/var/lib/vhostsockets/net1/pod6c270ef2f25`, insuring that it will be accessible from the compute container and be the same on source and destination pods in live migration case.

Do you think the NBP spec could be amended to add a shared volume between compute and NBD sidecars?

Thanks,

Benoit.

ormergi:
Hi @bgausse, well done!

It looks like Kubevirt calls sidecar hooks also at the migration target https://github.com/kubevirt/kubevirt/blob/328316d0083077cb151142c0250c0c6251479859/pkg/virt-launcher/virtwrap/live-migration-target.go#L110-L117

Could you please ensure the NBP sidecar executes at the target?

ormergi:
Hi @bgaussen , well done!

It looks like Kubevirt calls sidecar hooks also at the migration target https://github.com/kubevirt/kubevirt/blob/328316d0083077cb151142c0250c0c6251479859/pkg/virt-launcher/virtwrap/live-migration-target.go#L110-L117

Could you please ensure the NBP sidecar executes at the target?

bgaussen:
> Hi @bgaussen , well done!
> 
> It looks like Kubevirt calls sidecar hooks also at the migration target [here](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-launcher/virtwrap/live-migration-target.go#L175-L182)
> 
> Could you please ensure the NBP sidecar executes at the target?

Yes the sidecar hooks get executed, and from its logs it sent back the modified domainXML with the right socketDir on destination pod.

But from what I understand [here](https://github.com/kubevirt/kubevirt/blob/89ce45e4568cc75163a65223b9b4710eda876d2f/pkg/virt-launcher/virtwrap/live-migration-source.go#L924), the source domain is used for MigrateToURI3 libvirt call, with little modification in `PrepareDomainForMigration`.

bgaussen:
Hi All,

After some private discussion with @alicefr, it seems that the migration domain takes precedence over the one generated at pod creation. From what I understand, it could be difficult to fix this behavior but I will open an issue with that.

In the meantime, for live migration to work, we need the sockets paths be the same on the source and target pods. Hence the symlinks mechanism seems a good way to achieve that.
But we need the Network Binding Plugin to access the `compute` container mount namespace, with two options:

1. A new shared empty-dir volume between `compute` and NBP sidecar containers.
It requires to be added to NBP spec. Could be interesting if other NBP requires such a feature...

2. Enable the `shareProcessNamespace` pod feature on `virt-launcher`. The sidecar container can target on of the `compute` container process, and access its mount namespace through `/proc/PID/root`, where the NBP can create the symlink.
There is already this PR we can benefit of: kubevirt/kubevirt#11845

We already implemented and tested `2.`, it works as expected. The only drawback is that we need to target `virtlogd` process. For other ones (`virt-launcher-monitor`, `virt-launcher`, `qemu*`), `/proc/PID/root` is not accessible.

We can discuss this in a community meeting and/or mailing list.

Benoit.

aburdenthehand:
@bgaussen Hi Benoit, just to let you know your most recent commit was not signed. I know you're still working on this so whenever you next squash or rebase please ensure that all commits are signed off to avoid any issues further down the line. Thanks!

EdDev:
> In the meantime, for live migration to work, we need the sockets paths be the same on the source and target pods. Hence the symlinks mechanism seems a good way to achieve that.

I think that `shareProcessNamespace` is problematic. It flattens the process NS from the whole pod and in practice drops an expected isolation of containers in the pod.
It feels to me like a hack with major implications which are not easy to asses.

I think there is another option which could be explored: Using libvirt hooks.
Libvirt does have an option to mutate the domain configuration on the target through script files, which could answer the need here. In the past, at least at other projects (e.g. oVirt) this direction encountered opposition but I do not recall the exact reasons.

Libvirt hooks are documented here:
https://wiki.archlinux.org/title/Libvirt#Hooks
https://libvirt.org/hooks.html

bgaussen:
> I think that `shareProcessNamespace` is problematic. It flattens the process NS from the whole pod and in practice drops an expected isolation of containers in the pod. It feels to me like a hack with major implications which are not easy to asses.

Understand this drawback with `shareProcessNamespace`, but as discussed with @alicefr, if it is to be implemented in the PR #11845, we could be benefit of it.
If not we need another way.

> I think there is another option which could be explored: Using libvirt hooks. Libvirt does have an option to mutate the domain configuration on the target through script files, which could answer the need here. In the past, at least at other projects (e.g. oVirt) this direction encountered opposition but I do not recall the exact reasons.
> 
> Libvirt hooks are documented here: https://wiki.archlinux.org/title/Libvirt#Hooks https://libvirt.org/hooks.html

I briefly look at libvirt qemu hooks, indeed we could mutate the domain when it's received, before it's started. But we need to figure out how to add the scripts in compute container...

Another way would be to add a dedicated shared dir between compute container and hook sidecar binding plugin where we can create the links...

Benoit.

EdDev:
> Another way would be to add a dedicated shared dir between compute container and hook sidecar binding plugin where we can create the links...

That you for helping me understand this well in the unconference meeting.
This one seems to me a good candidate as well, especially when it is so simple.
First, it can be easily implemented with a mutation webhook. I do agree it is better we provide it as a formal integration option between the compute and other sidecars.

Was this option challenged by some risk?
@alicefr , having a well defined are where sidecars and the compute can share data seems a nice additions to the integration "protocol". What is your opinion on this option? (sorry if I missed a prev conversation about this)

alicefr:
> Was this option challenged by some risk? @alicefr , having a well defined are where sidecars and the compute can share data seems a nice additions to the integration "protocol". What is your opinion on this option? (sorry if I missed a prev conversation about this)

There is already [an option with a PVC](https://github.com/kubevirt/kubevirt/pull/10598) and the sidecars , we could modify it to allow an empty dir as well.

bgaussen:
> > Was this option challenged by some risk? @alicefr , having a well defined are where sidecars and the compute can share data seems a nice additions to the integration "protocol". What is your opinion on this option? (sorry if I missed a prev conversation about this)
> 
> There is already [an option with a PVC](https://github.com/kubevirt/kubevirt/pull/10598) and the sidecars , we could modify it to allow an empty dir as well.

@alicefr, @EdDev,

Following the unconference meeting and the discussion we had, we implemented the shared emtpy-dir logic with a mutating webhook.

The binding plugin creates the link (immutable path and socket name), in this empty-dir, to the real socket, and uses this link in the domainXML vhostuser interface.

This solves the issue observed with Live Migration.

This seems indeed to be the best solution, as there is already the basis to enable a new mount for this empty-dir for sidecar containers.

mw-0:
How long is drop using Live Migration?

bgaussen:
> How long is drop using Live Migration?

In our tests, we loose 2 ping packets during live migration, so around 2 seconds drop.

mw-0:
Hi,  Just wondered what was needed to push this through?

aburdenthehand:
@bgaussen Is this waiting for reviews of updates or is still in progress?

bgaussen:
> @bgaussen Is this waiting for reviews of updates or is still in progress?

@aburdenthehand, I updated de design proposal, with latest discussion we had. 
Waiting for reviews from @alicefr, @EdDev, @fabiand.

mw-0:
did this die?

TReZ42:
> did this die?

Hope it is not ! Waiting from community feedback, and how to proceed to go further.

bgaussen:
> did this die?

Hope it is not ! Waiting from community feedback, and how to proceed to go further.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,224 @@
...
+
+Network Binding Plugin then can leverage `downwardAPI` feature available from Kubevirt v1.3.0, in order to retrieve the `kubevirt.io/network-info` annotation values, and extract the socket path to configure the interface in the domain XML.
+
+But it can't use it directly as it would break Live Migration of VMs:   
+The socket directories `/var/run/vhostuser/<socketXX>` are not predictable, and new ones get allocated when the destination pod is being created.  
+Unfortunately the domain XML is the one from the source pod (migration domain), and references sockets paths allocated to source pod.
+
+Hence, Network Binding Plugin needs to use immutable paths to sockets. This can be achieved using the interface name (or its hash version) in symbolic links to the real socket path: `/var/run/kubevirt/vhostuser/net1` -> `/var/run/vhostuser/<socketXX>`.
+
+This requires an enhancement in KubeVirt, and Network Binding Plugin KubeVirt CRD spec, in order for `virt-launcher` pod to have a shared `emptyDir` volume, mounted in both `compute` and `vhostuser-network-binding-plugin` containers.
```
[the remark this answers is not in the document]
EdDev: Investigation: There is already a shared path for the grpc socket between the sidecar and compute container, an idea was raised to check if it can be re-used for this purpose as well.

EdDev (commented):
Sorry for the later feedback.
I'll make sure to stay available for this and target this proposal for v1.6 timeline (especially if we'll need Kubevirt core changes).

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,224 @@
...
+
+## User Stories
+- As a VM User, I want to create a VM with one or serveral `vhostuser` interfaces attached to a userspace dataplane.
+- As a VM User, I want the `vhostuser` interface to be configured with a specific MAC address.
+- As a VM User, I want to enable multi-queue on the `vhostuser` interface
+- As a VM User, I want to be able to configure the `vhostuser` interface as transitional
+- As a Cluster Admin, I want to be able to enable `network-vhostuser-binding`
+- As a Network Binding Plugin Developer, I want the shared socket path to be accessible to `virt-launcher` pod 
+- As a Dataplane Developer, I want to access all `vhostuser` sockets of VM pods
+- As a CNI Developer, I want to know whet vhostuser sockets are located
```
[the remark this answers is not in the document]
bgaussen: Thanks @EdDev for your feedback. Fixed this one ;)

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,224 @@
...
+  This makes the device plugin allocates a sub directory `/var/run/vhostuser/<socketXX>`, and mount it into the `virt-launcher` pod.
+
+The device plugin has to comply with [`device-info-spec`](https://github.com/k8snetworkplumbingwg/device-info-spec/blob/main/SPEC.md#device-information-specification). This allows information sharing between device plugin and the CNI. Thanks to Multus being compliant with this spec, the CNI can retrieve device information (socket path and and type) to be used to configure the dataplane accordingly. Multus will  annotate the `virt-launcher` pod with this information, KubeVirt extracts only a part into `kubevirt.io/network-info`.
+   
+The device plugin has to care about directory permissions and SELinux, for the sockets to be accessible from requesting pods.
+
+#### Network Binding Plugin and Kubevirt requirements
+
+Network Binding Plugin then can leverage `downwardAPI` feature available from Kubevirt v1.3.0, in order to retrieve the `kubevirt.io/network-info` annotation values, and extract the socket path to configure the interface in the domain XML.
```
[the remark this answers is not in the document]
bgaussen: The sidecar container is getting `kubevirt.io/network-info` thanks to `downwardAPI`.

The `kubevirt.io/network-info` annotation is created after `k8s.v1.cni.cncf.io/network-status`, that is added by Multus and the CNI.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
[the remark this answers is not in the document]
bgaussen: > I think that a working alternative is to deploy a mutating webhook for this. I am in favor of this `sharedDir`, in its customized from (choosing the path, as presented here) or stricter one which the path is hard-coded. The later seems safer and consistent, but I do not have a strong opinion here.

Indeed, we tested and run this thanks to a mutating webhook. Regarding the customization of the path, you're right it could be a hardcoded path per plugin. But I think we need to be sure that 2 plugins do not use a same path, otherwise overwrite could happen.



> Investigation: There is already a shared path for the grpc socket between the sidecar and compute container, an idea was raised to check if it can be re-used for this purpose as well.

We initially used this one, but same remark as above.

sbrivio-rh:
By the way,

> This design proposal aims at implementing a new network binding plugin to support vhostuser interfaces. This will allow fast userspace datapath when used with a userspace dataplane like OVS-DPDK or VPP.

Now that [passt](https://passt.top) supports vhost-user operation, the integration in KubeVirt is in progress ([one step here](https://github.com/kubevirt/kubevirt/pull/14101)), and that will be another form of fast user-mode data path. 

Perhaps this pull request (I didn't thoroughly go through all the commits, though) should mention this peculiarity ("hardware-attached"? "hardware direct"?) to avoid confusion.

> How long is drop using Live Migration?

Also by the way, what we do in passt now is to save, migrate, and restore TCP queues using the `TCP_REPAIR` socket option. I guess one could do something similar with DPDK, eventually (migrating pending queue slots). I'm mentioning this because, as a step to enable passt migration, vhost-user-based migration of virtio-net back-ends was implemented in QEMU (https://patchew.org/QEMU/20250115135044.799698-1-lvivier@redhat.com/). A direct back-end could probably do something similar.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+spec:
+  configuration:
+    network:
+      binding:
+        vhostuser:
+          sidecarImage: network-vhostuser-binding:main
+          sharedDir: /var/run/kubevirt/vhostuser
```
EdDev, line 144: Getting back to this.
We want to target this for v1.6 timeline.

Can you please remove this and give a work or two that the intent is to set it as a well-known fixed name.

bgaussen: Thanks @EdDev for the roadmap update.

Do you already have an idea of this well-known fixed name? I assume it will be generic to accommodate all binding plugin?

EdDev (approved):
Thank you, it looks great!

From the kubevirt core project (kubevirt/kubevirt) we would like to add the shared volume to enable communication between the compute and the sidecar. All other parts are to be implemented in other repos.

design-proposals/network-vhostuser-binding-plugin/network-vhostuser-binding-plugin.md
bgaussen: This would be appropriate! That means one emptyDir per plugin or sidecar?

EdDev: Plugin that requires sidecar == sidecar
Multi ref to a single plugin is still one sidecar.

oshoval:
Hi @bgaussen
Thanks for the feature and design.

Can you please check the approach we did about the requested shared volume ?https://github.com/kubevirt/kubevirt/pull/14623

The PR refactors the current shared mount, in a way that each container will have a dedicated SubPath,
and then in the new sidecar volumeMount you will be able to create a folder and your socket in it.
Thanks

bgaussen:
Hi @oshoval ,

Thank you for the VEP and its implementation!

> Can you please check the approach we did about the requested shared volume ?[kubevirt/kubevirt#14623](https://github.com/kubevirt/kubevirt/pull/14623)
> 
> The PR refactors the current shared mount, in a way that each container will have a dedicated SubPath, and then in the new sidecar volumeMount you will be able to create a folder and your socket in it. Thanks

This looks good to me. Thank you for having made it generic for any binding plugin.
Adapting our vhostuser binding implementation, getting the CONTAINER_NAME env and creating the sub directory in the hooks socket shared dir should be quite straightforward.

bgaussen:
Thank you @oshoval for your great work.

We'll implement the new env CONTAINER_NAME + dir ASAP in our binding plugin and keep you updated.

maxpain:
Is it possible to use `vhost-user` to connect a VM to VPP at the current time? Or is it still WIP?

oshoval:
Hi @bgaussen
Thanks
Please tag me on the issue you open,
If possible please enclose there the PR you work on and manifests / commands (such as kubectl patching of kubevirt CR)
so i will try it locally

Thank you

bgaussen:
/remove-lifecycle stale

The last remaining issue was solved by @oshoval in KubeVirt 1.6.3. Hence the vhostuser network binding plugin is fully working, even for Live Migration.

I don't know if it is relevant to merge this design proposal.

Thank all for the work!

mw-0:
@bgaussen do you have a release for you plugin? i have some bugs on mine

bgaussen:
> @bgaussen do you have a release for you plugin? i have some bugs on mine

I expect the plugin to be released soon now🤞... What kind of bugs to you have?

mw-0:
> > @bgaussen do you have a release for you plugin? i have some bugs on mine
> 
> I expect the plugin to be released soon now🤞... What kind of bugs to you have?

libvirt in kubevirt errors it expects ovs-vsctl in the path to configure which is not there.

PR #295, opened 2024-05-28
Drop feature gate in descheduler support design proposal

fossedihelm:
The descheduler support design proposal specifies that a feature gate will be added.
The feature gate can be dropped, since no work is added other than to add annotations.
The feature gate drop also has the advantage of saving work to the cluster admins; with the feature gate, if a cluster admin wants to enable the descheduler, they have to install it and add the feature gate to KubeVirt.
Without the feature gate they only have to install it.

Also, this feature gate only enables a support for the descheduler, which should always be enabled, and there are no advantages to not have it.

Fixes #

**Checklist**

**Release note**:

```release-note
NONE
```
/cc @acardace @Barakmor1 @iholder101 @jean-edouard @xpivarc

### DIFF
--- a/design-proposals/descheduler-support.md
+++ b/design-proposals/descheduler-support.md
@@ -30,21 +30,19 @@ The descheduler should be able to operate on `virt-launcher` pods.
 ## User Stories
 - As a cluster admin running a K8s cluster with a descheduler, I want to deploy and use KubeVirt.
 - As a cluster admin running a K8s cluster with KubeVirt, I want to deploy a descheduler that will correctly operate on my VMs.
-- As a cluster admin I want to enable cluster-wide KubeVirt's descheduler support.
 
 ## Repos
 1. kubevirt/kubevirt
-2. kubevirt/hyperconverged-cluster-operator
 
 # Design
 ## Descheduler Requirements
 1. In order for the descheduler to be able to differentiate `virt-launcher` pods from other pods, all `virt-launcher` pods should be annotated with `descheduler.alpha.kubernetes.io/request-evict-only`.
 2. In cases where the VM will be migrated, eviction requests on `virt-launcher` pods should return 429 HTTP code with a special reason message.
-3. In order for the deschduler to understand that a `virt-launcher` pod is pending migration, it should be annotated with `descheduler.alpha.kubernetes.io/eviction-in-progress` so it will not try to evict it again.
+3. In order for the descheduler to understand that a `virt-launcher` pod is pending migration, it should be annotated with `descheduler.alpha.kubernetes.io/eviction-in-progress` so it will not try to evict it again.
 4. In case the migration failed, the `descheduler.alpha.kubernetes.io/eviction-in-progress` annotation will be removed from the source `virt-launcer` so the descheduler will understand that it could try evicting it again.
 
 ## Feature Gate
-A `DeschedulerSupport` feature gate will be introduced in order to enable this feature.
+A feature gate is not needed.
 
 ## Changes to Pod Eviction Webhook
 Currently, the descheduler observes the one of the following responses:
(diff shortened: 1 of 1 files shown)


iholder101:
From my perspective, the most important thing here is that the descheduler is not a Kubevirt feature, hence it shouldn't be guarded with a feature gate.

If we want to provide a way of avoiding these extra annotations, we can always provide a **configuration** for doing so, but using a feature gate for it is wrong IMO.

However, I currently don't see a reason to provide such a configuration as a few possibly redundant annotations shouldn't cause any harm that I can see.

/lgtm

acardace:
/lgtm

+1 Totally agree with this.

fabiand:
Agree with @iholder101 - I tend to say that we want a cluster level way to opt out of setting the annotations.
Main reason: Some people might not want these annotations.

fabiand:
Actually @fossedihelm when are we adding the annotations? unconditionally or under certain conditions?

iholder101:
> Agree with @iholder101 - I tend to say that we want a cluster level way to opt out of setting the annotations. Main reason: Some people might not want these annotations.

Thanks @fabiand!

Just to clarify: If we want this to be configurable, then a cluster-wide configuration is the way to go rather than a feature gate. Saying that, I'm not sure we need to add such a configuration.

We're talking about pod-level annotations, which are considered as an "implementation detail" of our VM/VMI objects.

Do you think there's a reason for users to opt-out of these annotations?

fossedihelm:
> Actually @fossedihelm when are we adding the annotations? unconditionally or under certain conditions?

It depends on the annotation:
- `descheduler.alpha.kubernetes.io/request-evict-only` will be added for every virt-launcher pod
- `descheduler.alpha.kubernetes.io/eviction-in-progress` to the source virt-launcher pod only when an evacuation migration is triggered (not every migration).

IOW, from a practical point of view, we are going to add only the first one because the second one will be added to the pod that (eventually) will be `Completed`.

To me, a single annotation to the virt-launcher pod (not VM nor vmi) should not disturb users.

fabiand:
Ok, thanks. Reasonable.

So, is `descheduler.alpha.kubernetes.io/evict` https://github.com/kubernetes-sigs/descheduler/blob/17af29afe418d56aff1954ee4ee0392e29b3fd5e/README.md?plain=1#L947
still required today as well?

design-proposals/descheduler-support.md
```diff
@@ -30,21 +30,19 @@ The descheduler should be able to operate on `virt-launcher` pods.
...
 2. In cases where the VM will be migrated, eviction requests on `virt-launcher` pods should return 429 HTTP code with a special reason message.
-3. In order for the deschduler to understand that a `virt-launcher` pod is pending migration, it should be annotated with `descheduler.alpha.kubernetes.io/eviction-in-progress` so it will not try to evict it again.
+3. In order for the descheduler to understand that a `virt-launcher` pod is pending migration, it should be annotated with `descheduler.alpha.kubernetes.io/eviction-in-progress` so it will not try to evict it again.
 4. In case the migration failed, the `descheduler.alpha.kubernetes.io/eviction-in-progress` annotation will be removed from the source `virt-launcer` so the descheduler will understand that it could try evicting it again.
 
 ## Feature Gate
-A `DeschedulerSupport` feature gate will be introduced in order to enable this feature.
+No feature gate will be present, as no work will be done in KubeVirt. KubeVirt will only add annotations which, in the worst case, will have no added value.
```
orelmisan, line 45:
```suggestion
A feature gate is not needed.
```

orelmisan (commented):
Thank you @fossedihelm.
I support adding cluster-wide configuration to select whether the annotations will be added.
But it could be added last.

design-proposals/descheduler-support.md
fossedihelm: Done, thanks!

orelmisan (approved):

fabiand:
/approve

according to all previous positive reviews.

PR #303, opened 2024-06-16
Network binding plugin: Support compute container resource overhead

orelmisan:
Some network binding plugins require compute resource overhead, for example the passt plugin requires additional memory overhead in the virt-launcher pod's compute container.
Suggest several alternatives to address this issue.

Fixes #

**Checklist**

**Release note**:

```release-note
NONE
```

### DIFF
--- a/design-proposals/network-binding-plugin/network-binding-plugin.md
+++ b/design-proposals/network-binding-plugin/network-binding-plugin.md
@@ -224,6 +224,56 @@ pod.
     - Mount points
     - Downward API
 
+##### Compute Container Resource Overhead
+
+For some plugins, an additional resource consumption can be expected from the virt-launcher pod compute container.
+For example, there could be need to execute an additional binary in the compute container.
+Since this binary has its own CPU and memory limits, they should be somehow accounted for.
+Another example could be increased resource consumption of the virt-stack resulting from using the plugin.
+
+Suggested solution:
+
+Additional API for compute container resource overhead:
+
+The network binding plugin API in the KubeVirt CR could receive an additional input field to specify the resource requirements overhead for the compute container:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+metadata:
+  name: kubevirt
+  namespace: kubevirt
+spec:
+  configuration:
+    network:
+      binding:
+        mynetbindingplugin:
+          sidecarImage: quay.io/kubevirt/mynetbindingplugin
+          computeResourceOverhead:
+            requests:
+              memory: 200Mi
+```
+
+If specified, KubeVirt will add the resource overhead specified in `computeResourceOverhead` to the compute container of the virt-launcher pod.
+The specified resource overhead will be added to the compute container per unique usage of the plugin (not per NIC using the binding).
+
+For example, assuming there is a plugin registered with a 200Mi memory overhead for the `compute` container, and
+there are two interfaces using it.
+`virt-controller` will only add 200Mi of memory to the `compute` container.
+
+Pros:
+- Cluster-wide definition of compute resource overhead per plugin.
+- Additional resources could be requested other than CPU and memory.
+- The resource overhead specification is visible to cluster admins.
+
+Cons:
+- Requires an API change.
+- When upgrading KubeVirt / network binding plugin versions, the compute resource overhea specification might require adjustments.
+
+This solution was selected since it provides the cluster admin more control in regard to resource allocation.
+
+For the alternative solutions please see [Appendix G](#appendix-g-alternatives-to-compute-container-resource-overhead-specification)
+
 #### Configure Pod netns
 
 The CNI plugin has privileged access to the pod network namespace and
@@ -1163,3 +1213,67 @@ Libvirt configuration snippet:
 ```
 
 - The domain itself is created by the `virt-launcher`.
+
+# Appendix G:  Alternatives to compute container resource overhead specification
+
+1. Manually setting the VM's resources:
+
+Users could manually provide the additional memory overhead for the network binding plugin, under `spec.domain.resources.requests.memory`:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: vm-cirros
+spec:
+  template:
+    spec:
+      domain:
+        memory:
+          guest: 128Mi
+        resources:
+          requests:
+            memory: 640Mi # 128Mi for the guest + 512Mi for the network binding plugin
+```
+
+KubeVirt will create a virt-launcher pod's compute container with a memory request containing the following:
+- Guest's memory
+- Memory overhead for KubeVirt's components (calculated by virt-controller)
+- Memory overhead for the network binding plugin
+
+Pros:
+- Already implemented.
+
+Cons:
+- Error prune
+- Defined per VM and not cluster-wide.
+- Exposes the VM owner to unnecessary implementation details
+
+2. Mutating webhook
+
+For each unique network binding plugin used, the VMI controller will add a label on the virt-launcher pod with the following format:
+
+`kubevirt.io/network-binding-plugin:<plugin-name>`
+
+The binding plugin authors will provide a [mutating webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) that will intercept
+virt-launcher pod creation that have the above label, and add the appropriate resources requests/limits
+for the pod's `compute` container.
+
+The mutating webhook will be able to identify the plugin's compute container by its name (`compute`) or using the value of
+the `kubectl.kubernetes.io/default-container` annotation.
+
+Pros:
+- Plugin authors have full control over the compute resources
+- Additional API is not added to KubeVirt.
+- Opens the door for additional changes to the virt-launcher pod without changes to KubeVirt.
+- Code changes in KubeVirt are very small.
+
+Cons:
+- Plugin authors should provide another component and integrate it.
+- Additional point of failure.
+- Requires maintaining certificates for the webhook.
+- Additional latency when creating VMs with network binding plugins.
+- The additional resource specification is less visible to cluster admins.
+- Resource specification could collide with the support container resources specified on the KubeVirt CR or other webhooks.
+
+The requirement to maintain certificates for the webhook could be mitigated using tools such as [cert-manager](https://cert-manager.io/).


ormergi (commented):
Thanks for the PR, overall looks good, see my inline comments.

Regarding the the PR description and second commit message, I think it should also mention that memory overhead is necessary to avoid pod eviction due to passt VMs consume more memory then expected, and improve passt VMs scheduling results; passt VMs wont be scheduled on nodes that doesnt have enough memory.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1261,6 +1271,10 @@ metadata:
     k8s.v1.cni.cncf.io/networks: '[{"name":"netbindingpasst","namespace":"mynamespace","cni-args":{"logicNetworkName":"default"}}]'
 spec:
   containers:
+    - name: compute
+      resources:
+        requests:
```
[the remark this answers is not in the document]
orelmisan: AFAIK, there is logic [1] to automatically add memory limits when needed[1] (there is also an equivalent for CPU).

[1] https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-controller/services/renderresources.go#L189

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1162,152 @@ Libvirt configuration snippet:
...
+The API between the sidecar and `virt-launcher` is using gRPC with a
+well-defined interface.
+
+The following steps are expected:
+
+- When the domain configuration is created, it passes the VMI
+  specification and the domain specification from the `virt-launcher` to
+  the sidecar.
+- The sidecar computes the required domain configuration, specifically
+  for the bridge binding, to work and returns the mutated spec to the
```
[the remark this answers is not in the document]
orelmisan: This is a copy-and-past error.
Thanks.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1162,152 @@ Libvirt configuration snippet:
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: virt-launcher-123
+  annotations:
+    k8s.v1.cni.cncf.io/networks: '[{"name":"netbindingpasst","namespace":"mynamespace","cni-args":{"logicNetworkName":"default"}}]'
+spec:
+  containers:
+    - name: hook-sidecar-0  
+      image: quay.io/kubevirt/network-passt-binding:<tag>
```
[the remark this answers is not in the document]
orelmisan: Done.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1162,152 @@ Libvirt configuration snippet:
...
+## Appendix G: Concrete example for passt-binding
+
+Present a concrete example of the actions needed at each integration
+point, considering the option suggested in the
+[design section](#design-details).
+
+### Kubevirt CR Specification
+
+The passt-binding is registered to the Kubevirt CR, including the
+reference to the sidecar image.
```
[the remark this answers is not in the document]
orelmisan: Done.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1162,152 @@ Libvirt configuration snippet:
...
+  namespace: kubevirt
+spec:
+  configuration:
+    network:
+      binding:
+        passt:
+          networkAttachmentDefinition: default/netbindingpasst
+          sidecarImage: quay.io/kubevirt/network-passt-binding:<tag>
+          migration:
+            method: link-refresh
```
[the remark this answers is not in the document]
orelmisan: I removed it.
Lets address it in a follow-up since its not in the scope of the PR.

EdDev (commented):
First commit review.

Although I placed comments inline on that commit, I think it is not a must example addition.

At the design level, we are not really interested in a specific binding, but on the general concept.
The general concept can be applied on other bindings, just to provide an example on how to use it.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -189,6 +189,7 @@ the binding is referenced by name with optional additional arguments.
         - domainAttachment (a standard domain definition that exists in
           the core . E.g `tap`, `sriov`).
         - downwardAPI (E.g `device-info`)
+        - resourceOverhead (currently only additional memory overhead request for the compute container in the virt-launcher pod is supported)
```
EdDev, line 192: I would be specific here and mention the compute-container in the name.
E.g. `computeResources`.

This also opens up the different questions and options which should be discussed in a proposal:
- Should we consider only the compute container? What about the sidecar container?
- Ref the compute container, we can use a specific name for the field (e.g. `computeResources`) or we can make it configurable using a flag under a general `resources`.
- What is the reason for not just doing a `computeMemoryRequest` needs to convince the reviewers.

Now, we can make it even more general by moving away from the specific network usage of this and look at a general sidecar hook that KubeVirt supports.
E.g.: The Kubvirt CR will have a policy for allocating resources to sidecar containers and possible the compute container. Referencing this policy from the network binding definition or from a sidecar hook. This may be an overkill, but thinking about it and examining the pors/cons of it can be useful.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -387,6 +388,11 @@ Alternatives passing the domain attachment to sync VMI:
 Both of the options are not perfect, but the `virt-handler.SyncVMI`
 has fewer cons. Therefore, it was chosen.
 
+#### Additional resource requests for the virt-launcher compute container
+Some binding plugins may require an additional binary to be shipped inside the virt-launcher compute image.
+This binary requires additional resources that have to be explicitly stated.
```
EdDev, line 393: I do not think we need to limit it to an additional binary. It could be just more resources from the cgroupv2 of the compute container, consumed by an existing binary (e.g. libvirt).

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1189,6 +1196,8 @@ spec:
         passt:
           networkAttachmentDefinition: default/netbindingpasst
           sidecarImage: quay.io/kubevirt/network-passt-binding
+          resourceOverhead:
+            memory: 800Mi
```
EdDev, line 1200: I think you will have to explain why not:
- Directly `memoryOverhead: 500Mi` (BTW, it should be 500 and not 800).
- With explicit `request` and being open for adding `limit`.

EdDev (commented):
Second commit:

The commit message hints and explains the passt binding and everything is driven from it. While this is correct in terms of why we do all this, I do not think we should start the story there.

The story should start from the need, mentioning passt as an example.
The fact that there is a binary alongside libvirt is not that important. If it was part of libvirt we would have needed to take that into account as well.

Also, it would be better to leave the details to the design and in the commit just provide the topic/subject. That way we can easily review it and adjusts it per that review.

design-proposals/network-binding-plugin/network-binding-plugin.md
fabiand: runtimeClasses have a pretty similar concept, maybe worth aligning the API idea
https://kubernetes.io/docs/concepts/scheduling-eviction/pod-overhead/

ie.

    overhead:
      podFixed:
        memory: "120Mi"
        cpu: "250m"

EdDev: > runtimeClasses have a pretty similar concept, maybe worth aligning the API idea

Interesting, thanks for sharing this.
I think the API could be in sync with the general [resources](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#example-1) concept.
This allows to control the resource additions to the container, with the ability to extend it in the future to any type of resource and define both request and limits.

The oddity here, as I see it, is how we express the resources of a different container (i.e. compute) and still not lock the ability to do the same for the sidecar itself.

orelmisan:
[Change](https://github.com/kubevirt/community/compare/8432599c1ae2854c59a1bc280fc94964e9fced94..416926c734c448dd49b0a2c0f1f4beee5649921e): Removed the passt example, as it is not essential for to the proposal.

Addressed sidecar and compute container resource specification.

orelmisan:
@EdDev @ormergi PTAL.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,86 @@ pod.
...
+> **Note**
+>
+> Setting CPU/memory requests and/or limits for sidecar pods on the KubeVirt CR will apply uniformly on all [hook sidecar containers](https://kubevirt.io/user-guide/user_workloads/hook-sidecar/) and network binding plugin sidecar containers.
+
+In the common scenario of a regular VM and no spacial configuration on the KubeVirt CR, 
+the network binding plugin sidecar containers do not have CPU/Memory requests and limits.
+The sidecar container can have a memory leak and may cause node's destabilization.
+
+
+Alternatives:
```
EdDev, line 242: Please keep here the decided solution and place the alternatives (with a ref from here) to a dedicated appendix.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,86 @@ pod.
...
+Pros:
+- Already implemented.
+
+Cons:
+ - Applies uniformly on all hook sidecars and network binding plugins 
+   - Worst case CPU/Memory requests/limits should be defined on the KubeVirt CR which maybe wasteful for a combination of hook sidecars and network binding plugins.
+   - Coarse level of control.
+   - Only supports CPU and memory requests and limits.
+
+2. Additional API for sidecar resource configuration:
```
EdDev, line 255: This is specific for the binding plugin, not the sidecar. The sentence is not clear about this.
The definition is per the network binding plugin, and applied if the plugin uses a sidecar.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,86 @@ pod.
...
+
+Pros:
+ - Cluster-wide definition of network binding plugin sidecar resources per plugin.
+ - Finer grained control over resources allocated to network binding plugin sidecars.
+ - Decoupling from existing hook sidecars.
+ - Additional resources could be requested other than CPU and Memory.
+
+Cons:
+ - Require an API change.
+ - The API will probably evolve as additional plugins will be created.
```
EdDev, line 284: I do not understand this one.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,86 @@ pod.
...
+ - May require cluster admins to adjust plugin resources during the plugin's lifecycle.
```
EdDev, line 285: The cluster admin is responsible to register the binding plugin in the first place, so I am unclear what this point means.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,86 @@ pod.
...
+- Already implemented.
+
+Cons:
+ - Applies uniformly on all hook sidecars and network binding plugins 
+   - Worst case CPU/Memory requests/limits should be defined on the KubeVirt CR which maybe wasteful for a combination of hook sidecars and network binding plugins.
+   - Coarse level of control.
+   - Only supports CPU and memory requests and limits.
+
+2. Additional API for sidecar resource configuration:
+The network binding plugin API in the KubeVirt CR could receive an additional input field to specify the sidecar resource requirements:
```
EdDev, line 256: The specified resource is define for each instance of usage or per the sidecar, irrelevant of the usage?
E.g. there may be 1 interface using the plugin or there may be 3 interfaces using the plugins in the same VM.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,86 @@ pod.
...
+
+Cons:
+ - Require an API change.
+ - The API will probably evolve as additional plugins will be created.
+ - May require cluster admins to adjust plugin resources during the plugin's lifecycle.
+
+3. Mutating webhook for the virt-launcher pod:
+For each network binding plugin used, the VMI controller will add a label on the virt-launcher pod with the following format:
+
+`kubevirt.io/network-binding-plugin:<plugin-name>`
```
EdDev, line 290: How will the admission webhook be able to identify the relevant container?

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,86 @@ pod.
...
+The binding plugin authors will have to provide a mutating webhook that will intercept
+virt-launcher pod creation with the above label and add the appropriate resources requests/limits
+for every relevant sidecar container.
+
+Pros:
+- Plugin authors have full control over the sidecar's resources.
+- KubeVirt does not add additional API.
+- Very small change in KubeVirt.
+
+Cons:
```
EdDev, line 301: What about the need to manage certificates?

EdDev (commented):
First commit review.

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev, line 301: Another con is its collision with the existing methods or possible other webhooks that someone may use.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -303,6 +303,58 @@ Cons:
 - Additional point of failure.
 - Additional latency when creating VMs with network binding plugins.
 
+##### Compute Container Resource Requirements / Limits
+
+For some plugins, such as passt, there is a need to execute an additional binary in the compute container.
+Since this binary has its own CPU and memory limits, they should be somehow accounted for.
```
EdDev, line 309: Please do not limit it to running an additional binary, the addition in memory or other resources may come from different reasons (e.g. libvirt itself requiring more memory due to the expected configuration).

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -303,6 +303,58 @@ Cons:
...
+
+Alternatives:
+1. Manually setting the VM's resources:
+The user can override KubeVirt's algorithms and set resource requirements.
```
EdDev, line 313: The current logic that adds overhead per internal core logic is colliding with adding manually an overhead? Specifically, if one specifies explicitly resource requests, are these being increased by Kubevirt logic or something else is happening?

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -303,6 +303,58 @@ Cons:
...
+
+Pros:
+- Already implemented.
+
+Cons:
+- Error prune
+- The user does not take into account the overhead considerations KubeVirt takes when templating a virt-launcher pod.
```
EdDev, line 320: I guess this is related to my prev question. Can you please confirm this is indeed the case?
I would be surprise if this is how it works.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -303,6 +303,58 @@ Cons:
...
+```
+
+Pros:
+- Cluster-wide definition of network binding plugin compute resource overhead per plugin.
+- Additional resources could be requested other than CPU and Memory.
+
+Cons:
+- Require an API change.
+- The API will probably evolve as additional plugins will be created.
+- May require cluster admins to adjust plugin resources during the plugin's lifecycle.
```
EdDev, line 351: Like with the sidecar previously, these points are not clear.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -303,6 +303,58 @@ Cons:
...
+Pros:
+- Already implemented.
+
+Cons:
+- Error prune
+- The user does not take into account the overhead considerations KubeVirt takes when templating a virt-launcher pod.
+
+2. Additional API for compute container resource overhead:
+
+The network binding plugin API in the KubeVirt CR could receive an additional input field to specify the resource requirements overhead for the compute container:
```
EdDev, line 324: There is a need to explicitly specify if this overhead is added per plugin type or per plugin usage count (i.e. per the number of interfaces referencing it from the VM).

It is also important to specify if it is dependent on any other field/setup.
The previous sidecar resource are dependent on the existence of a sidecar, but this one may not have such a dependency.

EdDev (commented):
Thank you for the proposal.

The following general points should be considered and added:
- Effort cost estimation for each option.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Added it as a con + a mitigation method.

orelmisan: Done.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1162,152 @@ Libvirt configuration snippet:
...
+
+```yaml
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: netbindingpasst
+  namespace: default
+spec:
+  config: '{
+      "cniVersion": "0.3.1",
```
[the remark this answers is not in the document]
orelmisan: This was removed.
Please ack so we can resolve the thread.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1162,152 @@ Libvirt configuration snippet:
...
+
+- As a result, the pod is updated with the following annotations & spec:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: virt-launcher-123
+  annotations:
+    k8s.v1.cni.cncf.io/networks: '[{"name":"netbindingpasst","namespace":"mynamespace","cni-args":{"logicNetworkName":"default"}}]'
```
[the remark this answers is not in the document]
orelmisan: This was removed.
Please ack so we can resolve the thread.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1162,152 @@ Libvirt configuration snippet:
...
+spec:
+  containers:
+    - name: hook-sidecar-0  
+      image: quay.io/kubevirt/network-passt-binding:<tag>
+```
+
+### Configure Pod network namespace
+
+Not required for passt binding
```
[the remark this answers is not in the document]
orelmisan: This was removed.
Please ack so we can resolve the thread.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1162,152 @@ Libvirt configuration snippet:
...
+
+### Reacting to CNI results (Multus status annotation)
+
+Not required for passt binding
+
+### Run services in the virt-launcher pod
+
+The passt binary is shipped and executed inside the virt-launcher pod.
```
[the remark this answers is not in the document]
orelmisan: This was removed.
Please ack so we can resolve the thread.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1162,152 @@ Libvirt configuration snippet:
...
+
+```xml
+<interface type='user'>
+    <mac address='12:34:56:78:9a:bc'/>
+    <source dev='eth0'/>
+    <portForward proto='tcp'/>
+    <portForward proto='udp'/>
+    <model type='virtio-non-transitional'/>
+    <backend type='passt' logFile='/var/run/kubevirt/passt.log'/>
+    <alias name='ua-default'/>
```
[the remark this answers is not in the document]
orelmisan: This was removed.
Please ack so we can resolve the thread.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Done.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Done.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Done.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Discussed this point offline.
It was removed.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Discussed this point offline, improved the wording.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Addressed it in the text.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Addressed it in the text.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Done.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Discussed this point offline, adjusted the text.

orelmisan (commented):
Addressed most of the review comments made by @EdDev.
@EdDev PTAL.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1208,53 @@ Libvirt configuration snippet:
...
+        mynetbindingplugin:
+          sidecarImage: quay.io/kubevirt/mynetbindingplugin
+          sidecarResources:
+            requests:
+              cpu: 200m
+              memory: 20Mi
+```
+
+KubeVirt will create a sidecar container spec with the resource specification provided in `sidecarResources` in case a sidecar image was specified for the plugin.
+Since KubeVirt creates a sidecar container per unique plugin defined on the VMI spec, the plugin sidecar's resources will only be added once.
```
EdDev, line 1265: Please specify who will add the resources (e.g. which controller) and when. I should be clear then how the container on which the resources are added is known.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1208,53 @@ Libvirt configuration snippet:
...
+Pros:
+- Already implemented.
+
+Cons:
+- Applies uniformly on all hook sidecars and network binding plugins
+    - Worst case CPU/Memory requests/limits should be defined on the KubeVirt CR which maybe wasteful for a combination of hook sidecars and network binding plugins.
+    - Coarse level of control.
+    - Only supports CPU and memory requests and limits.
+
+2. Additional API for network binding plugin sidecar resource specification:
```
EdDev, line 1242: I am interested to understand the estimated cost of implementing this option compared to the admission mutation option.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Done.

orelmisan:
[Change](https://github.com/kubevirt/community/compare/e0b43e5c4c9ca7a86bfe6dfa3923803e5a8b9349..ba8f921e148355d47c92f46a7dece83baaec159a): Answered review [comment](https://github.com/kubevirt/community/pull/303#discussion_r1654405497).

design-proposals/network-binding-plugin/network-binding-plugin.md
ormergi, line 1242: The network binding plugin **sidecar** container resources could quickly become common challenge for plugin authors, delegating it to Kubevirt could make it easier for plugin authors.

Note that at the next commit, that address the compute container resource overhead, one of the suggestions extend the API in similar way, the compute container is hardcoded in the field name `computeContainerResources`.
I suggest avoid hardcoded container name in the filed name to imply for which container the spesifed resources required for,  for example:
```
binding:
        mynetbindingplugin:
          sidecarImage: quay.io/kubevirt/mynetbindingplugin
          containerResources:
            name: <compute / network binding plugin sidecar>                        
            resources:
              requests:
                cpu: 200m
                memory: 20Mi
               limits: 
                 ...
```
Here we see the new `containerResources` field, it has `name` and `resources`  fields.
1. `container` stands for which container the resource requirements should apply for.
     There should be subset of acceptable names "compute" and "sidecar".
     When `compute` is specified, the desired resources should be added to compute container, as overhead to the existing resources.
     When `sidecar` is spesifed, the desired resources should be defined in the plugin sidecar container spec.
2. `resources` represent the desired resources.

Since the binding API is alpha, another options is changing the API structure in a way it enable more options for  the sidecar container sped, for example:
```
binding:
        mynetbindingplugin:
          sidecar:
            image: quay.io/kubevirt/mynetbindingplugin
            resources:
              requests:
                cpu: 200m
                memory: 20Mi
```
Here we see the new `sidecar` field, it has `image` and `resources` fields:
1. `image` stands for the sidecar container image, it should replace the existing `sidecarImage` field.
2. `resources` represent the desired resources.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1208,53 @@ Libvirt configuration snippet:
...
+  name: kubevirt
+  namespace: kubevirt
+spec:
+  configuration:
+    network:
+      binding:
+        mynetbindingplugin:
+          sidecarImage: quay.io/kubevirt/mynetbindingplugin
+          sidecarResources:
+            requests:
```
ormergi, line 1259: Maybe it better to use the native resources API as in Pod [1] 

[1] https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#example-1

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1258,3 +1274,50 @@ Pros:
...
+metadata:
+  name: kubevirt
+  namespace: kubevirt
+spec:
+  configuration:
+    network:
+      binding:
+        mynetbindingplugin:
+          sidecarImage: quay.io/kubevirt/mynetbindingplugin
+          computeContainerResourceOverhead:
```
ormergi, line 1307: Same here, I would try to avoid hard coding the container name in the field name.

ormergi (commented):
Thanks for the proposal, please see my inline comments

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,52 @@ pod.
     - Mount points
     - Downward API
 
+##### Sidecar Resource Requirements / Limits
```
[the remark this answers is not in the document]
EdDev: This has been discussed offline and the decision was made to post it together as they are close related.
So I will repeat here my response: If there will be a a sign that there is no agreement, it will be splited.

It is also worth noting that the two topics are updating/extending an existing proposal. They are not introduced as a separate proposal.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,52 @@ pod.
...
+the network binding plugin sidecar containers do not have CPU/Memory requests and limits.
+The sidecar container can have a memory leak and may cause node's destabilization.
+
+Suggested solution:
+
+For each unique network binding plugin used, the VMI controller will add a label on the virt-launcher pod with the following format:
+
+`kubevirt.io/network-binding-plugin:<plugin-name>`
+
+The binding plugin authors will provide a [mutating webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) that will intercept
```
EdDev, line 248: After some thought, I think that in this particular case I would go with the Kubevirt API addition and not a webhook.
The reasoning is related to the need for the Cluster Admin to have control over the resources it allows to run.

With a mutation webhook, it is up to the plugin author to provide the expected resources details.
But IMO it is the Admin who should be able to control the resources. It should not depend on the deployment of the webhook or its correctness.

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev: The second option makes sense now, but I do not think it is worth breaking the API for it.
While it is Alpha, we know that it has already users and our interest is not to break them.

I prefer we keep the option to break backward compatibility only for stronger reasons.

Regarding the 1st option, I think it is problematic: It will allow one plugin to affect other containers in general. That is not a good API IMO.
I think it is better to limit the plugin control over its own container and in the compute one, that is it.

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev: IIUC this is exactly what was suggested here.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -270,6 +270,22 @@ The requirement to maintain certificates for the webhook could be mitigated usin
...
+##### Compute Container Resource Overhead
+
+For some plugins, an additional resource consumption can be expected from the virt-launcher pod compute container.
+For example, there could be need to execute an additional binary in the compute container.
+Since this binary has its own CPU and memory limits, they should be somehow accounted for.
+Another example could be increased resource consumption of the virt-stack resulting from using the plugin.
+
+Suggested solution:
+
+Shipping a mutating webhook as part of the plugin's deliverables.
```
EdDev, line 282: Same as for the sidecar resources (prev commit) reasoning, I would prefer the Kubevirt CR API extension option.

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev: Same answer as before.

orelmisan:
[Change](https://github.com/kubevirt/community/compare/ba8f921e148355d47c92f46a7dece83baaec159a..35ddd8b975399745f52bdbb2c095c1ad9047b34c): Moved the API alternative to be the proposed solution.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: This is indeed what is proposed here.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: > Regarding the 1st option, I think it is problematic: It will allow one plugin to affect other containers in general. That is not a good API IMO. I think it is better to limit the plugin control over its own container and in the compute one, that is it.

IIUC, what @ormergi meant is that this field would be limited to "compute" or "sidecar" options.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Done.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Done.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: Done.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1269,67 @@ Libvirt configuration snippet:
...
+- Additional point of failure.
+- Requires maintaining certificates for the webhook.
+- Additional latency when creating VMs with network binding plugins.
+- The additional resource specification is less visible to cluster admins.
+- Resource specification could collide with the support container resources specified on the KubeVirt CR or other webhooks.
+
+This solution provides flexibility for plugin authors while keeping the network binding plugin API in KubeVirt small.
+The requirement to maintain certificates for the webhook could be mitigated using tools such as [cert-manager](https://cert-manager.io/).
+
+# Appendix H:  Alternatives to compute container resource overhead specification
```
EdDev, line 1316: After a discussion with @enp0s3 , there is another more generic solution alternative:

Add a new resource overhead section under the KV config:
```yaml
          computeOverhead:
              selector: ....
              resources:
                  requests:
                      cpu: 200m
                      memory: 20Mi
```

The selector will determine to which pod it will apply to. The binding plugin, will create pods with a label that identifies the pod as using the relevant plugin. The code that adds the overhead, will query it and determine on which pod to add the overhead.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,113 @@ pod.
...
+  namespace: kubevirt
+spec:
+  configuration:
+    network:
+      binding:
+        mynetbindingplugin:
+          sidecarImage: quay.io/kubevirt/mynetbindingplugin
+          computeResourceOverhead:
+              cpu: 200m
+              memory: 20Mi
```
EdDev, line 311: It seems this is missing the `requests` (and `limits`) level, why?

orelmisan: Re-added the `requests` field.

orelmisan:
[Change](https://github.com/kubevirt/community/compare/e4827000b5a5ba860accab95dbb7bbda742f7711..848082d314785d6baa78ced89c67158071fd5a3c): Updated the compute container's manual resource specification alternative.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: The way it works is as follows:
A VM could be defined with both guest memory and memory resource specification:
```yaml
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  name: vm-cirros
spec:
  template:
    spec:
      domain:
        memory:
          guest: 128Mi
        resources:
          requests:
            memory: 640Mi # 128Mi for the guest + 512Mi for the network binding plugin
```

The virt-launcher pod's compute container will have a memory request which contains the sum of:
- Guest VM's memory (128Mi in this example).
- Memory overhead for KubeVirt's components (its size is dependent on the VMI's spec)
- Arbitrary memory overhead (512Mi in this example)

The domain XML will contains the the guest's memory specification (128Mi in this example).

As a side note, it is also possible to specify a memory request with less memory than the guest requires.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: This is untrue.
I removed this line.

design-proposals/network-binding-plugin/network-binding-plugin.md
enp0s3: @EdDev Hi. I think we would need to open a separate design proposal for the generic solution.

EdDev: Sure, but we still want to record it here as it is related.
A remark can be added to express that it it will be presented as an independent proposal.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,63 @@ pod.
...
+metadata:
+  name: kubevirt
+  namespace: kubevirt
+spec:
+  configuration:
+    network:
+      binding:
+        mynetbindingplugin:
+          sidecarImage: quay.io/kubevirt/mynetbindingplugin
+          sidecarResources:
```
ormergi, line 260: Please add a note saying this field type should be the native kubernetes resources API, with validation for cpu and memory.

design-proposals/network-binding-plugin/network-binding-plugin.md
ormergi: Yep, the container names should be validated in a way they can refer to the binding sidecar and compute container only.

design-proposals/network-binding-plugin/network-binding-plugin.md
ormergi: Would the spesifed resources apply for all containers in the selected pod?
Resources are configured on container types, it seem that it"ll require finer way to set overhead on containers, did you mean selector for containers?

FYI: I suggested something similar in this [comment](https://github.com/kubevirt/community/pull/303#discussion_r1654833989).
Note that the container name should be validated and allow refering to the binding  sidecar and compute containers only.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -281,6 +281,57 @@ This solution was selected since it provides the cluster admin more control in r
...
+          computeResourceOverhead:
+            requests:
+              cpu: 200m
+              memory: 20Mi
+```
+
+If specified, KubeVirt will add the resource overhead specified in `computeResourceOverhead` to the compute container of the virt-launcher pod.
+The specified resource overhead will be added to the compute container per unique usage of the plugin (not per NIC using the binding).
+
+For example, assuming there is a plugin registered with a 200Mi memory overhead for the `compute` container, and
```
ormergi, line 318: Does passt require mempry overhead per interface or based on whether its being executed or not?

design-proposals/network-binding-plugin/network-binding-plugin.md
ormergi: This is a mere yaml example it can be anything :slightly_smiling_face:

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -1162,4 +1270,89 @@ Libvirt configuration snippet:
...
+
+Pros:
+- Already implemented.
+
+Cons:
+- Error prune
+- Defined per VM and not cluster-wide.
+- Exposes the VM owner to unnecessary implementation details
+
+2. Mutating webhook
```
ormergi, line 1352: Is it another webook in addition to one that address the sidecar resources one?

It seem that it lacks description about what it"ll do, pros, cons section and how its going to integrate with kubevirt? ( I recall something about adding some label to launcher pods that use network binding plugins)

ormergi (commented):
Thanks for the changes

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev: That is still a less optimal API IMO. A good API hints the user the boundaries of what is possible and what is not. An enum-like option as suggested here fits cases where we suspect the options will extend/change. But if we know they are limited, especially to only 2, it does not make sense to go this path IMO.

If we had 5 options, then for the same of simplicity, it would make more sense. On the other hand, if we had 5 options, it is most likely it would potentially grow and then will really fit an enum.

We do not expect this to grow as we want to limit this to the `compute` and  self `binding-plugin` container with no option to extend it.

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev: I agree that we should be explicit about the supported API.
We should not specify the `cpu` but we can add a not that the structure is the native k8s one, allowing extending the support to the other fields (`cpu`, custom ones).

The definition of the API should not be limited to an example.
But this is not blocker for me.

Adding the validation section can also be useful. We could specify there that we ignore any data except the `memory` at this stage. We should be explicit and say that we do not fail if other data is added. Again, this is not a blocker and I am good with letting this stay gray for now.

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev: It is enough for the interface to be defined for the passt binary to run and consume the memory specified.

It is not clear what you mean by `executed`.

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -281,6 +281,57 @@ This solution was selected since it provides the cluster admin more control in r
...
+              cpu: 200m
+              memory: 20Mi
+```
+
+If specified, KubeVirt will add the resource overhead specified in `computeResourceOverhead` to the compute container of the virt-launcher pod.
+The specified resource overhead will be added to the compute container per unique usage of the plugin (not per NIC using the binding).
+
+For example, assuming there is a plugin registered with a 200Mi memory overhead for the `compute` container, and
+there are two interfaces using it.
+`virt-controller` will only add 200Mi of memory to the `compute` container.
```
EdDev, line 320: Where is this `200Mi` coming from? In the example above you used `20Mi`.

design-proposals/network-binding-plugin/network-binding-plugin.md
EdDev: > Would the spesifed resources apply for all containers in the selected pod?
> Resources are configured on container types, it seem that it"ll require finer way to set overhead on containers, did you mean selector for containers?

I'm sure author of the proposal will handle this detail.

> FYI: I suggested something similar in this comment.
> Note that the container name should be validated and allow refering to the binding sidecar and compute containers only.

The alternative here is generic, not specific to a binding plugin and requires other steps for it to work.
I think we can get into the details of what it needs to validate or not as part of a different proposal that will focus on this alternative. It should be enough to mention here the general idea, so it will be visible that we discussed it.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: It was a typo.
Fixed it in the example.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: I've added additional details, same as for the sidecar.

orelmisan:
[Change](https://github.com/kubevirt/community/compare/d4eb851bf0fa09a4a1768bcd132b1f3ad3d890d5..c33498b476eb9b2db6be962765c8ed3d6817efce): Split the sidecar's resource specification to #309

EdDev (approved):
Thank you!

design-proposals/network-binding-plugin/network-binding-plugin.md
vladikr: @EdDev I'm late to review this, but I don't understand why this API needs to add anything to the compute container and not simply advertise the resources needed to run this binding container?!
Same way as it's done for hotplug / other containers?
https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-controller/services/renderresources.go#L611

vladikr: > runtimeClasses have a pretty similar concept, maybe worth aligning the API idea https://kubernetes.io/docs/concepts/scheduling-eviction/pod-overhead/
> 
> ie.
> 
> ```
> overhead:
>   podFixed:
>     memory: "120Mi"
>     cpu: "250m"
> ```

@fabiand I doubt it aligns well. This overhead API was invented for kata and the overhead is being added to any pod created with this runtimeclass, I don't think it aligns.

vladikr: > @EdDev I'm late to review this, but I don't understand why this API needs to add anything to the compute container and not simply advertise the resources needed to run this binding container?! Same way as it's done for hotplug / other containers? https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-controller/services/renderresources.go#L611

@EdDev do we expect all network binding to affect the compute container or is it just passt?
passt specifically uses a feature in qemu there for the `GetMemoryOverhead` functional can identify that passt will be used and add a passt specific overhead - this is instead of a kubevirt CR API.
What do you think?

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,56 @@ pod.
     - Mount points
     - Downward API
 
+##### Compute Container Resource Overhead
+
+For some plugins, an additional resource consumption can be expected from the virt-launcher pod compute container.
+For example, there could be need to execute an additional binary in the compute container.
```
vladikr, line 230: Could you please explain what binary can be executed and how it gets there? 
How can an external plugin rely on the fact that a specific binary is part of the virt-launcher image?

design-proposals/network-binding-plugin/network-binding-plugin.md
```diff
@@ -224,6 +224,56 @@ pod.
...
+Since this binary has its own CPU and memory limits, they should be somehow accounted for.
```
vladikr, line 231: I think, we should speak of a specific functionality and not about arbitrary binaries.
virt-launcher should be well defined. 
If we know what functionality we're enabling that requires additional overhead, kubevirt needs to know the resources it requires. Similar to other aspects accounted for in `GetMemoryOverhead`.

I would say that if we want an API for this it should be based on a functionality, not a binding. 

 Each resource consumption of it.

design-proposals/network-binding-plugin/network-binding-plugin.md
orelmisan: This example was derived from the `passt` use case, on which the `passt` binary is shipped as part of the virt-launcher `compute` image.

Another example could be that the plugin configures the domain in a way that causes the virt-stack to consume additional compute resources that KubeVirt cannot account for.

orelmisan:
Thank you for reviewing this proposal @vladikr, sorry it took some time to respond.

design-proposals/network-binding-plugin/network-binding-plugin.md
fabiand: @vladikr my comment was only about the API design, not about using the POD api.

aburdenthehand:
On behalf of SIG-compute, can one of @jean-edouard @stu-gott @enp0s3 @vladikr please re-review to move this forward. I can merge if I receive an approve or second lgtm from one of you. Thank you.

design-proposals/network-binding-plugin/network-binding-plugin.md
enp0s3: @vladikr Hi
I see that according to the network binding plugin design document, one of the [integration points](https://github.com/kubevirt/community/blob/main/design-proposals/network-binding-plugin/network-binding-plugin.md#integration-points) that seems legit is the [pod definition phase](https://github.com/kubevirt/community/blob/main/design-proposals/network-binding-plugin/network-binding-plugin.md#pod-definition--creation)
There were no exceptions regarding resource rendering, and Kubevirt CR was advised as a potential API to extend in order to integrate into that point.

Given that the network binding design was accepted, and also the implementation of this design also got into the codebase, I am not sure how to proceed, I'm in favor of accepting this design. 

Can you please advice @vladikr ?

enp0s3:
Deferring the discussion for after the end year holidays

orelmisan:
> > Hi @EdDev, @bgartzi, is [VEP 144](https://github.com/kubevirt/enhancements/issues/144) supposed to replace the field introduced by this PR?
> 
> Yes, partially. But I think it is enough. This one allows both "request" and "limit" to be defined. I think the other one is just about "request".

There are cases where both requests and limits are important (when the VM should have guaranteed QOS).

bgartzi:
> > >  Hi @EdDev, @bgartzi, is https://github.com/kubevirt/enhancements/issues/144 supposed to replace the field introduced by this PR?
> >
> > Yes, partially. But I think it is enough. This one allows both "request" and "limit" to be defined. I think the other one is just about "request".
> 
> There are cases where both requests and limits are important (when the VM should have guaranteed QOS).

VEP 144 tries to cover both based on my understanding: On the one hand, a mechanism to update virt-launcher memory lock limits; users can basically say y/n to that. On the other hand, a mechanism that acts pretty much as `netbinding.CalculateMemoryOverhead` as the API value impacts `GetMemoryOverhead`'s output. The difference is that it is not applied per-NPB, but per-VMI.

orelmisan:
Closing this one before the automation will.
For future reviewers: this field exists on the KubeVirt API as Alpha.
Future enhancements / deprecation will take place on the enhancements repo.

PR #305, opened 2024-07-02
design-proposal: Generating disk serial numbers

akrejcir:
This design introduces automatic generation of serial numbers for disks.

**Issue**:
https://issues.redhat.com/browse/CNV-39082

**Checklist**

**Release note**:
```release-note
Added proposal for disk serial number generation.
```

### DIFF
--- a/design-proposals/generated-disk-serial.md
+++ b/design-proposals/generated-disk-serial.md
@@ -1,0 +1,202 @@
+# Overview
+This proposal introduces a new feature in KubeVirt to automatically
+generate serial numbers for each disk attached to a virtual machine.
+
+## Motivation
+- Windows guest OS in the VM requires that disk serial numbers are defined and persistent across reboots.
+- Generating the serial reduces manual configuration.
+
+## Goals
+- Add an option to the VM's `.spec` that will enable automatic serial number generation for disks that
+  don't have a serial number specified.
+
+- Generate unique serial numbers for each disk. The number has to be unique compared to all other disks attached
+  to all VMs with this disk.
+ 
+- Don't store the serial number in the VM's `.spec`. It would confuse gitops operators.
+
+- The serial number should remain the same across storage migration.
+
+## Non Goals
+
+- No customization options for serial number format.
+
+## Definition of Users
+
+- VM owners
+
+## User Stories
+
+- As a VM owner, I want to use the automated disk serial number assignment to reduce manual configuration.
+- As a VM owner, I want each disk to have a unique serial number.
+- As a VM owner, I want newly created Windows VMs to have serial numbers set on disks by default.
+
+## Repos
+- [kubevirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+
+A new boolean option `generateDiskSerials` will be added to the VMI `.spec`. 
+When it is set to `true`, kubevirt will generate a serial number for each disk, that does not have one specified.
+These serial numbers will then be set in the libvirt domain xml.
+
+This option can also be added to `VirtualMachinePreference`, so it can be set on new VMs and
+the user does not need to be aware of it.
+
+## Persistence of generated serial numbers
+
+The serial number has to be the same across VM restarts and storage migration,
+so it has to be stored somewhere, or generated deterministically.
+
+This section lists several options how to store serial numbers.
+Then the rest of the document describes details of option: **Annotation on the PVC object**.
+
+### Annotation on the PVC object
+The serial number will be stored as an annotation on the PVC object. It would be created when the VMI is created.
+Then it will be read by `virt-handler` when creating libvirt domain XML.
+
+#### Advantages:
+- Serial is associated with PVC.
+- Serial is clearly visible to anyone who can read PVC.
+
+#### Disadvantages:
+- Does not store serials for other volume types except PVC.
+
+
+### Storing it in the `status` of the VM
+Serial numbers will be stored in the `.status` of the VM.
+GitOps does not use the `/status` subresource to update the objects it creates, so it should not overwrite the serials.
+More information about `.status` is [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status).
+
+#### Advantages:
+- We can store serial numbers for all volume types.
+- Serial is clearly visible to anyone who can read VM.
+
+#### Disadvantages:
+- Harder to make sure that a disk has the same serial if it is mounted to multiple VMs.
+- It may be an unexpected place to store persistent information that cannot be regenerated if `.status` is wiped.
+
+
+### Deterministic generation
+We can use name and namespace, or UID fields of the PVC to deterministically generate the serial number. 
+For non-PVC volumes, we can use other metadata to generate a serial number.
+
+#### Advantages:
+- We don't need to store anything.
+
+#### Disadvantages:
+- If the storage is migrated to a PVC with different metadata, it would change the serial number.
+
+
+### Storing the serial in a ConfigMap  
+We can create a ConfigMap that would store all disk serials for a VM.
+
+#### Advantages:
+- We can store serial numbers for all volume types.
+
+#### Disadvantages:
+- Keeping the ConfigMap in sync with the VM is more complicated.
+- Harder to make sure that a disk has the same serial if it is mounted to multiple VMs.
+
+
+## API Examples
+
+### Existing API to set serial number manually
+
+The serial number can be specified manually in VM spec here:
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: example-vm
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          disks:
+            - name: disk
+              serial: 1234567890 # <-- serial number
+              disk:
+                bus: virtio
+      volumes:
+        - name: disk
+          persistentVolumeClaim:
+            claimName: example-pvc
+```
+
+### New API
+
+To enable the serial generation, new filed `generateDiskSerials` will be added to the VMI `.spec`:
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: example-vm
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          generateDiskSerials: true # <-- new field
+          disks:
+            - name: disk
+              disk:
+                bus: virtio
+      volumes:
+        - name: disk
+          persistentVolumeClaim:
+            claimName: example-pvc
+```
+
+The generated serial will be stored as an annotation in the PVC:
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: example-pvc
+  annotations:
+    kubevirt.io/disk-serial: 123456789 # <-- new annotation
+spec:
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: 10Gi
+```
+
+## Scalability
+Serials are generated during VMI creation, if the annotation does not exist.
+
+## Update/Rollback Compatibility
+The feature is backward compatible. Existing VMs remain unaffected unless the new field is set.
+Rollback ignores the new field.
+
+## Testing Approach
+
+### Unit tests
+We can add new unit tests to check if the generated libvirt XML contains the `<serial>` tag.
+
+### Functional Tests
+Unit tests will probably be enough. They will check that correct libvirt XML is created.
+
+If we want to add a functional test, we can copy [this test](https://github.com/kubevirt/kubevirt/blob/a37a2b317c335bc1a662d57bd9137d715f52e202/tests/storage/storage.go#L375),
+and modify it to enable `generateDiskSerials` field instead of specifying the serial manually.
+
+## Design questions
+
+- Do we want the serials to be generated by default on all VMs,
+  or do we want a configuration option to enable it?
+  - We would not need any API for generating them. 
+  - The downside is that existing disks that previously did not have a serial configured,
+    will have it. This could confuse the guest OS that expects the disk serial to not change.
+
+- Where is a good place for the `generateDiskSerials` field?
+- Is it a good idea to store the serial in an annotation, or a different approach is better?
+
+## Implementation considerations
+The length of the serial number is limited. The [libvirt documentation](https://libvirt.org/formatdomain.html#hard-drives-floppy-disks-cdroms) says:
+
+> Note that depending on hypervisor and device type the serial number may be truncated silently. IDE/SATA devices are commonly limited to 20 characters. SCSI devices depending on hypervisor version are limited to 20, 36 or 247 characters.
+>
+> Hypervisors may also start rejecting overly long serials instead of truncating them in the future so it's advised to avoid the implicit truncation by testing the desired serial length range with the desired device and hypervisor combination.


akrejcir:
@aglitke , can someone from the storage team take a look, if this approach is good?

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,120 @@
+# Overview
+This proposal introduces a new feature in KubeVirt to automatically
+generate serial numbers for each disk attached to a virtual machine.
+
+## Motivation
+Some applications running in the VM require the disk serial number to be defined and persistent across reboots.
+Generating the serial reduces manual configuration, and in the future it can be part of a `VirtualMachinePreference` for the OS.
```
codingben, line 7: nit: How it's configured manually today by the user? What if the user already configured something and you will generate a new serial number?

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,120 @@
...
+These serial numbers will then be set in the libvirt domain xml.
+
+## Persistence of generated serial numbers  
+
+The serial number has to be the same across VM restarts and storage migration,
+so it has to be stored somewhere, or generated deterministically.
+
+One possibility is to store it in an annotation on the PVC object.
+The annotation would be created when the VMI is created. 
+Then it will be read by `virt-handler` when creating libvirt domain XML.
```
codingben, line 49: Would it be possible to generate serial numbers, after my VM was already created? e.g. VM is created and then I decided to add `generateDiskSerials` to the spec file.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,120 @@
+# Overview
+This proposal introduces a new feature in KubeVirt to automatically
+generate serial numbers for each disk attached to a virtual machine.
```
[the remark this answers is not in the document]
akrejcir: > What will be the length of the serial number?

We can decide that. I will look what is the usual length.

> What if the user wants to have a custom serial number, that is set by him? Can we have that scenario?

There is already a filed for specifying it on the VM here: `.spec.template.spec.domain.devices.disks.[*].serial`. I will add it to this design document.

design-proposals/generated-disk-serial.md
akrejcir: The serial number can be set manually by the user in the path that I've mentioned in the above comment.

If a serial number is already specified, it will not be generated. I will mention it in the document.

design-proposals/generated-disk-serial.md
akrejcir: Yes. It would probably require VM restart.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Motivation
+Some applications running in the VM require the disk serial number to be defined and persistent across reboots.
+Generating the serial reduces manual configuration, and in the future it can be part of a `VirtualMachinePreference` for the OS. 
+
```
dominikholler, line 8: Is there a reference (doc or implemention) how this is addressed in other virtualization management system, especially oVirt and OpenStack?

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,151 @@
...
+- No customization options for serial number format.
+
+## Definition of Users
+
+- VM owners
+
+## User Stories
+
+- As a VM owner, I want to use the automated disk serial number assignment to reduce manual configuration.
+- As a VM owner, I want each disk to have a unique serial number for applications that need it.
```
dominikholler, line 31: Let's be more specific, this application is Windows.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,151 @@
...
+- As a VM owner, I want to use a `VirtualMachinePreference` that is properly configured for the OS without
```
dominikholler, line 32: newly created Windows guests get a serial set on the disks by default. If this is implemented by a preference, or enabled by default always, is part of the question how it is implemented.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Repos
+- [kubevirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+
+A new boolean option `generateDiskSerials` will be added to the VMI `.spec`. 
+When it is set to `true`, kubevirt will generate a serial number for each disk, that does not have one specified.
+These serial numbers will then be set in the libvirt domain xml.
+
```
dominikholler, line 43: This might be a better place to mention the preference

design-proposals/generated-disk-serial.md
dominikholler: The comment will be hard to find after the PR is merged, so it might be better to improve the text to contain the answer.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,151 @@
...
+## Persistence of generated serial numbers  
+
+The serial number has to be the same across VM restarts and storage migration,
+so it has to be stored somewhere, or generated deterministically.
+
+One possibility is to store it in an annotation on the PVC object.
+The annotation would be created when the VMI is created. 
+Then it will be read by `virt-handler` when creating libvirt domain XML.
+
+### Other persistence options
```
dominikholler, line 53: For me it would be more easy to read if there would be a list of all options, with pro and cons, and highlighted in the end which option will be implemented.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,151 @@
...
+## Scalability
+Serials are generated during VMI creation, if the annotation does not exist.
+
+## Update/Rollback Compatibility
+The feature is backward compatible. Existing VMs remain unaffected unless the new field is set.
+Rollback ignores the new field.
+
+## Functional Testing Approach
+
+There are already functional tests to check disk serial that
```
dominikholler, line 142: It would save the readers time to search for the existing test, if they would be linked here.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,151 @@
...
+is specified manually. We can copy these tests and enable `generateDiskSerials`
```
dominikholler, line 143: What would be the cost of this approach? Would it increase the test suite execution time? Is there a way to avoid longer test suite execution times?

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,151 @@
...
+field instead of specifying serial manually.
+
+## Design questions
+
+- Do we want the serials to be generated by default on all VMs,
```
dominikholler, line 148: What are the pros and cons?

dominikholler (changes requested):

design-proposals/generated-disk-serial.md
akrejcir: I will check oVirt code.
@lyarwood , do you know how this is done in OpenStack? Or can you point me to docs or source, where I could find it?

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,151 @@
...
+- Add an option to the VM's `.spec` that will enable automatic serial number generation for disks that
+  don't have a serial number specified.
+
+- Generate unique serial numbers for each disk. The number has to be unique compared to all other disks attached
+  to all VMs with this disk.
+ 
+- Don't store the serial number in the VM's `.spec`. It would confuse gitops operators.
+
+- The serial number should remain the same across storage migration.
+
```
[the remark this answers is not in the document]
akrejcir: I did not find any description of what Widows expects the serial number to be.
The libvirt documentation says this:

>Note that depending on hypervisor and device type the serial number may be truncated silently. IDE/SATA devices are commonly limited to 20 characters. SCSI devices depending on hypervisor version are limited to 20, 36 or 247 characters.
>
>Hypervisors may also start rejecting overly long serials instead of truncating them in the future so it's advised to avoid the implicit truncation by testing the desired serial length range with the desired device and hypervisor combination.

Microsoft documentation describes the structure that is used to retrieve the serial:
https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntddstor/ns-ntddstor-_storage_device_descriptor

There is also this page:
https://learn.microsoft.com/en-us/windows-hardware/drivers/storage/device-unique-identifiers--duids--for-storage-devices
But it only describes what Windows does with the serial number, not what format it expects from it.

I will add the libvirt reccomentations to the dcument.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,151 @@
...
+These serial numbers will then be set in the libvirt domain xml.
+
+## Persistence of generated serial numbers  
+
+The serial number has to be the same across VM restarts and storage migration,
+so it has to be stored somewhere, or generated deterministically.
+
+One possibility is to store it in an annotation on the PVC object.
+The annotation would be created when the VMI is created. 
+Then it will be read by `virt-handler` when creating libvirt domain XML.
```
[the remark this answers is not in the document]
akrejcir: I think the `status` may be emptied by a GitOps controller, when it creates a new version of the VM. I will check.

I will add this possibility to the document for discussion.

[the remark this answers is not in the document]
lyarwood: We should be ignoring and updates of `Status` outside of the `/Status` subresource IIRC so I don't think a GitOps workflow would overwrite it here. That's why we are looking to move instance type metadata out of the `Spec` etc.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

> The PUT and POST verbs on objects MUST ignore the status values, to avoid accidentally overwriting the status in read-modify-write scenarios. A /status subresource MUST be provided to enable system components to update statuses of resources they manage.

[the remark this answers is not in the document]
akrejcir: Thanks. It could work. I will add it to the design document.

design-proposals/generated-disk-serial.md
akrejcir: added

design-proposals/generated-disk-serial.md
akrejcir: I've added text about unit tests.

design-proposals/generated-disk-serial.md
akrejcir: Added.

design-proposals/generated-disk-serial.md
akrejcir: Done.

design-proposals/generated-disk-serial.md
akrejcir: done.

design-proposals/generated-disk-serial.md
akrejcir: done.

design-proposals/generated-disk-serial.md
akrejcir: Thanks!

oVirt uses the disk GUID as a serial number. It would be similar if we used the UID of a PVC.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,202 @@
...
+- As a VM owner, I want to use the automated disk serial number assignment to reduce manual configuration.
+- As a VM owner, I want each disk to have a unique serial number.
+- As a VM owner, I want newly created Windows VMs to have serial numbers set on disks by default.
+
+## Repos
+- [kubevirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+
+A new boolean option `generateDiskSerials` will be added to the VMI `.spec`.
```
0xFelix, line 39: Do we really need an option for it? What's the default right now? Is it 123456789, which means nobody cares about it anyway? Can't we just enable it if the ID was not set manually?

design-proposals/generated-disk-serial.md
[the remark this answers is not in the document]
0xFelix: Same question as for instancetypes. What if the status needs to be recreated? Where is the ID derived from? Can we recreate the same ID if the status needs to be recreated from scratch?

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,202 @@
...
+#### Advantages:
+- Serial is associated with PVC.
+- Serial is clearly visible to anyone who can read PVC.
+
+#### Disadvantages:
+- Does not store serials for other volume types except PVC.
+
+
+### Storing it in the `status` of the VM
+Serial numbers will be stored in the `.status` of the VM.
```
0xFelix, line 67: How to restore / transfer a VM with the serial number? IIUC you are not allowed to write to the status of a K8S object.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,202 @@
...
+#### Advantages:
+- We can store serial numbers for all volume types.
+- Serial is clearly visible to anyone who can read VM.
+
+#### Disadvantages:
+- Harder to make sure that a disk has the same serial if it is mounted to multiple VMs.
+- It may be an unexpected place to store persistent information that cannot be regenerated if `.status` is wiped.
+
+
+### Deterministic generation
```
0xFelix, line 80: I like this approach the most. Basically a stateless approach without configurables.

design-proposals/generated-disk-serial.md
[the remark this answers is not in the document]
akrejcir: That is the main issue with this approach.

If we could deterministically generate the serial number, then we don't need to store it in `.status`.
In this case, the serial number is randomly generated, so it needs to be stored somewhere. If `.status` is recreated, then a new number is generated.

design-proposals/generated-disk-serial.md
akrejcir: I will look into it.

design-proposals/generated-disk-serial.md
akrejcir: It would be the best choice, if it can be compatible with storage migration.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,202 @@
...
+## Persistence of generated serial numbers
+
+The serial number has to be the same across VM restarts and storage migration,
+so it has to be stored somewhere, or generated deterministically.
+
+This section lists several options how to store serial numbers.
+Then the rest of the document describes details of option: **Annotation on the PVC object**.
+
+### Annotation on the PVC object
+The serial number will be stored as an annotation on the PVC object. It would be created when the VMI is created.
```
fabiand, line 55: This will require some component to have PVC write permissions.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,202 @@
...
+This option can also be added to `VirtualMachinePreference`, so it can be set on new VMs and
+the user does not need to be aware of it.
+
+## Persistence of generated serial numbers
+
+The serial number has to be the same across VM restarts and storage migration,
+so it has to be stored somewhere, or generated deterministically.
+
+This section lists several options how to store serial numbers.
+Then the rest of the document describes details of option: **Annotation on the PVC object**.
```
fabiand, line 52: Generally speaking the ID can be scoped in two ways:
- VM/I Internal - serial is only defined/kept within the VM/I bject
- VM/I external - serial is sourced from an external object - this is the case with the PVC annotation approach

If it's internal, then usually the requirements of a persistent serial number are met, however, it could lead to trouble if a volume is changed on a VMI, or if a volume is used on multiple VMIs.

If it is external, then we also meet the requirement of a persistent serial number, the same volume will get the same serial whenever attached to a VM that's good for vols attached to multi VMs, however, we need to think about cloning, and restore, and what it means for disk serial numbers. Something also requires permissionsto write the PVC.

Just to name a few things.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,202 @@
...
+### Annotation on the PVC object
+The serial number will be stored as an annotation on the PVC object. It would be created when the VMI is created.
+Then it will be read by `virt-handler` when creating libvirt domain XML.
+
+#### Advantages:
+- Serial is associated with PVC.
+- Serial is clearly visible to anyone who can read PVC.
+
+#### Disadvantages:
+- Does not store serials for other volume types except PVC.
```
fabiand, line 63: Requires additional permissions to write PVCs

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Deterministic generation
+We can use name and namespace, or UID fields of the PVC to deterministically generate the serial number. 
+For non-PVC volumes, we can use other metadata to generate a serial number.
+
+#### Advantages:
+- We don't need to store anything.
+
+#### Disadvantages:
+- If the storage is migrated to a PVC with different metadata, it would change the serial number.
```
fabiand, line 88: Usually we do not "migrate" PVCs (aka modify them in-place).
Instead new PVCs are created. @alicefr @awels what does storage class migration do?

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+The generated serial will be stored as an annotation in the PVC:
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: example-pvc
+  annotations:
+    kubevirt.io/disk-serial: 123456789 # <-- new annotation
```
fabiand, line 159: please use `vm.kubevirt.io/disk-serial`

cc @EdDev @alaypatel07 can we have api conventions? Possibly worth to standardize how we prefix annotations accross kubevirt.

design-proposals/generated-disk-serial.md
```diff
@@ -0,0 +1,202 @@
...
+    requests:
+      storage: 10Gi
+```
+
+## Scalability
+Serials are generated during VMI creation, if the annotation does not exist.
+
+## Update/Rollback Compatibility
+The feature is backward compatible. Existing VMs remain unaffected unless the new field is set.
+Rollback ignores the new field.
```
fabiand, line 173: We should check todays state:

If a serial is not specified, between VM start, stop, start, does the serial of the disk inside the VM change?

If it is the same serial, then we might need a migraiton plan.

design-proposals/generated-disk-serial.md
awels: Storage class migration requires one to create new PVCs as the target of the migration. So there will be new PVCs. PVCs themselves are mostly immutable, in particular the name is immutable.

design-proposals/generated-disk-serial.md
lyarwood: I think @0xFelix 's point was that we then need a way of persisting the serial in the core spec of a VirtualMachine during a snapshot so that we can then restore the status (and the state of the VM) later.

PR #308, opened 2024-07-09
Removing demos folder

aburdenthehand:
I intially created this folder after collecting some silent and notation-less demos for use at a KubeCon booth, and thought it would be useful to include them here along with the resulting transcript. 

It was pointed out that this adds significant weight to any git clone or update. Our youtube (with notation or dub) or googledrive are better hosting platforms. 

**Release note**:

```release-note
NONE
```


alicefr:
@aburdenthehand can we upload the video on youtube and then simply share the link?

aburdenthehand:
@alicefr We can, but I think this particular demo is pretty dated by now. I think we would be better off recording an updated demo with either on-screen description wipes or voice commentary; the latter of which is preferable as it can generate subtitles, which can also be auto-translated by youtube for a broader audience.

alicefr:
> @alicefr We can, but I think this particular demo is pretty dated by now. I think we would be better off recording an updated demo with either on-screen description wipes or voice commentary; the latter of which is preferable as it can generate subtitles, which can also be auto-translated by youtube for a broader audience.

Ok, make sense

xpivarc (approved):
note: it will still leave on in our git

jean-edouard:
@aburdenthehand see the comment from @xpivarc, this PR won't solve the slow git checkout. Do you still think this PR makes sense?
Or do we maybe want to consider altering history to truly remove the files? (would that even do the trick?)

dankenigsberg:
As every time traveler would tell, modifying history is a delicate business. If you decide to do so, make sure to keep an archive of the current repo prior to the change (e.g community-with-demo) so it can be verified that the filetree of the last commit in the old repo is exactly the same as that of the first commit of the new repo.

PR #314, opened 2024-07-23
design proposal: new migration process for backend-storage PVCs

jean-edouard:
```release-note
NONE
```


dankenigsberg (changes requested):
Hi @jean-edouard ! Thanks for this much-needed work. I have a few questions, whose answers can be incorporated into the proposal?

design-proposals/backend-storage-block-migration.md
```diff
@@ -0,0 +1,62 @@
...
+kubevirt/kubevirt
+
+# Design
+- When creating a new VM with persistent features, we use either (first match wins):
+  - The storage class specified in the KubeVirt Custom Resource
+  - The storage class marked as the default for virtualization purposes
+  - The storage class marked as the default for the cluster
+- All new persistent storage PVCs will be created with the name `persistent-state-for-<vm_name>-<random_string>` and the new annotation `persistent-state-for-<vm_name>=true`
+- When starting an existing VM with persistent features, we will look for any PVC with the annotation, or any PVC with the legacy name, for backwards compatibility
+- For modes, the filesystem volume mode will be preferred (less complexity), and we default to RWO (because RWX won't be needed anymore)
```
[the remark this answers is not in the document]
jean-edouard: Yes, the storage profile will definitely be used, but if multiple access modes are available we still need to pick one

design-proposals/backend-storage-block-migration.md
```diff
@@ -0,0 +1,62 @@
...
+- On migration, no matter what volume/access mode the PVC uses, we create a new empty PVC with the name `persistent-state-for-<vm_name>-<random_string>` and no annotation
+- In the migration object status, we store the names of both the source and target PVCs (see API section below)
+- If a migration succeeds, we set the annotation `persistent-state-for-<vm_name>=true` on the new (target) PVC and delete the old (source) PVC, using the source PVC name from the migration object
```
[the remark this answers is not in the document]
jean-edouard: This is why status fields are added to the migration object. The migration won't move to the finished state as long as the PVCs are not sorted out

design-proposals/backend-storage-block-migration.md
```diff
@@ -0,0 +1,62 @@
...
+
+As of KubeVirt 1.3, Here is how each of the 4 combinations is supported by the backend-storage for persistent VM state:
+- RWO FS: supported by backend-storage but makes VMs non-migratable
+- RWX FS: preferred combination, supported by backend-storage and compatible with VM migrations
+- RWO block: not supported yet, but PRs exist to implement it in a non-migratable way 
+- RWX block: same as above, with migratability functional but considered unsafe (as it concurrently mounts an ext3 partition)
+
+This is the best we can do with the current implementation, which involves mounting the same partition on migration source and target.
+This proposal is about switching to a different model which would instead create a new backend-storage PVC on migration,
+allowing all 4 cases above to have proper migration support.
```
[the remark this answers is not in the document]
jean-edouard: The point of this proposal is to stop mounting the PVC on both source and target, to enable migratability of block-based backend-storage.

design-proposals/backend-storage-block-migration.md
```diff
@@ -0,0 +1,62 @@
...
+
+To make this document as simple as possible, let's consider that storage in Kubernetes can either be filesystem (FS) or
+block mode, and either ReadWriteOnce (RWO) or ReadWriteMany (RWX) access mode, with storage classes supporting any
+number of combinations of both.
+
+As of KubeVirt 1.3, Here is how each of the 4 combinations is supported by the backend-storage for persistent VM state:
+- RWO FS: supported by backend-storage but makes VMs non-migratable
+- RWX FS: preferred combination, supported by backend-storage and compatible with VM migrations
+- RWO block: not supported yet, but PRs exist to implement it in a non-migratable way 
+- RWX block: same as above, with migratability functional but considered unsafe (as it concurrently mounts an ext3 partition)
```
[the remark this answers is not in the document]
jean-edouard: This is because, unlike with VM disks, we have 2 components concurrently accessing the PVC (source and target virt-launchers).
They both maintain the filesystem metadata for the mounted partition, which will lead to conflicts. I can add more verbiage to explain that in the doc.

design-proposals/backend-storage-block-migration.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Scalability
+No new scalability concern will be introduced
+
+## Update/Rollback Compatibility
+Since the name of the backend-storage PVC will change, we will keep fallback code to look for the legacy PVC.
+
+## Functional Testing Approach
+All existing backend-storage-related functional tests will still apply. More tests could be added to ensure all 4 combinations
+of FS/block RWX/RWO work for backend-storage, if we think it's worth the additional stress on the e2e test lanes.
```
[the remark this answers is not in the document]
jean-edouard: Yes, this is definitely something to keep in mind when adding functional test coverage for this. It's indeed what I was trying to hint at there

jean-edouard:
> Hi @jean-edouard ! Thanks for this much-needed work. I have a few questions, whose answers can be incorporated into the proposal?

Thank you for the review! Comments answered in github, I will amend the design proposal to reflect them.

design-proposals/backend-storage-block-migration.md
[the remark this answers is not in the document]
dankenigsberg: > I can add more verbiage to explain that in the doc.

Please do. In particularly, please refer to my question

> Does libvirt have the same problem? If the only shared storage a user has is block, can they use vTPM+migration?

design-proposals/backend-storage-block-migration.md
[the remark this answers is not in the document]
dankenigsberg: I do not understand why we are interested with RWO. If we have only RWO, we cannot migrate the root disk. I don't understand why we conflate this proposal (and future tests) with this combination (migration+RWO).

design-proposals/backend-storage-block-migration.md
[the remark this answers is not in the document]
dankenigsberg: Please include a deeper discussion of the mode of failure. I don't understand how you can coordinate between the source and destination without interaction with qemu.

dankenigsberg (changes requested):

design-proposals/backend-storage-block-migration.md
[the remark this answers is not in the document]
stu-gott: Currently the TPM implementation uses multiple files under the hood.

This means one of two things must be true. Either some component in the stack must manage a filesystem on the block device or the TPM implementation must be rewritten from the ground up. The latter doesn't seem like a fair request to the underlying projects. At least not in a timeline that suits our needs.

The filesystem being used is ext3, so it can either be mounted on the source or destination only. These mounts are mutually exclusive with each other regardless of whether or not the PVC is using RWX mode. This is simply a limitation of ext3 (or most filesystems for that matter).

Given these constraints, KubeVirt is the best component to manage this problem because it can manage PVCs where libvirt can't--because libvirt runs inside a container.

design-proposals/backend-storage-block-migration.md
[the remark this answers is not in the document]
vladikr: The backend-sotage may be set to a different storage class than the other disks.
What is being discussed here is that since the content of this volume will be copied, the access mode of such volume will become irrelevant and might as well be RWO.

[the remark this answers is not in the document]
dankenigsberg: Thinking about RWO and discussing it makes this proposal more complex and slower to test. What is the benefit of it?

design-proposals/backend-storage-block-migration.md
```diff
@@ -0,0 +1,67 @@
...
+  - When multiple access modes are listed for the selected volume mode, we will arbitrarily default to RWO.
+  - When CDI is not present, or the StorageClass has no StorageProfile, or the StorageProfile has no modes, we'll default to RWO FS.
+- On migration:
+  - No matter what volume/access mode the PVC uses, we will create a new empty PVC for the target with the name `persistent-state-for-<vm_name>-<random_string>` and no annotation
+  - In the migration object status, we store the names of both the source and target PVCs (see API section below)
+  - If a migration succeeds, we set the annotation `persistent-state-for-<vm_name>=true` on the new (target) PVC and delete the old (source) PVC, using the source PVC name from the migration object
+  - If a migration fails, we delete the target backend-storage PVC that was just created, using the target PVC name from the migration object
+  - In the unlifely event that a VM shuts down towards the very end of a migration, the migration finalizer will decide which PVC to keep and which one to get rid of
```
dankenigsberg, line 47: how could the finalizer make this decision? I don't think it has enough information.

design-proposals/backend-storage-block-migration.md
```diff
@@ -0,0 +1,67 @@
...
+block mode, and either ReadWriteOnce (RWO) or ReadWriteMany (RWX) access mode, with storage classes supporting any
+number of combinations of both.
+
+As of KubeVirt 1.3, Here is how each of the 4 combinations is supported by the backend-storage for persistent VM state:
+- RWO FS: supported by backend-storage but makes VMs non-migratable
+- RWX FS: preferred combination, supported by backend-storage and compatible with VM migrations
+- RWO block: not supported yet, but PRs exist to implement it in a non-migratable way 
+- RWX block: same as above, with migratability functional but considered unsafe, because it leads to 2 pods (source and target virt-launchers) mounting the same ext3 filesystem concurrently.
+
+This is the best we can do with the current implementation, which involves mounting the same partition on migration source and target.
```
dankenigsberg, line 14: Would you elaborate about partitioning? I see "partitioning complexity" below, but the significance is not clear to me.

What do you mean by "current implementation"? KubeVirt's? libvirt's? Have we considered other (possibly better, yet more intrusive) implementations?

design-proposals/backend-storage-block-migration.md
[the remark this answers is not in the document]
dankenigsberg: @stu-gott do I understand that your the answer to my question is affirmative? libvirt has the same problem and cannot cannot use shared block+vTPM+migration?

Have we asked them to support this combination?
Do they not mind?

dankenigsberg (changes requested):

design-proposals/backend-storage-block-migration.md
vladikr: swtpm, qemu and/or libvirt don't support saving tpm state on block storage.
libvirt supports saving this state in an in-memory secret but it won't be migrateable.
live migration is only supported with a state saved on shared storage.
There is no plan to support any other implementation. To the best of my knowledge.

To support block storage as backed for vTPM, kubevirt will need to have access to in the swtpm container, build a filesystem on top, and mount it. 
The issues arise when kubevirt attempts to mount this volume on both the source and the target containers. 

I hope it answers your question.


This proposal suggests creating two separate block volumes and allowing qemu to copy the state, however, it will require kubevirt to manage the volumes on migration - which creates other challenges to overcome.

design-proposals/backend-storage-block-migration.md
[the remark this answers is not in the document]
vladikr: From my side, we can ignore it. However, in many cases, this is the RWO is the default access mode. We don't have to test or support this.

design-proposals/backend-storage-block-migration.md
vladikr: The finalizer itself doesn't solve this issue. It only makes sure that the migration object won't get deleted before the correct volume is annotated by the controller.

However, there is still a race that this proposal will need to address. 

There is a short period when the migration is finalized on the host, but before the controller has annotated the correct volume. The VM could shutdown and start again, picking up an incorrect volume.

I think that this issue can be addressed by simply re-queueing the newly created VMI object if it has an unfinalized migration. 
This is an issue on its own that a new VM can start while it has an unfinished migration.

design-proposals/backend-storage-block-migration.md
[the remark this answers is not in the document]
vladikr: @dankenigsberg I answered in another comment. The answer is, to the best of my knowledge, no.
swtpm, qemu and/or libvirt don't support saving tpm state on block storage and there is no plan to support this.
If we want to support this scenario in KubeVirt we will need to provide the management around this.

dankenigsberg (changes requested):

design-proposals/backend-storage-block-migration.md
vladikr: Yes, I agree. @jean-edouard could you please highlight this race and the suggested solution?

design-proposals/backend-storage-block-migration.md
vladikr: To my knowledge, other platforms are fine with a shared filesystem volume for migration or don't support it entirely[1].

IIRC, we asked. @jean-edouard @stu-gott could add your thoughts here?

OpenStack Nova: "Live migration, evacuation, shelving and rescuing of servers with vTPMs is not currently supported."
[1] https://docs.openstack.org/nova/latest/admin/emulated-tpm.html

vladikr: @xpivarc pointed out that it looks like swtpm supports block storage, so I take my previous statement back until we figure out if we can use this option.

jean-edouard:
This could also be a concern: https://github.com/libvirt/libvirt/blob/e40a533118efdc3d5f6fd1dcc9bb8143b54691e2/src/qemu/qemu_tpm.c#L571
(if the TPM is stored under a "shared" (nfs/ceph) storage, it looks like it's automatically considered as shared between the source and the target)

design-proposals/backend-storage-block-migration.md
jean-edouard: Indeed, with `--tpmstate backend-uri=file://`, swtpm can serialize its files into something like a block storage device.
This option is however not supported by libvirt. That also doesn't work with EFI.

design-proposals/backend-storage-block-migration.md
```diff
@@ -0,0 +1,67 @@
...
+
+# Design
+- When creating a new VM with persistent features, we use either (first match wins):
+  - The storage class specified in the KubeVirt Custom Resource
+  - The storage class marked as the default for virtualization purposes
+  - The storage class marked as the default for the cluster
+- All new persistent storage PVCs will be created with the name `persistent-state-for-<vm_name>-<random_string>` and the new annotation `persistent-state-for-<vm_name>=true`
+- When starting an existing VM with persistent features, we will look for any PVC with the annotation, or any PVC with the legacy name, for backwards compatibility
+- For volume/access modes:
+  - When multiple volume modes are listed in the StorageProfile of the StorageClass, we will default to filesystem (to avoid partitioning complexity)
```
mhenriks, line 39: I think we can avoid partitioning entirely.  Any StorageClass that supports block should natively support filesystem (like rook-ceph-block)

design-proposals/backend-storage-block-migration.md
```diff
@@ -0,0 +1,67 @@
...
+  - When multiple access modes are listed for the selected volume mode, we will arbitrarily default to RWO.
+  - When CDI is not present, or the StorageClass has no StorageProfile, or the StorageProfile has no modes, we'll default to RWO FS.
```
mhenriks, line 41: RWO FS is the lowest common denominator in k8s you will be hard pressed to find a provisioner that does not support that. Not sure we should even consider any other combinations

design-proposals/backend-storage-block-migration.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Update/Rollback Compatibility
+Since the name of the backend-storage PVC will change, we will keep fallback code to look for the legacy PVC.
+
+## Functional Testing Approach
+All existing backend-storage-related functional tests will still apply. More tests could be added to ensure all 4 combinations
+of FS/block RWX/RWO work for backend-storage, if we think it's worth the additional stress on the e2e test lanes. 
+
+# Implementation Phases
+First phase is block support. That effort is well underway already.
```
mhenriks, line 65: I don't think it is worth investing in block PVC support. What provisioner only supports block PVCs?

design-proposals/backend-storage-block-migration.md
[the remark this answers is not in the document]
mhenriks: What about containerdisk VMs?  They can be live migrated, correct?  Can we use emptydir volumes for backend storage in that case?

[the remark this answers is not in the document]
jean-edouard: Using emptydir for backend storage is effectively what happens for non-persistent TPM/EFI.
If we want persistence, we need a PVC.

design-proposals/backend-storage-block-migration.md
jean-edouard: That's huge. Block support is messy and somewhat illogical. Reducing this design to only add RWO FS support would make everything so much simpler and sensible.

design-proposals/backend-storage-block-migration.md
dankenigsberg: > RWO FS is the lowest common denominator in k8s you will be hard pressed to find a provisioner that does not support that

That's a very important note. How common is RWX FS which we already support with migration?

There may be operational problems with splitting VM persistence between several storage classes (disaster recovery and storage migration may become more complex, virt-default StorageClass) which may keep this proposal useful.

mhenriks: > That's a very important note. How common is RWX FS which we already support with migration?

RWX FS is not very popular.  Shared filesystems have been declining in popularity since object storage entered the scene.  Shared filesystems are generally considered complex and slow. For legacy applications. Avoided whenever possible

> There may be operational problems with splitting VM persistence between several storage classes (disaster recovery and storage migration may become more complex, virt-default StorageClass) which may keep this proposal useful.

I'm not following this.  Where's the splitting happening?  My point is that the same storageclass can be used for block and fs.  For example the rook-ceph-block StorageClass can do RWX block and RWO fs.

dankenigsberg: > For example the rook-ceph-block StorageClass can do RWX block and RWO fs.

I didn't know that. Thanks.

So to support TPM on `rook-ceph-block` we have two options: either ask libvirt to support RWX block, or add our own support for RWO file. This proposal should discuss the pros and cons of each.

jean-edouard: Well this proposal is all about option 2, but I guess we could touch on option 1 in the intro.
A brief pros/cons:
Option 1 requires some work on our side (multi-PVC support) and a lot of work on libvirt's (add support for swtpm block + work with qemu to design and implement EFI block). The main con IMO is the unreasonable ask of libvirt/qemu to support pre-1961 technology (disks without partitions/filesystems).
Option 2 just requires implementing this design proposal, including ensuring there's no race when rebooting during a migration. Seems a lot saner to me, even if it's more work for us.

mhenriks: > work with qemu to design and implement EFI block

Seems like a big chunk of work for another team and would ultimately require 2 PVCs for VMs with EFI+TPM, correct?  With filesystem PVCs, TPM and EFI can share a single PVC, right?

mhenriks:
> This could also be a concern: https://github.com/libvirt/libvirt/blob/e40a533118efdc3d5f6fd1dcc9bb8143b54691e2/src/qemu/qemu_tpm.c#L571 (if the TPM is stored under a "shared" (nfs/ceph) storage, it looks like it's automatically considered as shared between the source and the target)

Interesting. So could be possible to ask for a RWO FileSystem PVC, get nfs, and at migration time, data will not be copied to the new PVC?

I believe we've encountered a similar problem before, but the reverse. User requests RWX PVC but get ext4 because VM is scheduled on the same node as the nfs server. I believe LD_PRELOAD hacks were used to get around the issue until there was a proper fix

design-proposals/backend-storage-block-migration.md
jean-edouard: Exactly yes

mhenriks: Fewer PVCs is big PRO for option 2

design-proposals/backend-storage-clone-on-migration.md
```diff
@@ -0,0 +1,82 @@
...
+That, however, is out-of-scope for this design proposal, and may never be needed, since most/all block storage classes also support RWO FS.
+
+## Alternative approach
+The alternative to this design is for upstream components (libvirt - qemu - swtpm) to add block storage support for TPM/EFI.  
+KubeVirt could then create one backend-storage PVC per feature (EFI/TPM) and pass them directly to libvirt.  
+That alternative approach would enable RWX block backend-storage (as opposed to RWO FS in this proposal).  
+
+Pros:
+- Less code to write and maintain in KubeVirt
+- No need to duplicate backend-storage PVCs on migrations
```
dankenigsberg, line 27:
```suggestion
- No need to duplicate backend-storage PVCs on migrations and delete the correct one after migration success/failure
- libvirt gains an important feature
```

dankenigsberg: Please file a public request on libvirt to add this functionality, and include the URL to it here. The simplest way to "kill" this alternative is if libvirt rejects it. I would think that this would be a mistake on their behalf, but that's their question to answer.

dankenigsberg (changes requested):

design-proposals/backend-storage-clone-on-migration.md
jean-edouard: I discussed this with a few members of the libvirt team, and they actually don't appear to have strong objections against it.
There would be 4 parts to this:
- block storage support in swtpm: already done
- libvirt glue for swtpm block storage: draft PR open (https://gitlab.com/libvirt/libvirt/-/merge_requests/380)
- block storage support in qemu for NVRAM: could work out of the box, since the NVRAM is already just 1 file (just have to ensure they don't run file-specific operations on it)
- libvirt glue for NVRAM block storage: could also work out of the box, although unlikely, but shouldn't require too much work (just ensuring the "file" is never moved/deleted/chmoded...)

So really not to much left to do overall, however we would have to wait for the first release of every component that needs changes, which could take a long time.
This approach is also definitely not my first choice, and I would hate to request something, have upstream devs work hard on it and ultimately not consume it...
That's why I didn't file any request yet.

design-proposals/backend-storage-clone-on-migration.md
```diff
@@ -0,0 +1,82 @@
...
+
+To make this document as simple as possible, let's consider that storage in Kubernetes can either be filesystem (FS) or
+block mode, and either ReadWriteOnce (RWO) or ReadWriteMany (RWX) access mode, with storage classes supporting any
+number of combinations of both.
+
+As of KubeVirt 1.3, Here is how each of the 4 combinations is supported by the backend-storage for persistent VM state:
+- RWO FS: supported by backend-storage but makes VMs non-migratable. **This entire design proposal is about removing that limitation.**
+- RWX FS: preferred combination, supported by backend-storage and compatible with VM migrations
+- RWO block: not supported yet, but PRs exist to implement it in a non-migratable way 
+- RWX block: same as above, with migratability functional but considered unsafe, because it leads to 2 pods (source and target virt-launchers) mounting the same ext3 filesystem concurrently.
```
alicefr, line 12: Sorry, I'm a bit late on this. Have we ever consider to use a clustered filesystem like gfs2 instead of ext3?

jean-edouard: That's a really good question. `gfs2` somehow didn't pop up when I looked for potentially suitable filesystems for formatting the block storage, probably worth investigating!
That being said, this proposal (and even the alternative it contains) doesn't consider the option of formatting the block storage (anymore), and not just because I couldn't find a suitable filesystem.
The fact that it requires adding a bunch of tools to virt-launcher and virt-handler, and the general complexity of managing partitions ourselves, make it something we should avoid, in my opinion (see [draft PR](https://github.com/kubevirt/kubevirt/pull/12337)).

alicefr: Yes, I agree. I read a bit about gfs2 supported by Red Hat, and it requires clvm. So, I think this also exclude it as a solution

design-proposals/backend-storage-clone-on-migration.md
dankenigsberg: > This approach is also definitely not my first choice

Please make sure to spell out all your reasoning as its "cons". For example, I did not see the issue of timing written in the proposal.

> That's why I didn't file any request yet.

Let have a formal request and link to it here. We should be explicit that because the reasons spelled out here we may not use it, or may not use it immediately. But let us give the libvirt team an opportunity to refute our assumptions - maybe they can deliver this functionality fast?

design-proposals/backend-storage-clone-on-migration/backend-storage-clone-on-migration.md
```diff
@@ -0,0 +1,92 @@
...
+block mode, and either ReadWriteOnce (RWO) or ReadWriteMany (RWX) access mode, with storage classes supporting any
+number of combinations of both.
+
+As of KubeVirt 1.3, Here is how each of the 4 combinations is supported by the backend-storage for persistent VM state:
+- RWO FS: supported by backend-storage but makes VMs non-migratable. **This entire design proposal is about removing that limitation.**
+- RWX FS: preferred combination, supported by backend-storage and compatible with VM migrations
+- RWO block: not supported yet, but PRs exist to implement it in a non-migratable way 
+- RWX block: same as above, with migratability functional but considered unsafe, because it leads to 2 pods (source and target virt-launchers) mounting the same ext3 filesystem concurrently.
+
+The above is the best we can do with the current implementation, which involves mounting the same partition on migration source and target.
```
dankenigsberg, line 14:
```suggestion
The above is the best we can do with the current libvirt implementation, which involves mounting the same partition on migration source and target.  
```

I'm not at all sure I understand to which "current implementation" you refer.

jean-edouard: I'll rephrase, what I mean is "without partitioning block / creating new PVCs on migration / libvirt block support"

design-proposals/backend-storage-clone-on-migration.md
jean-edouard: > Please file a public request on libvirt to add this functionality, and include the URL to it here.

https://issues.redhat.com/browse/RHEL-54325
https://issues.redhat.com/browse/RHEL-54326

dankenigsberg: thanks. github comments can easily get lost. Please add these to the design doc - the fact that they are not done is a reason against the alternative option.

jean-edouard: Yes, I will rework this design proposal this week and include those links.

design-proposals/backend-storage-clone-on-migration/backend-storage-clone-on-migration.md
```diff
@@ -0,0 +1,106 @@
...
+- Less code to write and maintain in KubeVirt
+- No need to duplicate backend-storage PVCs on migrations and delete the correct one after migration success/failure
+- Libvirt gains a feature
+
+Cons:
+- Relies on third party projects adding and maintaining options to store config files into a block devices without filesystems
+- Requires multiple PVCs per VM (one per persistence-enabled feature)
+- We would have to either
+  - Maintain 2 different backend-storage implementations (the existing RWX FS one and this RWX block one)
+  - Deprecate RWX FS and require RWX block to enable backend-storage
```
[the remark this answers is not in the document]
jean-edouard: With the primary design, we keep support for RWX FS and extend it to support RWO FS. Since both use virtually the same codepath, it's just one thing to maintain.
However, RWX block would need a completely different approach, so if we want to avoid maintaining 2 different ways to do the same thing we'd have to deprecate RWX FS.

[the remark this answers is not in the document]
dankenigsberg: I don't know the code, but ideally, by delegating the trouble to libvirt, the only difference between RWX FS and RWX Block would be passing a block device to libvirt instead of a filesystem path.

In the primary implementation, we need to add code to synchronize between source and destination during success and various modes of failures.

design-proposals/backend-storage-clone-on-migration/backend-storage-clone-on-migration.md
```diff
@@ -0,0 +1,106 @@
...
+- When starting an existing VM with persistent features, we will look for any PVC with the annotation, or any PVC with the legacy name, for backwards compatibility
+- The volume mode will be Filesystem
+- The access mode will be RWO, unless a StorageProfile exists for the Storage Class and shows only RWX is supported
+  - Note: the CR field documentation will be adjusted to reflect that RWX is no longer needed
+- When a migration is created:
+  - We create a new empty PVC for the target, with the name `persistent-state-for-<vm_name>-<random_string>` and no annotation
+  - In the migration object status, we store the names of both the source and target PVCs (see API section below)
+  - If a migration succeeds, we set the annotation `persistent-state-for-<vm_name>=true` on the new (target) PVC and delete the old (source) PVC, using the source PVC name from the migration object
+  - If a migration fails, we delete the target backend-storage PVC that was just created, using the target PVC name from the migration object
```
dankenigsberg, line 74: Does this hold true also with post-copy migration?

design-proposals/backend-storage-clone-on-migration/backend-storage-clone-on-migration.md
```diff
@@ -0,0 +1,106 @@
...
+  - In the unlikely event that a VM shuts down towards the very end of a migration, the migration finalizer will decide which PVC to keep and which one to get rid of
```
dankenigsberg, line 75: How would the migration finalizer know which of the two PVCs is the correct one?

dankenigsberg (changes requested):
The toughest issue in the primary solution is handling migration and VM failures (and the reason I prefer the alternative solution). Would you elaborate on that issue?

design-proposals/backend-storage-clone-on-migration/backend-storage-clone-on-migration.md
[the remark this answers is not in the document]
jean-edouard: > I don't know the code, but ideally, by delegating the trouble to libvirt, the only difference between RWX FS and RWX Block would be passing a block device to libvirt instead of a filesystem path.

Yes, well multiple block devices. That means changing every function that touches backend-storage to handle multiple PVCs.
It would also require changing the volumes and mounts we pass to virt-launcher. Finally the libvirt options would indeed look different.
That is a non-trivial amount of differences.

> In the primary implementation, we need to add code to synchronize between source and destination during success and various modes of failures.

Yes, before we can mark a (RWO) migration as completed we have to add a label to a PVC and delete another one. It is an extra step, and we have to make sure it's done right, but it's not much code at all.

design-proposals/backend-storage-clone-on-migration/backend-storage-clone-on-migration.md
jean-edouard: Yes, migration mode doesn't make a difference here.

design-proposals/backend-storage-clone-on-migration/backend-storage-clone-on-migration.md
jean-edouard: This needs to be reworded. If the VMI disappears (i.e. shutdown), the migration is automatically failed, and we fall into the bullet point above.

design-proposals/backend-storage-clone-on-migration/backend-storage-clone-on-migration.md
```diff
@@ -0,0 +1,105 @@
...
+- When starting an existing VM with persistent features, we will look for any PVC with the annotation, or any PVC with the legacy name, for backwards compatibility
+- The volume mode will be Filesystem
+- The access mode will be RWO, unless a StorageProfile exists for the Storage Class and shows only RWX is supported
+  - Note: the CR field documentation will be adjusted to reflect that RWX is no longer needed
+- When a migration is created:
+  - We create a new empty PVC for the target, with the name `persistent-state-for-<vm_name>-<random_string>` and no annotation
+  - In the migration object status, we store the names of both the source and target PVCs (see API section below)
+  - If a migration succeeds, we set the annotation `persistent-state-for-<vm_name>=true` on the new (target) PVC and delete the old (source) PVC, using the source PVC name from the migration object
+  - If a migration fails, we delete the target backend-storage PVC that was just created, using the target PVC name from the migration object
+  - In the unlikely event that a VM shuts down during a migration, the migration will be considered failed and the target PVC will get discarded
```
dankenigsberg, line 75: What if right after migration ends successfully in the libvirt level, the VM makes a change to the TPM and shuts down? Discarding the target PVC would make the data lost.

design-proposals/backend-storage-clone-on-migration/backend-storage-clone-on-migration.md
jean-edouard: Your concern is valid, but it's just a part of the dangers of post-copy. Yes, if a guest is ongoing a post-copy migration and, after the handoff, makes changes to the TPM, then before the end of the migration someone trips on the network cable, the changes made to the TPM will be lost. But that's exactly why 1) post-copy comes with a big warning, 2) bitlocker strongly encourages you to write down your recovery key.
You can also imagine the opposite scenario, if we decided to keep the target PVC and the (post-handoff post-copy) migration failed right during a change to the TPM. The PVC on the target would now contain a corrupt state and switching to it would break the VM (that's what I was referring to above). Same concern applies in RWX btw.
Post-copy migration might break stuff, I don't think there's any way around it...

design-proposals/backend-storage-clone-on-migration/backend-storage-clone-on-migration.md
jean-edouard: Yes, see https://github.com/kubevirt/community/pull/314#discussion_r1791924517

jean-edouard: This is not a KubeVirt problem, same thing would happen on bare-metal... If one doesn't find the risks of post-copy acceptable, they should just not use it.
What we can and should do is document those risks thoroughly.

jean-edouard: > What if right after migration ends successfully in the libvirt level, the VM makes a change to the TPM and shuts down? Discarding the target PVC would make the data lost.

Sorry I misread your question. If a libvirt migration ends successfully then the KubeVirt migration will also end successfully and all will be fine.
I read something like "What if right after migration handoff [...]", my bad.

stu-gott: Hi @dankenigsberg can you clarify what user scenario you're focused on/concerned about? In all cases, a machine only modifies the registers of its TPM during initial power on. The protected contents are only changed if the user changes their LUKS settings (or windows equivalent). This is not the same as boot order type changes. this is setting the initial password for your encrypted volume. This is not a sensible thing to do immediately before shutdown.

vladikr: I think it is a valid concern.

> > What if right after migration ends successfully in the libvirt level, the VM makes a change to the TPM and shuts down? Discarding the target PVC would make the data lost.
> 
> Sorry I misread your question. If a libvirt migration ends successfully then the KubeVirt migration will also end successfully and all will be fine. I read something like "What if right after migration handoff [...]", my bad.

KubeVirt finalizes the migration in the source pod. It relies on getJobStats which is being retrieved every 400ms.
There is a period of time (less than 400ms) during which the migration can potentially be finalized but the metadata won't be updated.
If something happens to the source pod at this time and the target manages to write to the new volume, we will be in trouble.

jean-edouard: > If something happens to the source pod at this time and the target manages to write to the new volume, we will be in trouble.

Could you please detail this scenario a bit more? Are we still talking about a failed post-copy migration? Also what do you mean by "something happens to the source pod"?
Thanks!

dankenigsberg: > "something happens to the source pod"

e.g the node running it can crash.

Also, the VM can finish cleanly or crash during the window between libvirt's end of migration and etcd being aware that migration ended successfully. The guest may write everything to its storage and tpm, but we end up deleting the tpm.

vladikr: > > If something happens to the source pod at this time and the target manages to write to the new volume, we will be in trouble.
> 
> Could you please detail this scenario a bit more? Are we still talking about a failed post-copy migration? Also what do you mean by "something happens to the source pod"? Thanks!

Sure. Consider a regular pre-copy migration.
libvirt completed the migration and unpaused the target.
The migration results to kubevirt objects take some time to propagate (speaking in general).
If the source pod is terminated (due to node crash, OOM, kubelet killing it) and the target writes the TPM state to the new volume during this window the target volume will be lost.

jean-edouard: I see. To repeat back and summarize, if all the following is true we have a problem:

- We're inside the, at most 400ms, window when the libvirt migration is finished but KubeVirt doesn't know yet
- The source pod/node crashes
- The VM writes to the TPM

Then the VMIM will fail and that TPM write will be dropped.
Honestly, to me, this scenario feels quite statistically impossible...
If we want to address this, do you have a suggestion how?
I brainstormed this with @stu-gott and this is the only thing we could come up with:

- Libvirt+swtpm implement a state lock that's automatically enabled at the end of a migration
- Once KubeVirt sees that the libvirt migration succeeded, it instructs libvirt to release that lock

If that makes sense to you we can discuss it with the libvirt team.
However, would you agree to do that as a follow-up effort?

Thanks!

vladikr: The 400ms is only the time it takes for the migration monitor to update the domain metadata.
Then there is also the time required to propagate the migration status to other kubevirt components/objects.
Overall, it's possible that this can be more than just a statistically insignificant issue.

I didn't have much time to think this through.
However, If we are going to continue with this effort we need to come up with a mechanism that will allow us to choose the right volume on the next VM startup.
I don't think that locking is enough - I don't see how it solves the issue. It doesn't look like there is a competing process that can write into the source volume. 

Instead of discarding the target volume for unfinalized migrations, we should keep it around.
Then, on the next VM start up we can select the volume with the latest time stamp.

jean-edouard: > Instead of discarding the target volume for unfinalized migrations, we should keep it around.
> Then, on the next VM start up we can select the volume with the latest time stamp.

We can easily change the code to keep target volumes around for migrations that didn't succeed.
However, that timestamp thing doesn't make sense to me. We have no idea what the state of the target PVC is as long as we don't see the migration succeed. It could very well be empty or partially populated. What we can do is keep it around and clean it up next time we migrate the VM? That way it's there for recovery purposes in case the highly unlikely scenario above occurs.

vladikr: > > Instead of discarding the target volume for unfinalized migrations, we should keep it around.
> > Then, on the next VM start up we can select the volume with the latest time stamp.
> 
> We can easily change the code to keep target volumes around for migrations that didn't succeed. 

For migrations that didn't reach the completed status - not for any failed migration.

> However, that timestamp thing doesn't make sense to me. We have no idea what the state of the target PVC is as long as we don't see the migration succeed. It could very well be empty or partially populated. What we can do is keep it around and clean it up next time we migrate the VM? That way it's there for recovery purposes in case the highly unlikely scenario above occurs.

We are only speaking about the scenario when the VM is on the target and is writing to the target PVC when the source is crashed/unavailable. There should be something on the target as the migration has actually been completed. (only the source doesn't know about it and didn't update all other components)

I feel that there is a disconnect, let's take it offline when I'm back.

design-proposals/backend-storage-clone-on-migration/backend-storage-clone-on-migration.md
berrange: >  if a guest is ongoing a post-copy migration and, after the handoff, makes changes to the TPM, then before the end of the migration someone trips on the network cable, the changes made to the TPM will be lost.

That is not quite correct from the QEMU/libvirt POV. In the event of failed post-copy, both the source and dest QEMU will remain in existence, but both in paused state, since neither has the full state in isolation. 

When the network problem is resolved,  you can call virDomainMigrateStartPostCopy with the VIR_MIGRATE_POSTCOPY_RESUME flag, to recover from the problem and complete the migration, without data loss.

design-proposals/backend-storage-clone-on-migration/backend-storage-clone-on-migration.md
vladikr: @alicefr This issue and https://github.com/kubevirt/kubevirt/issues/13112 may affect the already implemented volume migration.

PR #316, opened 2024-08-14
design-proposal: Feature configurables

jcanocan:
This design document states how features that require to have a mechanism to change it's state, e.g., enabled/disabled, should be implemented in KubeVirt.

**Checklist**

**Release note**:

```release-note
Added proposal to introduce how configurable features should be implemented.
```


design-proposals/configurable-features.md
```diff
@@ -0,0 +1,77 @@
...
+
+Before the introduction of [KubeVirt Feature Lifecycle](https://github.com/kubevirt/community/blob/main/design-proposals/feature-lifecycle.md) policy, many feature gates remained after feature's graduation to GA with the sole purpose of acting as a switch for the feature. Generally speaking, this is a bad practice, because feature gates should be reserved for controlling a feature until it reaches maturity. i.e., GA. Therefore, in the case that a developer wants to provide the ability to tune/change the state of the feature, configurables exposed in the KubeVirt CR should be provided. Ideally, this should be accomplished while achieving [eventually consistency](https://en.wikipedia.org/wiki/Eventual_consistency). This forces us to avoid the feature state control checking on webhooks and moving the feature state control closer to the responsible code. Moreover, it has to be decided how the system should behave if a virtual machine instance (VMi) is requiring a feature in a state different from what was configured in the KubeVirt CR, or what should happen if the configuration of a feature in use is changed. (see matrix below).
```
dankenigsberg, line 7:
```suggestion
Before the introduction of [KubeVirt Feature Lifecycle](https://github.com/kubevirt/community/blob/main/design-proposals/feature-lifecycle.md) policy, many feature gates remained after feature's graduation to GA with the sole purpose of acting as a switch for the feature. Generally speaking, this is a bad practice, because feature gates should be reserved for controlling a feature until it reaches maturity. i.e., GA. Therefore, in the case that a developer wants to provide the ability to tune/change the state of the feature, configurables exposed in the KubeVirt CR should be provided. This should be accomplished while achieving [eventually consistency](https://en.wikipedia.org/wiki/Eventual_consistency). This forces us to avoid the feature state control checking on webhooks and moving the feature state control closer to the responsible code. Moreover, it has to be decided how the system should behave if a virtual machine instance (VMi) is requiring a feature in a state different from what was configured in the KubeVirt CR, or what should happen if the configuration of a feature in use is changed. (see matrix below).
```

This is a very important point. Thank you very much for addressing it.

P.s it would be easier to review the text and propose changes if broken to shorter lines.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,77 @@
...
+## Non Goals
+- Describe how features protected with features gates should work.
+
+## Definition of Users
+Development contributors.
+
+Cluster administrators.
+
+## User Stories
+As a developer, I want to make a given feature configurable.
```
dankenigsberg, line 27:
```suggestion
As a VM owner, I want to use a given feature.
```

I don't know if VM owners really want to configure a feature, they just want to use it.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,77 @@
...
+As a cluster administrator, I want to be able to change the cluster wide state of a feature by editing configurables.
+
+
+## Repos
+Kubevirt/Kubevirt
+
+# Design
+If a developer wants to make a feature configurable, he needs to do so by adding new fields to the KubeVirt CR under `spec.configuration`.
+
+> **NOTE:** The inclusion of these new KubeVirt API fields should be carefully considered and justified. The feature configurables should be avoided as much as possible.
```
dankenigsberg, line 38: Correct. There should be very good reasons to complicate our API with new fields.

I think that this proposal would benefit if it include concrete examples of features that really require it. In fact, I'd love to see a comprehensive list of all GAed features that need a cluster-wide configuration.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Goals
+- Establish how the features enablement switch should work.
+- Describe how the system should react in these scenarios:
+    - A feature in KubeVirt is set to state A and a VMi requests the feature to be in state B.
+    - A feature in KubeVirt is set to state A, there are running VMis using the feature in state A, and the feature is changed in KubeVirt to state B.
+    - A feature in KubeVirt is set to state A, and pending VMis want to use it.
+    - A feature in KubeVirt is set to state A, and running VMis using the feature in state B wants to live migrate.
+- Graduate as many features as possible from features gates to configurables.
```
dankenigsberg, line 16: I think we should have a stronger statement in line with https://github.com/kubevirt/community/blob/main/design-proposals/feature-lifecycle.md:
After 3 releases in Beta, all feature gates must be dropped. All features that need configurables must have them by then.

dankenigsberg (changes requested):

design-proposals/configurable-features.md
jcanocan: Thanks for the feedback. I've broken the text in shorter lines. Hope it is easier to review now.

design-proposals/configurable-features.md
jcanocan: I've added the "VM owner" use case. However, IMHO the use case in which the "As a developer, I want..." is valid. This design proposal should provide the guidelines for developers to create feature configurables. WDYT?

dankenigsberg: I'm not sure who is the "developer" persona. Do you refer to a VM owner that want to develop and application inside it?

Can you give an example of a feature that a developer would like to "configure" instead of just "use in their VM".

design-proposals/configurable-features.md
jcanocan: I tried my best to add a list.
I've created two separate list of feature gates: 
- Those that can be tuned in HCO and in the downstream documentation we ask the user to switch them to get a specific feature. These should have a configurable.
- Those that are always enabled in HCO by default, which IMHO we should either, remove the feature gate entirely or create a configurable.
Could you please confirm if the features listed here are GA'd?

design-proposals/configurable-features.md
jcanocan: The intention of that goal is to move GA'd features to configurables or drop them.
In Section Update/Rollback compatibility, we state that before moving to GA, they should drop the feature gate and include a configurable if required. Does it adjust to what are you asking?

design-proposals/configurable-features.md
jcanocan: I meant the KubeVirt developer itself. For example, I'm the owner of the downward metrics feature, and me, as the developer, I want to make this feature configurable, i.e, provide a mechanism to users to enable and disable the feature cluster-wide.

dankenigsberg: The purpose of the "User Stories" section of the proposal is to disassociate us developers a bit. In a user story, you need to put yourself in the shoes of the end user, either the cluster-admin or a VM owner, and think - why would that person want to configure this feature? Isn't it better for them to have it always on?

There may be features that are directed at KubeVirt developers, intended to help testing or debug, but typically they are few. Most of the code we write is for others and we should have them in mind.

dankenigsberg (changes requested):

design-proposals/configurable-features.md
jcanocan: Got it! Thanks for the wide explanation. I've dropped the "developer" user history.

design-proposals/configurable-features.md
jcanocan: I see. I've adjusted the goal and rewrote the Update/Rollback section to be a bit cleaner and to include your concern. Please, take a look.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,155 @@
...
+
+
+> **NOTE:** The inclusion of these new KubeVirt API fields should be carefully considered and justified. The feature
+> configurables should be avoided as much as possible.
+
+
+This is current list of GA'd features present in KubeVirt/KubeVirt which are still using feature gates and are shown as
+configurables in [HCO](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/controllers/operands/kubevirt.go#L166-L174):
+
+- DownwardMetrics
```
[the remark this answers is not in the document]
jcanocan: I've picked up the downward metrics and added an example in section Motivation, since that is the feature I'm more familiar with.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,155 @@
...
+- Root (not sure about this one)
+- DisableMDEVConfiguration
+- PersistentReservation
+- AutoResourceLimitsGate
+- AlignCPUs
+
+This is the current list of GA'd features present in KubeVirt/KubeVirt which are still using feature gates and are always
+enabled by [HCO](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/controllers/operands/kubevirt.go#L125-L142):
```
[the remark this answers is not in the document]
jcanocan: I've assumed that those features are GA since in downstream they are enabled in a hardcoded way, i.e., the user can't disable them by any means. Therefore, IMHO, they should drop the feature flag.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,159 @@
...
+responsible code. Moreover, it has to be decided how the system should behave if a virtual machine instance (VMi) is
+requiring a feature in a state different from what was configured in the KubeVirt CR, or what should happen if the
+configuration of a feature in use is changed. (see matrix below).
+
+## Goals
+
+- Establish how the features status switch should work.
+- Describe how the system should react in these scenarios:
+    - A feature in KubeVirt is set to state A and a VMi requests the feature to be in state B.
+    - A feature in KubeVirt is set to state A, there are running VMis using the feature in state A, and the feature is
```
lyarwood, line 38: s/VMis/VMIs/g

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Interactions with the VMis requests
+
+In case that, the VM exposes a configuration field to request the feature as well as the KubeVirt CRD, the system may
+encounter some inconsistent states that should be handled in the following way:
+
+- If the feature is set to state A in the KubeVirt CR and the VMi is requesting the feature in state B, the VMis must
+  stay in Pending state. The VM status should be updated, showing a status message, highlighting the reason(s) for the
+  Pending state.
+- Feature status checks should only be performed during the scheduling process, not at runtime. Therefore, the feature
```
lyarwood, line 126: Please drop the use of `scheduling` here, this needs to happen during `VMI` reconciliation.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,159 @@
...
+  certificateRotateStrategy: {}
+  configuration:
+    feature-A:  {}
+[...]
+```
+The VM object may or may not include a configuration field inside the relevant spec.
+
+## Interactions with the VMis requests
+
+In case that, the VM exposes a configuration field to request the feature as well as the KubeVirt CRD, the system may
```
lyarwood, line 120: You're interchanging VMs and VMIs quite a bit in this document. Given we need this to work for both VMs and VMIs it might be better to focus on the latter and then later explain how this would then interact with any VM separately.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,155 @@
...
+encounter some inconsistent states that should be handled in the following way:
+
+- If the feature is set to state A in the KubeVirt CR and the VMi is requesting the feature in state B, the VMis must
+  stay in Pending state. The VM status should be updated, showing a status message, highlighting the reason(s) for the
+  Pending state.
+- Feature status checks should only be performed during the scheduling process, not at runtime. Therefore, the feature
+  status changes in the KubeVirt CR should not affect running VMis. Moreover, the VMi should still be able to live
+  migrate, preserving its original feature state.
+- Optionally, It could enable the possibility to reject the KubeVirt CR change request if running VMis are using the
+  feature in a given state. However, by the default the request should be accepted.
```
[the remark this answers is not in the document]
lyarwood: I suggested this use case and agree that as written it can easily race if a creation request from a user comes in while the admin is attempting to disable a given feature.

That said I think that at a high level the use case is still valid and that some cluster admins aren't going to want to disable a feature cluster wide that's being used by running VMs.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,159 @@
...
+policy, features reaching General Availability (GA) need to drop their use of feature gates. This applies also to
+configurable features that we may want to disable.
+
+## Motivation
+
+Users may want certain features to be in a given state, for example to make the best use out of given
+resources or for compliance reasons features may expose sensitive information from the host to the virtual machines (VM)
+or add additional containers to the launcher pod, which are not required by the user. For instance, the downward metrics
+feature is a good example of why some clusters may want to have it enabled or disabled.
+The downward metrics feature exposes some metrics about the host node where the VM is running to the guest. This
```
lyarwood, line 14: nit - can you break this into a separate paragraph for easier review?

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,159 @@
...
+# Implementation Phases
+
+The feature status check should be placed in the VMi reconciliation loop. In this way, the feature status evaluation is
+close to the VMi scheduling process, as well as allowing KubeVirt to reconcile itself if it is out of sync temporally.
+
+Regarding already existing features transitioning from feature gates as a way to set the feature status to configurable
+fields, this change is acceptable, but it should be marked as a breaking change and documented. Moreover, all feature
+gates should be evaluated to determine if they need to be dropped and transitioned to configurables.
+
+## About implementing the checking logic in the VM controller
```
lyarwood, line 155: VMI controller?

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,155 @@
...
+
+In case that, the VM exposes a configuration field to request the feature as well as the KubeVirt CRD, the system may
+encounter some inconsistent states that should be handled in the following way:
+
+- If the feature is set to state A in the KubeVirt CR and the VMi is requesting the feature in state B, the VMis must
+  stay in Pending state. The VM status should be updated, showing a status message, highlighting the reason(s) for the
+  Pending state.
+- Feature status checks should only be performed during the scheduling process, not at runtime. Therefore, the feature
+  status changes in the KubeVirt CR should not affect running VMis. Moreover, the VMi should still be able to live
+  migrate, preserving its original feature state.
```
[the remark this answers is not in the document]
dominikholler: > If a feature is disabled, I don't expect things like new nodes to support it (e.g they would not have needed daemonset running on them) and I would not expect the VMI to migrate to them.

Ack.

> I don't think that we should actively kill them,

Agree.

>  but we should not promise anything about them

> If a feature is disabled, VMs that use it may die.

I do not agree. Which reason do you see that we should not promise the same as we promised at the time the VM was started?

>  I think that we should just alert that VM is using a feature that is no longer supported by the cluster.

Let's check if this is implementable using a reasonable amount of effort at runtime.

design-proposals/configurable-features.md
jcanocan: Done.

design-proposals/configurable-features.md
jcanocan: Done.

design-proposals/configurable-features.md
jcanocan: You are right. Changed all references of VM to VMIs. The only reference to VMs is in the last section, where we open the possibility to implement some logic in the VM controller to get an early feedback about the feature statuses.

design-proposals/configurable-features.md
jcanocan: Done.

design-proposals/configurable-features.md
jcanocan: My intention here is to reflect that we can implement some logic in the VM controller itself to get an early feedback, i.e., before starting the VM, about the feature status in the VM conditions.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,159 @@
...
+us to avoid the feature state control checking on webhooks and moving the feature state control closer to the
+responsible code. Moreover, it has to be decided how the system should behave if a virtual machine instance (VMi) is
+requiring a feature in a state different from what was configured in the KubeVirt CR, or what should happen if the
+configuration of a feature in use is changed. (see matrix below).
+
+## Goals
+
+- Establish how the features status switch should work.
+- Describe how the system should react in these scenarios:
+    - A feature in KubeVirt is set to state A and a VMi requests the feature to be in state B.
```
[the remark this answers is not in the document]
jcanocan: Done.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Definition of Users
+
+Development contributors.
+
+Cluster administrators.
+
+## User Stories
+
+As a cluster administrator, I want to be able to change the cluster wide state of a feature by editing configurables.
```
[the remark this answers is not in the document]
jcanocan: Done.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,159 @@
...
+
+As VM owner, I want to use a given feature.
+
+## Repos
+
+Kubevirt/Kubevirt
+
+# Design
+
+If a developer wants to make a feature configurable, he needs to do so by adding new fields to the KubeVirt CR
```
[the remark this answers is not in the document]
jcanocan: I've reworded a bit the sentence.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,159 @@
...
+## API Examples
+The proposal configuration field, for a given feature in the KubeVirt CR, may look like:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+[...]
+spec:
+  certificateRotateStrategy: {}
+  configuration:
```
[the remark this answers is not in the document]
jcanocan: It would be enabled with:
```yaml
apiVersion: kubevirt.io/v1
kind: KubeVirt
[...]
spec:
  certificateRotateStrategy: {}
  configuration:
    downwardMetrics: {}
[...]
```
And disabled by removing `spec.configuration.downwardMetrics: {}`.

You can find a working PoC in: https://github.com/kubevirt/kubevirt/pull/12650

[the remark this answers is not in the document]
dankenigsberg: What is the benefit of having the `configuration` subelement over
```
apiVersion: kubevirt.io/v1
kind: KubeVirt
[...]
spec:
  certificateRotateStrategy: {}
  downwardMetrics: {}
[...]
``` 
?

And how about an even simpler `downwardMetrics: true` or `downwardMetrics: false` with `false` as the default?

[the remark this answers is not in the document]
tiraboschi: Using an empty struct with the meaning of `true` and its absence with the implicit meaning of `false` sounds a bit odd to me.
I think that in general we should aim to have such kind of configuration values always explicitly represented in our APIs, rather than asserting that "unspecified fields get the default behavior".
Something like `featureXxxx: true/false` will gave us the ability to always explicitly show the expected configuration still letting us choose if we want to default to `true` or `false` via the default property in the openAPIv3 schema, see: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,161 @@
...
+feature. Generally speaking, this is a bad practice, because feature gates should be reserved for controlling a feature
+until it reaches maturity. i.e., GA. Therefore, in the case that a developer wants to provide the ability to tune/change
+the state of the feature, configurables exposed in the KubeVirt CR should be provided. This should be
+accomplished while achieving [eventually consistency](https://en.wikipedia.org/wiki/Eventual_consistency). This forces
+us to avoid the feature state control checking on webhooks and moving the feature state control closer to the
+responsible code. Moreover, it has to be decided how the system should behave if a VMI is
+requiring a feature in a state different from what was configured in the KubeVirt CR, or what should happen if the
+configuration of a feature in use is changed. (see matrix below).
+
+## Goals
```
rmohr, line 35: I think it would be great if the goals could be extended to "I want to have a clear understanding which features are enabled and disabled".

One take which we did lately in another project, was expressing feature gates like this:

Spec:

```
spec:
  featureGates:
    DownwardMetrics: Enabled
    FeatureB: Enabled
    FeatureC: Disabled
```

Status:

```
status:
  featureGates:
    Downwardmetrics: Enabled # explicitly enabled
    FeatureB: Enabled # explicitly enabled
    FeatureD: Enabled # enabled by default
    FeatureC: Disabled # explicitly disabled
    FeatureD: Disabled # disabled by default
```

The status section makes it eventually very clear and discoverable what's **effectively** enabled.

The example here is specficially about feature gates and not about toggling features after they GAed, but I think a simliar approach may make sense to make it visible what's effectively enabled at the end. Especially if we add a second layer of configurables independent of feature gates, it may be even more valuable, since it get's harder to understand what's truly enabled.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
rmohr: Just to throw in other possiblities which may be worth considering, values which transport potentially more meaning are possible as well instead of true/false:

```
spec:
  featureGates:
    DownwardMetrics: Enabled
    FeatureB: Enabled
    FeatureC: Disabled
```

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Interactions with the VMIs requests
+
+In case that, the VMI exposes a configuration field to request the feature as well as the KubeVirt CRD, the system may
+encounter some inconsistent states that should be handled in the following way:
+
+- If the feature is set to state A in the KubeVirt CR and the VMI is requesting the feature in state B, the VMIs must
+  stay in Pending state. The VMI status should be updated, showing a status message, highlighting the reason(s) for the
+  Pending state.
+- Feature status checks should only be performed during the VMI reconciliation process, not at runtime. Therefore, the
```
rmohr, line 128: We should probably go a little bit more into detail here. There can be cases where feature enablement has no API visibility on the VMI and may only happen at virt-handler side, or where virt-operator has to redeploy components in different configurations. If you want to do this in the virt-controller reconciliation stage, some things may need additional hints in the vmi status to kind of snapshot the current configuration. Just to ensure we think about such cases before an agreement is found.

design-proposals/configurable-features.md
rmohr: May be worth (if possible, not entirely sure yet if is possible at all independent of the location, to still consolidate runtime checks on the VMI lievel and propagate from the vmi status to vm status.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,161 @@
+# Overview
+
+With the introduction
+of [KubeVirt Feature Lifecycle](https://github.com/kubevirt/community/blob/main/design-proposals/feature-lifecycle.md)
+policy, features reaching General Availability (GA) need to drop their use of feature gates. This applies also to
+configurable features that we may want to disable.
+
+## Motivation
```
rmohr, line 8: Is the idea of the proposal to outline a general way of handling kubevirt configurables, or is it specifically for VMI features?

rmohr (commented):
Great proposal!

design-proposals/configurable-features.md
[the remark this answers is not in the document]
0xFelix: > And how about an even simpler downwardMetrics: true or downwardMetrics: false with false as the default?

What about features that are not simply turned on or off? Using a struct also leaves it open to add more fields in the future while using a bool makes things pretty awkward in such cases.

design-proposals/configurable-features.md
0xFelix: This proposal targets VM/VMI features respectively every configurable that needs to be taken into account when handling VMs/VMIs.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
jcanocan: Yes. Correct. Thanks for the double check.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
jcanocan: > What is the benefit of having the configuration subelement over

It just feels convenient, IMHO. But there is not much difference in terms of functionally. Moreover, we can also open a new field `spec.featuresConfiguration` if we don't find it either `spec` or `spec.configuration` suitable.

```yaml
[...]
spec:
    featuresConfiguration:
        downwardMetrics: {}
[...]
```


 >   Using a struct also leaves it open to add more fields in the future while using a bool makes things pretty awkward in such cases

>> Would you suggest an example that would be awkward? If one day we find out that we need to control the "color" of downwardMetrics I would consider this quite readable:

Another option is to it in this way:

```yaml
[...]
spec:
    featuresConfiguration:
        downwardMetrics:
            enabled: true/false
            color: purple
[...]
```
It's a bit more wordy, but more readable and organized. WDYT?

design-proposals/configurable-features.md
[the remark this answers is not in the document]
dankenigsberg: > That said I think that at a high level the use case is still valid and that some cluster admins aren't going to want to disable a feature cluster wide that's being used by running VMs.

Right. I think that cluster admins would want to know if the functionality that they are disabling is currently being used. They may also want to know by whom (as not all VMs are created equal). But a mere VM owner should not be in the position to block the cluster admin from expressing an intent that a feature is to be reconfigured.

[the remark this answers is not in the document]
jcanocan: I'm just a bit concerned about the time and resources needed to fetch all VMI using the feature.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,162 @@
...
+
+Development contributors.
+
+Cluster administrators.
+
+## User Stories
+
+* As a cluster administrator, I want to be able to change the cluster-wide state of a feature by editing configurables.
+
+* As VMI owner, I want to use a given feature.
```
iholder101, line 61: I'd add:

* As a VMI owner / cluster admin, I want to understand what's the current configuration of the various features.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,162 @@
...
+In order to make a feature configurable, it must be done by adding new fields to the KubeVirt CR under
+`spec.configuration`.
+
+
+> **NOTE:** The inclusion of these new KubeVirt API fields should be carefully considered and justified. The feature
+> configurables should be avoided as much as possible.
+
+
+This is current list of GA'd features present in KubeVirt/KubeVirt which are still using feature gates and are shown as
+configurables in [HCO](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/controllers/operands/kubevirt.go#L166-L174):
```
iholder101, line 78: While these lists are valuable for discussion's sake, I'm not sure they belong to the proposal itself. WDYT about moving these into the PR description?

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,162 @@
...
+Kubevirt/Kubevirt
+
+# Design
+
+In order to make a feature configurable, it must be done by adding new fields to the KubeVirt CR under
+`spec.configuration`.
+
+
+> **NOTE:** The inclusion of these new KubeVirt API fields should be carefully considered and justified. The feature
+> configurables should be avoided as much as possible.
```
iholder101, line 74: Apart for the list of feature gates this is the only thing written under the "Design" section, although many questions are left unanswered like the questions you've listed under "Goals".

Here I'd expect the document to generally outline the approach we're going to commit to.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
iholder101: I'm with @0xFelix on this one.
Using booleans as APIs is, in general, a smell since it's unextendible, and this is especially true for designing APIs for features configuration.

> Would you suggest an example that would be awkward? If one day we find out that we need to control the "color" of downwardMetrics I would consider this quite readable:

```yaml
apiVersion: kubevirt.io/v1
kind: KubeVirt
[...]
spec:
  certificateRotateStrategy: {}
  downwardMetrics: true
  downwardMetricsColor: purple
[...]
```

@dankenigsberg To me this is much less readable, but that's pretty fine when you have only a single unstructured configuration. Please consider the following example (which is, obviously, entirely made up):

```yaml
kind: KubeVirt
[...]
spec:
  downwardMetrics:
    hostInformation:
      - MACAdress
      - IP
      - ListOfUsers
    style:
      color: White
      textSize: 17
[...]
```

I think you can agree that it looks much better than (the order is messed up on purpose. bear in mind that the order of yaml fields is usually sorted in an alphabetical order):
```yaml
kind: KubeVirt
[...]
spec:
  downwardMetrics: true
  downwardMetricsColor: purple
  downwardMetricsIncludeHostIP: true
  downwardMetricsTextSize: 17
  downwardMetricsIncludeHostMacAddress: true
  downwardMetricsIncludeHostListOfUsers: true
[...]
```

My point is: this is unscalable and will fastly turn into a complete mess. We should take advantage of the fact that YAML/JSON already supports structured data.

In addition, it's common in k8s to `omitempty` and leave only whatever is relevant / enabled, so I'm not sure how it's odd.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,162 @@
...
+The VMI object may or may not include a configuration field inside the relevant spec.
+
+## Interactions with the VMIs requests
+
+In case that, the VMI exposes a configuration field to request the feature as well as the KubeVirt CRD, the system may
+encounter some inconsistent states that should be handled in the following way:
+
+- If the feature is set to state A in the KubeVirt CR and the VMI is requesting the feature in state B, the VMIs must
+  stay in Pending state. The VMI status should be updated, showing a status message, highlighting the reason(s) for the
+  Pending state.
```
iholder101, line 128: Can you elaborate on this one / give an example?

design-proposals/configurable-features.md
iholder101: +1

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,162 @@
...
+responsible code. Moreover, it has to be decided how the system should behave if a VMI is
+requiring a feature in a state different from what was configured in the KubeVirt CR, or what should happen if the
+configuration of a feature in use is changed. (see matrix below).
+
+## Goals
+
+- Get a clear understanding about the feature status.
+- Establish how the features status swapping should work.
+- Describe how the system should react in these scenarios:
+    - A feature in KubeVirt is set to state A and a VMI requests the feature to be in state B.
```
iholder101, line 40: You assume that all of the features are granular to the VM/VMI level, but I don't think it's necessarily true. Things like `UseEmulation`, `MinimumClusterTSCFrequency`, `ImageRegistry` and so on are cluster-wide by nature that don't even have APIs at the VM/VMI level.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,162 @@
...
+## Interactions with the VMIs requests
+
+In case that, the VMI exposes a configuration field to request the feature as well as the KubeVirt CRD, the system may
+encounter some inconsistent states that should be handled in the following way:
+
+- If the feature is set to state A in the KubeVirt CR and the VMI is requesting the feature in state B, the VMIs must
+  stay in Pending state. The VMI status should be updated, showing a status message, highlighting the reason(s) for the
+  Pending state.
+- Feature status checks should only be performed during the VMI reconciliation process, not at runtime. Therefore, the
+  feature status changes in the KubeVirt CR should not affect running VMIs. Moreover, the VMI should still be able to
```
iholder101, line 130: What about things like `EvictionStrategy`?
Are we sure that we don't want to ever impact running VMs? To me it sounds valuable from an admin's perspective to change the eviction strategy before an upgrade for example.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
iholder101: > > That said I think that at a high level the use case is still valid and that some cluster admins aren't going to want to disable a feature cluster wide that's being used by running VMs.
> 
> Right. I think that cluster admins would want to know if the functionality that they are disabling is currently being used. They may also want to know by whom (as not all VMs are created equal). But a mere VM owner should not be in the position to block the cluster admin from expressing an intent that a feature is to be reconfigured.

+1
I'm not sure we can make a simplistic rule here, it sounds possible that in some cases it's useful to change configuration for running VMs and in others it isn't. Perhaps it's useful to start off with more concrete examples and have a discussion regarding them.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,162 @@
...
+    - A feature in KubeVirt is set to state A and a VMI requests the feature to be in state B.
+    - A feature in KubeVirt is set to state A, there are running VMIs using the feature in state A, and the feature is 
+      changed in KubeVirt to state B.
+    - A feature in KubeVirt is set to state A, and pending VMIs want to use it.
+    - A feature in KubeVirt is set to state A, and running VMIs using the feature in state B wants to live migrate.
+- Graduate features status swapping from features gates to configurables.
+
+## Non Goals
+
+- Describe how features protected with features gates should work.
```
iholder101, line 49: I would add:

- Change how feature gates are managed. Feature gating and configuration are two completely distinct issues.

iholder101 (commented):
Thank you for this proposal, this is a very important topic. Happy to see this discussion!

design-proposals/configurable-features.md
jcanocan: Maybe we can allow impacting running VMs if the change does not require rebooting the VM. WDYT?

design-proposals/configurable-features.md
jcanocan: Included.

design-proposals/configurable-features.md
jcanocan: I've extended the section to include more details and to make the Goal Section a bit easier to follow. Hope it is better now.

design-proposals/configurable-features.md
jcanocan: I agree. Done.

design-proposals/configurable-features.md
jcanocan: Added.

design-proposals/configurable-features.md
jcanocan: Regarding the status: I've included this in the goals and an example of how it might look like.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
jcanocan: I tend to agree with Felix and Itamar. However, do you think these kinds of configurations should be under `spec.configuration` :
```yaml
spec:
  configuration:
    downwardMetrics: {}
```
or just under `spec`:
```yaml
spec:
  downwardMetrics: {}
``` 
?

design-proposals/configurable-features.md
dankenigsberg: > Could you please confirm if the features listed here are GA'd?

I cannot, but we must know the answer, even if it means reaching out to every developer that introduced each feature gate.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,205 @@
...
+accomplished while achieving [eventually consistency](https://en.wikipedia.org/wiki/Eventual_consistency). This forces
+us to avoid the feature state control checking on webhooks and moving the feature state control closer to the
+responsible code. Moreover, it has to be decided how the system should behave if a VMI is
+requiring a feature in a state different from what was configured in the KubeVirt CR, or what should happen if the
+configuration of a feature in use is changed. (see matrix below).
+
+## Goals
+
+- Get a clear understanding about the features status.
+- Establish how the features status swapping should work.
```
dankenigsberg, line 38: I find the terminology here confusing. Is "status" the request of the cluster admin that can be "swapped"? Typically it is the actual condition reported by the cluster. Maybe you can explain the goal here in terms of a user story? What does the cluster admin want to do/understand?

design-proposals/configurable-features.md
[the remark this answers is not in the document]
dankenigsberg: > I'm just a bit concerned about the time and resources needed to fetch all VMI using the feature.

Would you elaborate, possibly with a specific example? Would it be much harder than `kubectl get vm -o yaml|grep someRequestedFeature`?

+1 for considering several concrete examples.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
dankenigsberg: Are empty dictionaries signifying "true" common in other k8s resources? I find them unnatural. When a feature is defined, we need to think about all the way it can be configured. The API of a feature should be the first thing to be discussed. Most features should have nothing at all - they should be enabled on all clusters. I think it would be rare for us to realize after the fact that an existing feature needs a config session. In such a rare condition we can add a new section such as
```
spec:
  downwardMetrics: enabled
  downwardMetricsOptions:
     - opt1: 7
     - opt2: yes
```

I don't understand the motivation for adding the `configuration` element, and therefor prefer not having it.

Also, I don't think that we need a `features` section or a `feature` prefix. I expect each feature to be unique and unrelated to other.

design-proposals/configurable-features.md
dankenigsberg: I find the concrete list very helpful to understand what this proposal is about. I would prefer to understand what we plan for each of them.

design-proposals/configurable-features.md
dankenigsberg: Why? I think that we should eliminate all feature gates.

dankenigsberg (changes requested):

design-proposals/configurable-features.md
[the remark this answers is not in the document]
iholder101: > Are empty dictionaries signifying "true" common in other k8s resources? 

Here's an example:
```yaml
kind: Pod
spec:
  containers:
  - name: example-container
    volumeMounts:
    - mountPath: /data
      name: my-volume
  volumes:
  - name: my-volume
    emptyDir: {}
```

The empty struct here means "use an emptyDir volume source". It's possible to configure the emptyDir as follows:
```yaml
kind: Pod
spec:
...
  volumes:
  - name: my-volume
    emptyDir:
      medium: Memory
      size: 1Gi
```

> I think it would be rare for us to realize after the fact that an existing feature needs a config session.

I respectfully disagree.
A feature has a life of its own - it evolves in ways that are often impossible to foresee in advance.

As I see it, the most important part of designing APIs is not to outline them as best for **today**, but to keep them open for extensions in **the future**. There are many software engineering principles revolving around that idea, like [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), and the idea of incremental changes.

> In such a rare condition we can add a new section such as
```
spec:
  downwardMetrics: enabled
  downwardMetricsOptions:
     - opt1: 7
     - opt2: yes
```

Apart from the fact that I don't think this is very elegant:
* What happens if `downwardMetrics` is disabled but the options are still provided? That's both confusing and verbose for no reason.
* If `downwardMetrics` is enabled, than the first field here is redundant. By providing options that configure downwardMetrics it already obvious it's enabled.

But the worst part is that it's very often that a field that starts off as a boolean suddenly demands to support a third value. A classic example is the VM's `running: true/false` spec field. At the beginning we thought that true/false is enough, then we needed to deprecate this field (which is still there and is hard to get rid of!) and replace it with a RunStrategy field.

IMHO **using booleans is a code smell** and in most cases using a two-valued enum or an empty struct is better and opens the door for future extensions.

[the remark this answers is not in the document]
iholder101: > I don't understand the motivation for adding the configuration element, and therefor prefer not having it.

I agree. IMO spec is another word for "configuration", they basically mean the same thing.
However I'm not sure how easy it will be to get rid of it in terms of backward compatibility.

design-proposals/configurable-features.md
jcanocan: Perfect. I agree, maybe we can create an issue for each feature gate and ping the last 2 o 3 contributors of them.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
jcanocan: No, it won't be much harder. However, from my limited point of view, that would imply to fetch each VM in the system and check if the feature is present. That would result in an algorithmic complexity O(n). I'm not sure if such complexity is acceptable. Maybe there is a fancier way to do this that I'm not aware of.

Regarding of changing configurations in running VMs: I've proposed to allow changing configurations that do not require a VM restart to take effect, e.g., `EvictionStrategy`.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
jcanocan: Looks like there is an agreement on no placing new configurables under `spec.configuration` but in `spec`. Adjusted the document to reflect this.

design-proposals/configurable-features.md
jcanocan: Okay. I've reintroduced the FG list.

design-proposals/configurable-features.md
jcanocan: > Why? I think that we should eliminate all feature gates.

Sorry, I'm not sure what do you mean. Could you please elaborate?

design-proposals/configurable-features.md
jcanocan: I've adjusted the goal. The idea is to reflect how the features can be tuned, e.g., enabled or disabled.

design-proposals/configurable-features.md
dankenigsberg: Makes sense. I think that contacting them should be part of this proposal, as they are the stakeholder who would need to do the work prescribed here.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
dankenigsberg: This complexity is paid by the human who is responsible to make the cluster-wide changes. I don't think there's any way around it. The cluster-admin must be familiar with the workload and whether the feature he or she tweaks is needed now or in the near future.

>  I've proposed to allow changing configurations that do not require a VM restart to take effect, e.g., EvictionStrategy

Do you suggest to disallow modification of the kubevirt cr? Isn't it against https://github.com/kubevirt/community/pull/316/files#diff-48d9142aa505032f3643eef6ca3fb52e36bf8e84608a9649565619133f07490dR61 ?

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,233 @@
...
+> configurables should be avoided as much as possible.
+
+Moreover, the KubeVir CR `status` field should clearly indicate the current state of each feature, providing a
+comprehensive overview of the operational status of these features:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+[...]
+status:
```
dankenigsberg, line 103: this is too abstract for me. How would this look like for existing features?

design-proposals/configurable-features.md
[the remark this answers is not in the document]
dankenigsberg: @jcanocan can you follow up on empty-map-means-true discussion?

design-proposals/configurable-features.md
dankenigsberg: Thanks. I'd like to understand how the proposal is going to affect many/all of them.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,233 @@
...
+## Motivation
+
+Users may want certain features to be in a given state, for example to make the best use out of given
+resources or for compliance reasons features may expose sensitive information from the host to the virtual machines
+instances (VMI) or add additional containers to the launcher pod, which are not required by the user. 
+
+The downward metrics feature is a good example of why some clusters may want to have it enabled or disabled.
+The downward metrics feature exposes some metrics about the host node where the VMI is running to the guest. This
+information may be considered sensitive information.
+If there is no mechanism to disable the feature any VMI could request the metrics and inspect information that, in some
```
dankenigsberg, line 17:
```suggestion
If there is no mechanism to disable the feature, any VMI could request the metrics and inspect information that, in some
```

design-proposals/configurable-features.md
dankenigsberg: > Sorry, I'm not sure what do you mean. Could you please elaborate?

We have many feature gates that do not comply with https://github.com/kubevirt/community/blob/main/design-proposals/feature-lifecycle.md . Some of them expose a VM/VMI API, some does not. But imo that's not the main issue. All feature gates must comply. They all must either graduate or be reverted, and those that require a configurable in kubevirt cr must define it - not only those with VM/VMI APIs.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
iholder101: > > The empty struct here means "use an emptyDir volume source"
> 
> Sorry if my question was unclear. I'm asking if in the history of `emptyDir` it was ever defined as an empty dictionary with no field at all?

Hey @dankenigsberg!
Sorry for not being clear.

Yes indeed. As can be seen [here](https://github.com/kubernetes/kubernetes/pull/5166/files#diff-dfee74d02fd350fb1ee1ffbbbc32acf4cb075c9ae42b6e31c8c7acefe7fed623L190), emptyDir used to be an empty struct. Only later in time the `Medium` field was added, then later the `Size` field was added as well.

design-proposals/configurable-features.md
dankenigsberg: I don't like to give this promise.

Rebooting VMs is not fun, we should try to avoid it. But if as a cluster-admin, I no longer want `downwardMetrics`, I would like to be able to express this - even if eventually this means that all VMs must be restarted.

Typically, any change to kubevirt cr is going to affect how currently-running or future VMs are going to behave. We can warn the admin, but not block them.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
iholder101: > Looks like there is an agreement on no placing new configurables under `spec.configuration` but in `spec`. Adjusted the document to reflect this.

In theory I support that.
But as said above, we already have fields living under configuration and we need to mind backward compatibility.

Here are some options that come into mind:
* Duplicate configuration fields into `.spec`, deprecate the ones under `.spec.configuration`, and basically duplicate them until we advance Kubevirt CR to v2.
* Document that in the future we want to move these fields, but keep them as-is for now.

IMHO, you might consider splitting these two efforts to two distinct PRs: this PR to determine the feature toggles policy, and a different PR to plan deprecating `.spec.configuration`.

design-proposals/configurable-features.md
iholder101: > Could you please confirm if the features listed here are GA'd?

FYI @stu-gott @acardace, perhaps we can collaborate to list the different FGs and their graduation state.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
jcanocan: I agree. My point is: new feature configurable specs like `downwardMetrics`, should be placed in `spec` in the Kubevirt CR, e.g., 

```yaml
spec:
  downwardMetrics: {}
```
Regarding existing specs under `.spec.configuration`, do you think we should reflect this somehow in this design document?

> @jcanocan can you follow up on empty-map-means-true discussion?

Sure. IMHO, arguments provided by @iholder101 and @0xFelix sounds good enough.

design-proposals/configurable-features.md
jcanocan: I see. I've included the goal "-Graduate features status swapping from features gates to configurables." with that intention. Maybe it's not clear enough. WDYT of changing it to "Drop feature gates that do not require a configurable and graduate feature gates to feature configurables that requires them."?

design-proposals/configurable-features.md
jcanocan: Should we allow live migration in the first place if, for instance, the feature is disabled and the VMI is requesting it?

design-proposals/configurable-features.md
jcanocan: Maybe "state" would be more accurate?

design-proposals/configurable-features.md
jcanocan: For instance, let's suppose that we have `downwardMetrics` and `maxAllowedCPUsPerVM` feature. The `maxAllowedCPUsPerVM` controls the maximum amount of CPUs that a given VM can get. We enable the `downwardMetrics` and we want to restrict that any given VM just can get 2 CPUs. The Kubevirt CR will looks like:

```yaml
apiVersion: kubevirt.io/v1
kind: KubeVirt
[...]
spec:
  downwardMetrics: {} # this means "enabled"
  maxCPUsPerVM: 2
[...]
status:
  featureStatus:
    downwardMetrics:
      status: Enabled
    maxCPUsPerVM:
      status: 2 CPUs
```

WDYT?

design-proposals/configurable-features.md
jcanocan: Done.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
iholder101: > new feature configurable specs like downwardMetrics, should be placed in spec in the Kubevirt CR, e.g.,

> Regarding existing specs under .spec.configuration, do you think we should reflect this somehow in this design document?

IMO it would be easier to discuss this in a different proposal

design-proposals/configurable-features.md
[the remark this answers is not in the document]
jcanocan: > This complexity is paid by the human who is responsible to make the cluster-wide changes. I don't think there's any way around it. The cluster-admin must be familiar with the workload and whether the feature he or she tweaks is needed now or in the near future.

Therefore, in your opinion, KubeVirt should warn the admin if there's affected VMs by the change, right? But in any means, we shouldn't block the update.

> Do you suggest to disallow modification of the kubevirt cr? Isn't it against https://github.com/kubevirt/community/pull/316/files#diff-48d9142aa505032f3643eef6ca3fb52e36bf8e84608a9649565619133f07490dR61 ?

No. At least by default, any modification should be allowed in the kubevirt CR.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
jcanocan: All right. Thanks for the clarification.

design-proposals/configurable-features.md
dankenigsberg: I would avoid the term "status swapping", we are talking here about configuring or specifying how a feature behave, it is not about the current status or about swapping two things. Howe about
`Graduate features by dropping their gates and (optionally) adding `spec` options for them`

design-proposals/configurable-features.md
dankenigsberg: > Should we allow live migration in the first place if, for instance, the feature is disabled and the VMI is requesting it?
I would expect migration to fail if the destination host does not support the requested feature. Otherwise, we should not block migration (e.g to a host that still has this feature for some reason)

design-proposals/configurable-features.md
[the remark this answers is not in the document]
dankenigsberg: > Therefore, in your opinion, KubeVirt should warn the admin if there's affected VMs by the change, right? But in any means, we shouldn't block the update

Correct. And in "warning" I means "raise and alert or event or condition" on any affected VMs.

> No. At least by default, any modification should be allowed in the kubevirt CR.

Good. But we should not have a non-default attempt to block modification of the spec. It cannot be done safely (it is raceful), so we should not try.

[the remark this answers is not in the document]
jcanocan: > Correct. And in "warning" I means "raise and alert or event or condition" on any affected VMs.

Ah I see, so should be the VMs itself who triggers the alert, event, warning, etc. Did I understand correctly?

> Good. But we should not have a non-default attempt to block modification of the spec. It cannot be done safely (it is raceful), so we should not try.

Works for me.

design-proposals/configurable-features.md
jcanocan: Sounds good! 
All right, I will replace any occurrence of "status swapping".
Many thanks for the suggestion. Much appreciated.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
dankenigsberg: > should be the VMs itself who triggers the alert, event, warning, etc.

As a VM owner, I see value in knowing that my VM is using a feature that is being removed from the cluster. So yes, a VM condition would be useful.

design-proposals/configurable-features.md
jcanocan: No, it shouldn't differ. This idea was proposed here: https://github.com/kubevirt/community/pull/316#discussion_r1727443301

design-proposals/configurable-features.md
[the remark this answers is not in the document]
jcanocan: All right. Would that also work for the cluster admin changing configurations? I.e., as a cluster admin, I'm changing configurables and want to know if there's any affected VM? or should be the cluster admin careful enough to run a command the one you have posted before `kubectl get vm -A -o yaml|grep someRequestedFeature` to know it in advance?

design-proposals/configurable-features.md
jcanocan: All right. I will drop it.

design-proposals/configurable-features.md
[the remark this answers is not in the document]
dankenigsberg: > should be the cluster admin careful enough to run a ...  `kubectl get vm -A -o yaml|grep someRequestedFeature` to know it in advance?

I don't see any other way. Note that this is also raceful. Someone may create a VM with `someRequestedFeature` a microsecond later. That's the nature of our platform - the cluster admin cannot "know" at any point. They can only declare what is desired, and eventually the cluster would get there.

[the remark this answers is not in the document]
jcanocan: All right. Thanks for the input. I will drop the sentence or change it by something like "the kubevirt CR can't be rejected, the cluster admin is responsible to know in advance if the feature configurable change will affect running VMIs".

design-proposals/configurable-features.md
jcanocan: Understood. I've updated the text, trying to better reflect what we have discussed here.

design-proposals/configurable-features.md
jcanocan: Updated. Just kept "state" in those situations where I believe is clear what are we referring to. Please, take a look and if it is still confusing, I will change it.

design-proposals/configurable-features.md
jcanocan: Dropped.

jcanocan:
Hi @sradco!

We are discussing the possibility of triggering an alert in case that a given feature is configured differently in a given VM from cluster-wide configuration. Let me provide an example:

I, as a sysadmin, enable the downward metrics feature and some users create a VMI using that feature. At a given point, I decide to disable the downward metrics feature cluster-wide. In such case, we agreed to no intentionally force VMIs to reboot, kill, stop, etc. but rather let them continue running but with no guarantees that the VM won't face issues in the future that makes it to stop, reboot, etc. 

In order to inform the user, we agreed that the VMI should update its status and log a warning to reflect this situation.

The question is, do you think that the VMI should also trigger an alert? If yes, it this technically possible?

Thanks!

sradco:
Hi,

I would address this with a metric, not sure an alert is needed.
We try to keep the alerts for when they are truely needed and actionable.

We can show metrics loke this in the dashboards(I'm in the process of building them for ACM and they will also be available in-cluster)

and I think we can also add this information to the VM page and show it as a warning.

Wdyt?

jcanocan:
> Hi,
> 
> I would address this with a metric, not sure an alert is needed. We try to keep the alerts for when they are truely needed and actionable.
> 
> We can show metrics like this in the dashboards(I'm in the process of building them for ACM and they will also be available in-cluster)
> 
> and I think we can also add this information to the VM page and show it as a warning.
> 
> We can add to kubevirt_vmi_info metric or to a new kubevirt_vmi_labels metric, the features as labels and value enabled/disables.
> 
> Wdyt?

Yes, it sounds amazing!
Sorry, I'm not sure which option would be more convenient. Could you please elaborate the ups and downs of each option?
Thanks for the input. Much appreciated

sradco:
On Wed, Sep 11, 2024, 5:37 PM Javier Cano Cano ***@***.***>
wrote:

> Hi,
>
> I would address this with a metric, not sure an alert is needed. We try to
> keep the alerts for when they are truely needed and actionable.
>
> We can show metrics like this in the dashboards(I'm in the process of
> building them for ACM and they will also be available in-cluster)
>
> and I think we can also add this information to the VM page and show it as
> a warning.
>
> We can add to kubevirt_vmi_info metric or to a new kubevirt_vmi_labels
> metric, the features as labels and value enabled/disables.
>
> Wdyt?
>
> Yes, it sounds amazing!
> Sorry, I'm not sure which option would be more convenient. Could you
> please elaborate the ups and downs of each option?
> Thanks for the input. Much appreciated
>

I would go with the second option, since the kubevirt_vmi_info hold basic
information about the vmi and adding these labels would be too much.
Also, it should serve a specific target, so I would even call it something
like kubevirt_vmi_features_status_info to make it clear what its about.

It should be easy to query with this approach all the vmis that have a
specific feature enable/disable.

Best,
Shirly

0xFelix:
I think the problem we have with this approach is that we don't have a list of active/disabled features on a VM? In case of features which are configured in the VM's spec it is easier to detect. But what about features that do not expose configurables on the VM level?

0xFelix:
Still, how do you know configurations are inconsistent? It comes back to the initial issue of not knowing which feature is used by which VM.

jcanocan:
In order to adjust the scope of this design proposal. The alerting concern has been dropped, as well as the detection to configurables checks that do not expose VMI API fields. Those concerns will addressed in a follow-up design proposal.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,229 @@
...
+
+* As a VMI owner / cluster admin, I want to understand what's the current configuration of the various features.
+
+## Repos
+
+Kubevirt/Kubevirt
+
+# Design
+
+The feature configuration must be done by adding new fields to the KubeVirt CR under `spec`:
```
dankenigsberg, line 73: I'd start by stating that
However, some feature may be harmful for some clusters and 

```suggestion
Ideally, a graduated feature would just works out the box, with no further complexity to the cluster admin. 
Features that must be configured must add new fields to the KubeVirt CR under `spec`:
```

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,229 @@
...
+
+Regarding already existing features transitioning from feature gates as a way to enable/disable a feature to configurable
+fields, this change is acceptable, but it should be marked as a breaking change and documented. Moreover, all feature
+gates should be evaluated to determine if they need to be dropped and transitioned to configurables.
+
+## About implementing the checking logic in the VM controller
+
+The checking in the VM controller could be added to let the user know if a VM has requested a feature configuration which 
+is different from what it is specified in the KubeVirt CR. This will provide an early feedback to the user before starting
+the VM. However, it should not prevent the user to start the VM, the VMI controller should take care of checks preventing
```
dankenigsberg, line 228: I don't understand the English here, e.g what is the "early feedback" you refer to.

Can you rephrase the paragraph to ensure that
* The controller must not start a VM that requires a feature that is not available in the cluster.
* The fact that a VM cannot start despite its owner asking to do so must be reported by the controller in the `status` field of the VM.

dankenigsberg (changes requested):

design-proposals/configurable-features.md
jcanocan: Done.

design-proposals/configurable-features.md
jcanocan: Done.
Let me clarify what I wanted to express with "early feedback". Let's suppose that the `feature-b` is disabled cluster-wide and the user creates a VM object requesting this feature. The VM is not requested to start, just the VM object is created. The idea is that the VM controller updates the `status` field reflecting that this VM won't start because it is requesting a feature no available.
I'm not sure if this would be helpful for users or just will create more noise.

dankenigsberg: Ah. You would like to inform a stopped VM that it won't be able to generate a VMI. This can be quite expensive (as it requires you to track all stopped VMs for no immediate reason). I would not make this a requirement.

dankenigsberg (approved):

design-proposals/configurable-features.md
jcanocan: AFAIK, it is not that expensive. It's a check in the VM controller reconciliation loop, what it is true is that if you have a stopped VM and disable the feature, the reconciliation loop is not triggered immediately. Takes some time. I will make this optional.

jcanocan: Done. Could you please add the lgtm again if you are fine with the change? Thanks.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,232 @@
...
+- Updates to KubeVirt CR to update a feature configuration should not be rejected.
+
+## Scalability
+
+The feature configurables should not affect in a meaningful way the cluster resource usage.
+
+## Update/Rollback Compatibility
+
+The feature configurables should not affect forward or backward compatibility once the feature GA. A given feature,
+after 3 releases in Beta, all feature gates must be dropped. Those features that need a configurable should define it ahead
```
0xFelix, line 208: > A given feature, after 3 releases in Beta, all feature gates must be dropped.

Do we want to mention this here? I was under the impression this proposal is not about features gates?

0xFelix (commented):
Do I understand correctly that if a VM is unable to start, because it is requesting a non-valid configuration, a VMI that remains in pending state is still created? Wouldn't it be better to not create a VMI if we know early that it won't be able to start?

design-proposals/configurable-features.md
dankenigsberg: > > A given feature, after 3 releases in Beta, all feature gates must be dropped.
> 
> Do we want to mention this here? I was under the impression this proposal is not about features gates?

I think it is important to mention. Some old features have been using gates to configure if they are off or on. This has to change - features that cannot be always on, must add configurable so that we can graduate the feature and drop the feature gate.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,232 @@
...
+
+Users may want certain features to be configurable, for example to make the best use out of given
+resources or for compliance reasons features may expose sensitive information from the host to the virtual machines
+instances (VMI) or add additional containers to the launcher pod, which are not required by the user. 
+
+The downward metrics feature is a good example of why some clusters may want to have it enabled or disabled.
+The downward metrics feature exposes some metrics about the host node where the VMI is running to the guest. This
+information may be considered sensitive information.
+If there is no mechanism to disable the feature, any VMI could request the metrics and inspect information that, in some
+cases, the admin would like to hide, creating a potential security issue.
```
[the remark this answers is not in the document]
jcanocan: Added this.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,232 @@
...
+[...]
+spec:
+  certificateRotateStrategy: {}
+  feature-A:  {}
+  feature-C:
+    configA: integer
+    configB: string
+[...]
+```
+Please note that if the feature spec field is not present, the feature is assumed to be completely disabled.
```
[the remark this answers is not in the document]
jcanocan: Dropped. IMO, this could be left up to the developers to decide.

jcanocan:
> Do I understand correctly that if a VM is unable to start, because it is requesting a non-valid configuration, a VMI that remains in pending state is still created? Wouldn't it be better to not create a VMI if we know early that it won't be able to start?

No, it should not create the VMI object. Only if the VMI object is created directly, it will remain in `Pending` state. I've added a clarification sentence in the "About implementing the checking logic in the VM controller" Section.

design-proposals/configurable-features.md
0xFelix: ACK

0xFelix (approved):
/lgtm

Thanks! Time to get this merged.

dankenigsberg:
> > Do I understand correctly that if a VM is unable to start, because it is requesting a non-valid configuration, a VMI that remains in pending state is still created? Wouldn't it be better to not create a VMI if we know early that it won't be able to start?
> 
> No, it should not create the VMI object. Only if the VMI object is created directly, it will remain in `Pending` state. I've added a clarification sentence in the "About implementing the checking logic in the VM controller" Section.

In one sense it is better to fail the creation of the VMI (less redundant objects in the system); But in another sense it is worse, because this would force the VM controller to replicate logic that is already in the VMI controller. I prefer the cleaner design, with a clear separation of responsibilities.

jcanocan:
> > > Do I understand correctly that if a VM is unable to start, because it is requesting a non-valid configuration, a VMI that remains in pending state is still created? Wouldn't it be better to not create a VMI if we know early that it won't be able to start?
> > 
> > 
> > No, it should not create the VMI object. Only if the VMI object is created directly, it will remain in `Pending` state. I've added a clarification sentence in the "About implementing the checking logic in the VM controller" Section.
> 
> In one sense it is better to fail the creation of the VMI (less redundant objects in the system); But in another sense it is worse, because this would force the VM controller to replicate logic that is already in the VMI controller. I prefer the cleaner design, with a clear separation of responsibilities.

Could you please expand how the "clear separation of responsibilities" would look like?

dankenigsberg:
> Could you please expand how the "clear separation of responsibilities" would look like?

The sync look for the VMI controller is responsible to check if the VMI can start on the cluster. It must have this code for people defining their VMI without a VM or changing cluster config while starting a VM.

Making another component (such as the VM controller) know about this logic mixes the responsibilities. Architecturally speaking, it is better for each component to [do one thing and do it well](https://en.wikipedia.org/wiki/Unix_philosophy#Origin).

So let us have the VM controller create the VMI object, which would float aimlessly until the VM is stopped or the feature is enabled. The VM controller should still bubble up the reason for the failure in the `VM.status` field.

design-proposals/configurable-features.md
```diff
@@ -0,0 +1,230 @@
...
+temporally.
+
+Regarding already existing features transitioning from feature gates as a way to enable/disable a feature to configurable
+fields, this change is acceptable, but it should be marked as a breaking change and documented. Moreover, all feature
+gates should be evaluated to determine if they need to be dropped and transitioned to configurables.
+
+## About implementing the checking logic in the VM controller
+
+The VM controller should not allow starting a VM if it is requesting a feature that it is not available in the cluster.
+Moreover, the VMI object must not be created. The controller must report the reasons in the `status` field of the VM.
```
dankenigsberg, line 226:
```suggestion
KubeVirt should not allow starting a VM if it is requesting a feature that it is not available in the cluster.
The VM controller must report the reasons in the `status` field of the VM.
```

Let us consolidate feature testing in the VMI controller.

dankenigsberg (changes requested):

design-proposals/configurable-features.md
jcanocan: Adjusted.

jcanocan:
> > Could you please expand how the "clear separation of responsibilities" would look like?
> 
> The sync look for the VMI controller is responsible to check if the VMI can start on the cluster. It must have this code for people defining their VMI without a VM or changing cluster config while starting a VM.
> 
> Making another component (such as the VM controller) know about this logic mixes the responsibilities. Architecturally speaking, it is better for each component to [do one thing and do it well](https://en.wikipedia.org/wiki/Unix_philosophy#Origin).
> 
> So let us have the VM controller create the VMI object, which would float aimlessly until the VM is stopped or the feature is enabled. The VM controller should still bubble up the reason for the failure in the `VM.status` field.

Got it. Thanks for the clarifications.

dankenigsberg (approved):

aburdenthehand:
@iholder101 - can you please take a look and sign off on behalf of sig-compute.
If this is good to go then I can merge it.

iholder101:
@jcanocan thank you for this proposal and your dedication on this very important subject!

As per the goals you've listed, there are two main subjects discussed in this proposal:
1. The need to convert GAed feature gates to configurable.
2. Define a contract in which we make promises around how workloads should behave when the configuration changes.

I think that the first subject above is less controversial. I also really like how you've handled it, mainly the Design section in your proposal.

However, the second subject is much more sensitive, and I feel there are many questions left unanswered, for example:
* How can we know which features a VM is using? Some features don't have vm/vmi level configs. We theoretically can list the used features as part of a VM's status, but this only raises more questions to me, such as what is exactly considered a feature and which information exactly we need to store.
* How running workloads should be affected by config changes? Because a "feature" is such a vague term I think it's hard to come up with a single-for-all policy, as it might be sensible for some features to behave very differently from others.

As [expressed above](https://github.com/kubevirt/community/pull/316#discussion_r1741590238), I'm not sure the policy presented in this proposal is mature enough, or if we will ever be able to tackle this globally rather than on a case-to-case basis.

What I suggest is to remove the second subject from this proposal, keep the Design section and merge the less controversial part first. We can currently say that changing configs is done on a best-effort basis and that it's up to the admin not to mess stuff around. Then we can start discussing how a VM should report which features it is using, and how to perform validation when features are changed in a way that provides enough flexibility for features to differ from one another.

Alternatively, we could dive deeper into the sensitive part of your proposal and continue discussing it here if you prefer.

Thanks again for your efforts!

PR #325, opened 2024-09-24
design-proposal: Primary Network Binding for seamless migration

EdDev:

Choose the most appropriate Kubevirt network binding for the primary pod network, considering basic functionality support, including seamless migration.

Fixes #

**Checklist**

**Release note**:
```release-note
NONE
```

### DIFF
--- a/design-proposals/primary_network_binding_for_seamless_migration.md
+++ b/design-proposals/primary_network_binding_for_seamless_migration.md
@@ -1,0 +1,597 @@
+# Primary Pod Network Binding for Seamless Migration
+
+## Summary
+
+Choose the most appropriate Kubevirt network binding for the primary pod
+network, considering basic functionality support, including seamless
+migration.
+
+## Overview
+
+### Motivation
+
+Kubevirt has been using the primary pod network with two main bindings:
+masquerade and bridge. In order to support migration,
+masquerade binding is required as it is the only binding that ensures
+the guest IP remains the same while the Pod IP changes.
+The primary pod network binding requirements allow migration, but it
+does not promise any seamless traffic continuity on the primary pod
+network (i.e. existing connectivity passing through the primary pod
+network is expected to interrupt or break).
+
+Seamless traffic continuity is assured only for secondary networks with
+bridge bindings (or any alternative L2 connectivity). Usually the
+secondary networks come without IPAM support, i.e. Kubevirt has no part
+in preserving or managing the IP addresses for these networks. In other
+words, it is up to the users to manage IP assignment and management for
+secondary networks to allow seamless connectivity on migration.
+
+> **Note**: On the primary pod network, classic IPAM solutions would
+> assign IP addresses from different subnets (usually one per node).
+> Therefore, on migration the primary pod network IP changes, causing
+> active connections to be broken.
+
+While secondary networks have been serving many user scenarios, they
+lack the option to utilize many of the common Kubernetes functionality
+(e.g. services). Network providers that support secondary networks,
+sometimes provide only partial functionality compared to the primary pod
+network.
+
+To cope with the limitations network providers introduce to secondary
+networks, Kubevirt needs to provide an alternative to work with the
+primary pod network and still support core features like the seamless
+migration.
+
+### Goals
+
+- Support seamless migration through the primary pod network.
+- Preserve functionality of existing setups (with all their limitations).
+- Define requirements from the network provider & CNI to assure a
+  successful integration.
+
+### Non Goals
+
+- Migrate (automatically) the current existing setup to a new setup binding.
+- Propose a method to deploy the solution (e.g. deploy a binding
+  plugin). This may be covered as a follow-up or in a different proposal
+  per need.
+
+## Network Bindings
+
+Kubevirt uses the network binding to enable connectivity between the VM
+and the pod network interface.
+
+### Core & Plugin
+
+Kubevirt network bindings can be classified into two groups:
+
+- Core network bindings.
+- [Plugin network bindings](https://kubevirt.io/user-guide/network/network_binding_plugins/).
+
+The core bindings are embedded into the Kubevirt codebase and usually
+are the ones frequently used by users (e.g. masquerade, bridge and
+SR-IOV bindings).
+The plugin bindings are not part of the Kubevirt codebase, they
+integrate with Kubevirt through an API. Some are developed and supported
+by the project (e.g. passt binding) and some by 3rd party contributors.
+
+The plugin bindings provide the ability to develop a binding
+out-of-tree, innovating and examining usage without committing for long
+term support in advance.
+Therefore, it will be a natural fit to introduce the new network binding
+effort with a pluggable binding. Once enough usage and experience is
+collected, a pluggable binding may be considered for “promotion” as a
+core binding.
+
+### Primary & Secondary
+
+Kubevirt network bindings can be classified based on their ability to
+work with the primary or secondary networks.
+
+As a general rule, bindings that can consume a CNI IPAM will usually
+work on a primary pod network. The masquerade, bridge and passt bindings
+work with the primary binding.
+The SR-IOV and macvtap bindings are examples of the ones which cannot be
+used on the primary pod network.
+
+> **Note**: Consuming a CNI IPAM requirement for the primary pod network
+> is not a hard requirement. There are exceptions to this, e.g. instead
+> of consuming the CNI IPAM, the guest can consume an IP from the
+> network-provider, set statically through cloud-init or other means.
+
+### Migration Support
+
+Kubevirt allows a VM to migrate only if it uses the masquerade binding
+or a plugin binding that is marked
+as [supporting migration](https://kubevirt.io/user-guide/network/network_binding_plugins/#migration)
+for its primary pod network (or if the primary is not used at all).
+The ability to migrate is agnostic to the support for seamless migration
+with no traffic degradation. In fact, the primary pod network traffic is
+currently expected to get broken during the migration (e.g. with
+masquerade, the pod IP change/officials).
+
+Users are recommended to use the secondary network with a L2
+connectivity to preserve network connectivity with minimal impact. This
+also comes with a limitation of not attempting to manage IP on secondary
+networks, due to dependencies on sticky IP allocations for the pods up
+to the ability to pass the IP allocation down to the VM itself.
+
+The proposed solutions need to take into account the challenges the
+existing known bindings have with the primary pod network and with migration.
+
+## Proposal
+
+The following proposal attempts to provide a solution for VMs that use
+the primary pod network for the majority of the traffic and require a
+seamless migration with no (or minimal) traffic connectivity
+interruption.
+
+The solution should consider two important points:
+
+- Be agnostic to a specific network provided and CNI, by defining the
+  requirements well.
+- Co-exit with the current solutions without adding too much burden on
+  maintenance. E.g. the logic and codebase should not diverge too much
+  from the existing core network flow.
+
+### Pre Requirements
+
+The main challenge when working with the primary pod network is to
+assure connectivity recovers after a VM migration. A new requirement is
+added with this proposal, to preserve existing connectivity (e.g. TCP
+connection) with minimal (to none) disruption.
+
+As we are focusing here on the primary pod network, these requirements
+translate to several pre-requirements from the network provider & CNI:
+
+- Sticky IP: An IPAM plugin assigns an IP for each pod. In order to
+  support migration and preserve seamless connectivity, the pod IP on
+  the target node must use the same IP.
+  The network-provider/CNI/IPAM needs to assure such an option exists so
+  the same IP used for the primary pod network can be moved to the
+  target pod.
+- Flat Cluster Subnet for the Pod network: In order to allow the same IP
+  to be assigned on a different node when migration occurs, only a
+  single subnet needs to exist from the entire cluster on the primary
+  pod network.
+- Traffic hand-off on migration: With the need to keep the same IP on
+  both the source and target nodes/pods, the network-provider needs to
+  assure traffic is correctly directed to the active VM. I.e.during the
+  migration process, the source VM is the active one (which may take
+  several tens of seconds), afterward when the hand-off is moved to the
+  target, traffic should start being directed to the new location.
+  A Kubevirt VM signals that it is now active by sending a GARP message,
+  letting all the bridges on the network know that the mac address
+  moved. This may be enough for some network-providers while others may
+  use other techniques to correctly forward the traffic to the right
+  port.
+
+> **Note**: In case this is not handled properly, traffic may be disrupted
+> during and following a migration in unexpected ways.
+
+These are the common pre-requirements for all the solutions options that
+this proposal includes. Each proposed binding may have unique
+requirements of their own.
+
+### Network Binding Options
+
+There are a few existing bindings (passt, bridge) and non-existing
+futuristic ones which can be developed to solve the needs. This section
+presents each with their pros and cons.
+
+The chosen solution and a table summary is provided first, followed by a
+detailed description of each.
+
+<!-- TOC -->
+* [The Binding Winner](#the-binding-winner)
+* [Summary](#summary-1)
+* [Passt Binding](#passt-binding)
+* [Bridge Binding](#bridge-binding)
+  * [As a Binding Plugin](#as-a-binding-plugin)
+  * [As a DomainAttachmentType](#as-a-bridge-domainattachmenttype)
+* [Macvtap Binding](#macvtap-binding)
+  * [As a DomainAttachmentType](#as-a-macvtap-domainattachmenttype)
+* [Other L2 Bindings](#other-l2-bindings)
+<!-- TOC -->
+
+#### The Binding Winner
+
+The bridge binding plugin option with a `DomainAttachmentType` of
+`managedTap` has been chosen after a successful initial PoC succeeded.
+
+The original plan has been preserved for ref below.
+
+Considering the requirements, development effort and maintenance cost,
+the following plan was chosen:
+1. Develop a **macvtap** network binding plugin. [**Failed**]
+2. In case option [1] fails, develop a **bridge** network binding plugin. [**Succeeded**]
+3. In case option [2] fails, enhance the existing **bridge** core binding.
+4. Develop and enhance the Passt network binding plugin for the mid/long term.
+   Passt will then replace the previous binding in a way that is
+   transparent to users.
+   > **Note**: This effort can run in parallel to the previous points
+
+#### Summary
+
+| Feature\\Binding            | Passt                                                                                                           | Bridge[*]                            | Macvtap                                          |
+|:----------------------------|:----------------------------------------------------------------------------------------------------------------|:-------------------------------------|:-------------------------------------------------|
+| Deployment                  | Binding Plugin                                                                                                  | Core                                 | Binding Plugin                                   |
+| IP & Route delivery         | Self (IPv4 & IPV6)                                                                                              | Self for IPv4, Net Provider for IPv6 | Net Provider                                     |
+| Pod Probes Support          | Yes                                                                                                             | No                                   | No                                               |
+| Seamless Live-Migration     | Requires Development                                                                                            | Yes                                  | Yes                                              |
+| Supported Traffic           | TCP/UDP/ICMP (All IP in case seamless migration is resolved)                                                    | IP                                   | IP                                               |
+| Migration Support Indicator | Yes                                                                                                             | Requires Development                 | Yes                                              |
+| Mesh / Istio support        | Yes                                                                                                             | No                                   | No                                               |
+| IP Status (reporting)       | Yes                                                                                                             | Yes                                  | Requires Development                             |
+| Observability Integration   | Yes                                                                                                             | No                                   | No                                               |
+| Resource Overhead           | 500MB per VM, 250MB per VM already available on U/S, Negligible overhead when resolving the seamless migration. | Negligible                           | Negligible                                       |
+| Performance                 | N/A                                                                                                             | N/A                                  | N/A , Most likely better than the bridge option. |
+| Kubevirt Support            | Low                                                                                                             | High                                 | Medium                                           |
+
+[*] Core Bridge binding can be deployed as a binding plugin with a subset
+    of its existing feature sets. E.g. as a plugin, IP & routes can be
+    supported only through the network-provider.
+    As a plugin, it benefits from having a migration indicator and lower
+    support cost.
+
+#### Passt Binding
+
+Passt is available through
+an [existing](https://kubevirt.io/user-guide/network/net_binding_plugins/passt/)
+network binding plugin that provides user space network connectivity.
+It is considered a candidate to replace the existing masquerade binding
+due to the simplification it introduces.
+
+Pros:
+
+- Self-contained IP stack solution (IPv4 & IPV6), independent of the
+  network-provider. This allows it to work with a larger number of
+  network-providers and CNIs.
+- Trivial integration with Mesh solutions like [Istio](https://istio.io)
+  with no special treatment from Kubevirt.
+- Integrates with the VM interfaces network status reporting logic.
+- Integrates with 3rd party network monitoring & collection solutions (
+  e.g. [stackrox](https://www.stackrox.io/)).
+- Maintenance & support cost is low as most functionality is deferred on
+  the Passt project (which is actively used in Podman, libkrun, krun,
+  optional usage in Docker via rootlesskit/moby, and only supported
+  user-mode network solution starting from RHEL 10 for libguestfs,
+  libvirt, as libslirp is deprecated there).
+
+Cons:
+
+- Attempting to use it for live-migration results in breaking TCP
+  connections.
+- Passt supports TCP, UDP and ICMP traffic only. While most traffic
+  today is encapsulated in one of these common L4 protocols, it is hard
+  to predict what customers are using.
+- Resource overhead: When all ports are to be forwarded to the guest,
+  Passt is expected to use around 500MB of kernel memory. Work has been
+  invested in this to reduce the overhead to around 250MB.
+
+Performance:
+
+Unfortunately, there are no measurements done to compare it to other
+bindings (mainly comparable to the masquerade binding).
+
+##### Required Development:
+
+In order to comply with the needs set in this design, Passt must solve
+the seamless live-migration limitation of breaking TCP connections and
+hopefully also treat the other two cons (limited traffic protocols and
+resource usage).
+
+Several options have been raised to tackle these challenges:
+
+1. Instead of using TCP/UDP/ICMP sockets, use an IP socket.
+   * TCP connections are not expected to break.
+     * Traffic will no longer be limited to one of these protocols.
+     * Resource overhead will be negligible.
+     * Challenge: Opening an IP socket requires a NET_RAW capability.
+       * One option to resolve this is for the virt-handler to add the
+         capability on the Passt binary.
+         * Another option is to create the socket at the virt-handler and
+           pass it to the virt-launcher using a unix socket.
+2. Migrate the VM without migrating Passt, which runs on a third, "
+   stable" node, providing a detached "exit point" for network traffic.
+   * This does not seem to be an acceptable solution because the
+     expectation is for the VM to migrate when the node is evicting or
+     when overloaded. It is enough to consider a node upgrade reason
+     for the eviction to see that eventually Passt will also have to
+     move out and break connections.
+3. Using the TCP_REPAIR socket option, the old Passt instance dumps
+   existing connection states from the kernel to a file, which is loaded
+   by the new instance.
+   * Requires development and integration with Qemu.
+4. Run passt in its own bridged VM and migrate it with the main VM.
+   * Migrating two VMs in parallel, assuring they reach the same node
+     and that both perform the handoff in a correct timing seems challenging.
+
+Attempting to go with option (1) above requires a PoC that changing the
+capability of the Passt binary is possible (est. 2 days) and Passt
+development to support this work mode (est. 20 days).
+
+#### Bridge Binding
+
+The bridge binding is
+an [existing](https://kubevirt.io/user-guide/network/interfaces_and_networks/#bridge)
+core binding that provides a L2 connectivity with a limited support for
+IPAM consumption. The binding is commonly used on secondary networks
+without IPAM and on the primary pod network with IPAM.  
+Kubevirt blocks migrations when the binding is used on the primary pod
+network (although a special informal annotation on the VMI may be used
+to overcome this limitation).
+
+Pros:
+
+- Self-contained, supports the reflection of the IPv4 stack to the
+  guest (independent of the network-provider).
+- Provides seamless live-migration due to its L2 wiring.
+- Supports all traffic types.
+- Integrates with the VM interfaces network status reporting logic.
+- Core binding with no need to depend on an external out-of-tree plugin.
+
+Cons:
+
+- Does not support the reflection of the IPv6 stack to the guest (
+  requires support from the network-provider).
+    - While DHCPv6 is supported, RA is not.
+- It is missing a live-migration support indicator to allow Kubevirt to
+  determine if a migration is allowed or not.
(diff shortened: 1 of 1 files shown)


design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,459 @@
...
+- Tap device directly connected to the node, passed into the pod for
+  consumption (saving the veth pair that links the pod interface to the
+  node).
+
+#### Summary
+
+| Feature\\Binding            | Passt                                                                                                           | Bridge[*]                            | Macvtap                                          |
+|:----------------------------|:----------------------------------------------------------------------------------------------------------------|:-------------------------------------|:-------------------------------------------------|
+| Deployment                  | Binding Plugin                                                                                                  | Core                                 | Binding Plugin                                   |
+| IP & Route delivery         | Self (IPv4 & IPV6)                                                                                              | Self for IPv4, Net Provider for IPv6 | Net Provider                                     |
```
qinqon, line 423: We should add here another row related to VM probes, I think with passt we support them with the bridge and macvtap we do not, the ovn-k version of primary network the probes traffic goes to eth0, same as kubevirt not supporting probes for secondary interfaces.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,459 @@
...
+  assure traffic is correctly directed to the active VM. I.e.during the
+  migration process, the source VM is the active one (which may take
+  several tens of seconds), afterward when the hand-off is moved to the
+  target, traffic should start being directed to the new location.
+  A Kubevirt VM signals that it is now active by sending a GARP message,
+  letting all the bridges on the network know that the mac address
+  moved. This may be enough for some network-providers while others may
+  use other techniques to correctly forward the traffic to the right
+  port.
+
```
qinqon, line 161: Another pre-requisite is related to this primary network being able to access some default network services, using default gw, if not we don't have stuff like dns resolution.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,459 @@
...
+The bridge binding can potentially be implemented as a binding plugin.
+The community has already shown interest in this direction with an
+active [contribution](https://github.com/kubevirt/kubevirt/pull/12671).
+
+A plugin solves the live-migration support indicator (as it is part of
+the [plugin API](https://kubevirt.io/user-guide/network/network_binding_plugins/#migration))
+but on the other hand introduces a challenge with the DHCP servers and
+data sharing between the CNI and virt-launcher (in the core, this is
+solved by the virt-handler persisting the data on a file for the
+virt-launcher to consume).
```
qinqon, line 346: We can skip DHCPv4 too to simplify this effort, it can also be provided by the CNI provider.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,459 @@
...
+macvtap bindings, e.g.:
+
+- Passing the IP & route configuration from the pod interface to the VM
+  guest interface is left to the network provider (e.g. DHCPv4, DHCPv6, RA).
+
+Some of the suggested L2 alternatives are:
+
+- OVS bridge
+- eBPF
+- Tap device directly connected to the node, passed into the pod for
```
qinqon, line 414: In the past we asked about this and looks like tap device get disconnected when moved to the network namespace, although I don't remeber the details.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,459 @@
...
+  consumption (saving the veth pair that links the pod interface to the
+  node).
+
+#### Summary
+
+| Feature\\Binding            | Passt                                                                                                           | Bridge[*]                            | Macvtap                                          |
```
qinqon, line 420: We should split core and pluging net binding since it has different characteristics.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,459 @@
...
+    supported only through the network-provider.
+
+### The Binding Winner
+
+Considering the requirements, development effort and maintenance cost,
+the following plan is chosen:
+1. Develop a **macvtap** network binding plugin.
+2. In case option [1] fails, develop a **bridge** network binding plugin.
+3. In case option [2] fails, enhance the existing **bridge** core binding.
+4. Develop and enhance the Passt network binding plugin for the mid/long term.
```
qinqon, line 444: I think this do not have to wait for 1, 2 and 3, ideally we move to passt in the future so it can be developed in parallel.

qinqon (changes requested):

design-proposals/primary_network_binding_for_seamless_migration.md
maiqueb: huge +1.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,459 @@
...
+> interface and multiple mac addresses are active in the guest.
+
+Deployed as a binding plugin with no option to self-serve IPAM to the VM.
+It is dependent on the network-provider to be able and serve IP
+addresses to the guest.
+
+Compared to the bridge binding (as a plugin), it is simpler and with
+probably better performance.
+In addition, macvtap devices have a device representation and can be
+relabeled (SELinux) post creation.
```
maiqueb, line 390: I'm nearly positive this the issue goes beyond selinux; I'm not sure the corresponding tap device in the pod will be usable to send/receive packets, unless you adjust the pod's cgroups.

@qinqon how is your PoC going ?...

qinqon: @maiqueb didn't have time for it :-/.

EdDev: > I'm nearly positive this the issue goes beyond selinux; I'm not sure the corresponding tap device in the pod will be usable to send/receive packets, unless you adjust the pod's cgroups.

The CNI is expected to create the `macvtap` device in the pod netns. I would expect it to just do the right thing.
If these kind of problems will appear, we will have to mitigate it somehow (use a dp, change the cgroup using the virt-handler, etc.).

@orelmisan is working on a PoC to check this.

EdDev: > I'm only trying to say that bridge (to me) sounds like the path of least resistance / simpler thing to do.
> 
> And that "it is simpler and with probably better performance." is opinionated (at least the simpler part).

I'm unsure to what "simpler" option you refer to.
The bridge binding option involves a tap device, which I tried to express it is much harder to work with in a binding plugin.
Do you refer to something else?

EdDev:
I have added the option of using `domainAttachmentType` without the need of a CNI or sidecar.

design-proposals/primary_network_binding_for_seamless_migration.md
qinqon: You need to colums for bridge, you with core and another with plugin and put the different options

EdDev: I think the macvtap column is representing best a plugin and adding another one will overload the table.

The information is available for a reader that wants to get in more details.

qinqon: I mean the Bridge deployment type say "Core" but looks like we may deploy it as binding and the non core linux bridge has different characteristics.

enp0s3 (commented):
@EdDev Hi! 
I have a general question - you are using the term `network provider`, can you please tell how its different from a CNI? what functionality does it provides that is missing in a CNI?

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+
+- The live-migration support indicator options:
+  - Drop the existing limitation that blocks the migration when bridge
+    binding is used with the primary pod network. I.e. always allow
+    migration.
+    - Not recommended because the VM may lose connectivity after migration.
+  - Add an API field to indicate if the VM is migratable or not:
+    - Per interface, being relevant only when used for a primary pod network.
+    - Per the pod network type.
+    - At the cluster level, in the Kubevirt CR.
```
enp0s3, line 310: Today we determine automatically whether the VM is migratable or not, can you please elaborate on the user-case to let external entity to decide about it via API?

RamLavi (commented):
Thank you for the thorough review of options! Added a small typo comment

design-proposals/primary_network_binding_for_seamless_migration.md
EdDev: The "user-case" is expressed in the cons section:
> - It is missing a live-migration support indicator to allow Kubevirt to
  determine if a migration is allowed or not.
  
As the core bridge binding, when used on the primary pod network, is blocking migration. We need a way to indicate it can support migration and one option is to expose an API that a user or cluster admin can mark this.

In comparison, the plugin bindings have a knob for this.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,462 @@
...
+futuristic ones which can be developed to solve the needs. This section
+presents each with their pros and cons.
+A summary is given at the end of this section to ease comparison.
+
+#### Passt Binding
+
+Passt is available through
+an [existing](https://kubevirt.io/user-guide/network/net_binding_plugins/passt/)
+network binding plugin that provides user space network connectivity.
+It is considered a candidate to replace the existing masquerade binging
```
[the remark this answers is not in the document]
EdDev: Thanks! Done

EdDev (commented):
> I have a general question - you are using the term `network provider`, can you please tell how its different from a CNI? what functionality does it provides that is missing in a CNI?

The term was a bit more visible in the past, the documentation seems to have reduce the clarification on the difference between the CNI and the network provider.

At the end of the network-plugin [installation](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#installation) section it is mentioned. From there, you could see it is stated at a few networking addons [here](https://kubernetes.io/docs/concepts/cluster-administration/addons/#networking-and-network-policy).

While the CNI is a specification that describes how to connect the pod to the node, or better said to the network. The network provider is the actual implementation of that network, e.g. connecting all the pods or nodes together. 
In practice, each provider solution comes with its own CNI plugin, so it could be understandable while it is harder to differentiate the two parts.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+> PoC confirmation.
+
+As a plugin, it can be developed independently to the Kubevirt release
+cycle on a different repo.
+
+##### As a DomainAttachmentType
+
+The network binding plugin exposes a
+[DomainAttachmentType](https://kubevirt.io/user-guide/network/network_binding_plugins/#domainattachmenttype)
+filed which provides pre-defined core Kubevirt methods to attach an
```
orelmisan, line 413:
```suggestion
field which provides pre-defined core Kubevirt methods to attach an
```

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+    subgraph virt-launcher
+    VM <--> macvtap0
+    macvtap0 <--> eth0
+    end
+    subgraph Node
+    eth0 <--> veth123
+    end
+```
+
+The macvtap binding is an alternative to the bridge binding with very
```
orelmisan, line 385: Please emphasize that this suggestion is not referring to the existing macvtap binding plugin:
https://kubevirt.io/user-guide/network/net_binding_plugins/macvtap/

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+##### As a DomainAttachmentType
+
+The network binding plugin exposes a
+[DomainAttachmentType](https://kubevirt.io/user-guide/network/network_binding_plugins/#domainattachmenttype)
+filed which provides pre-defined core Kubevirt methods to attach an
+interface to the domain.
+
+The macvtap binding plugin may utilize this field by adding a new
+attachment type that can both create the macvtap device in the pod and
+define it in the domain configuration.
```
orelmisan, line 418: Is this the macvtap binding plugin suggested above? or the [existing](https://kubevirt.io/user-guide/network/net_binding_plugins/macvtap/) one?

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+
+The new type is similar to a managed interface in a (libvirt) domain,
+where the definition of the interface also attempts to create the
+relevant devices.
+Therefore, the new type will be called `managedMacvtap`.
+
+When the `domainAttachmentType` is set with a value of `managedMacvtap`,
+the virt-handler will make sure to create the macvlan device and connect
```
orelmisan, line 426: Is macvlan a typo?

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+- Does not require a CNI, reducing maintenance and deployment cost.
+- Avoids SELinux and cgroup issues by creating the device in the context
+  of the virt-launcher pod compute container, similar to the creation of
+  the `tap` device (for the core bindings).
+
+> **Note**: The attachment is expected to include basic attributes which
+> should fit the Kubevirt use case. E.g. using `passthru` as the macvtap
+> mode.
+
+To register the network binding plugin, no CNI or sidecar is required:
```
orelmisan, line 442:
```suggestion
To register the network binding plugin, no CNI or sidecar are required:
```

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+| Observability Integration   | Yes                                                                                                             | No                                   | No                                               |
+| Resource Overhead           | 500MB per VM, 250MB per VM already available on U/S, Negligible overhead when resolving the seamless migration. | Negligible                           | Negligible                                       |
+| Performance                 | N/A                                                                                                             | N/A                                  | N/A , Most likely better than the bridge option. |
+| Kubevirt Support            | Low                                                                                                             | High                                 | Medium                                           |
+
+[*] Core Bridge binding can be deployed as a binding plugin with a subset
+    of its existing feature sets. E.g. as a plugin, IP & routes can be
+    supported only through the network-provider.
+
+### The Binding Winner
```
orelmisan, line 497: Could you please update this section following the results of the macvtap CNI PoC?

orelmisan (commented):
Thank you for the detailed proposal @EdDev

design-proposals/primary_network_binding_for_seamless_migration.md
orelmisan: I have completed the PoC for creating a macvtap in the virt-launcher pod using the proposed CNI. It has its challenges and while the results were informative, we've decided not to invest further in this approach at this time, as it does not align with our current priorities. We may consider this option in the future.

design-proposals/primary_network_binding_for_seamless_migration.md
qinqon: All good we can close this.

design-proposals/primary_network_binding_for_seamless_migration.md
qinqon: Ok.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+| IP & Route delivery         | Self (IPv4 & IPV6)                                                                                              | Self for IPv4, Net Provider for IPv6 | Net Provider                                     |
+| Pod Probes Support          | Yes                                                                                                             | No                                   | No                                               |
+| Seamless Live-Migration     | Requires Development                                                                                            | Yes                                  | Yes                                              |
+| Supported Traffic           | TCP/UDP/ICMP (All IP in case seamless migration is resolved)                                                    | IP                                   | IP                                               |
+| Migration Support Indicator | Yes                                                                                                             | Requires Development                 | Yes                                              |
```
maiqueb, line 485: maybe bridge doesn't require this at all. 

One of the options you present in your enhancement is literally:
> Drop the existing limitation that blocks the migration when bridge
    binding is used with the primary pod network. I.e. always allow
    migration.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+| IP Status (reporting)       | Yes                                                                                                             | Yes                                  | Requires Development                             |
+| Observability Integration   | Yes                                                                                                             | No                                   | No                                               |
+| Resource Overhead           | 500MB per VM, 250MB per VM already available on U/S, Negligible overhead when resolving the seamless migration. | Negligible                           | Negligible                                       |
+| Performance                 | N/A                                                                                                             | N/A                                  | N/A , Most likely better than the bridge option. |
+| Kubevirt Support            | Low                                                                                                             | High                                 | Medium                                           |
```
maiqueb, line 491: This entry is misleading: if bridge is deployed as a plugin, your maintenance costs will be quite lower.

You also don't mention the migration indicator (which I am unsure is required ...) is also not called out as a difference.

I think you should mention that in the callout below.

design-proposals/primary_network_binding_for_seamless_migration.md
maiqueb: There are more differences than the (single) one listed in the callout immediately after the table - at least maintenance cost, and Migration Support Indicator (which is granted for free, since it is a plugin). 

I also believe having a separate column would be a lot clearer / easier to understand.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+    of its existing feature sets. E.g. as a plugin, IP & routes can be
+    supported only through the network-provider.
+
+### The Binding Winner
+
+Considering the requirements, development effort and maintenance cost,
+the following plan is chosen:
+1. Develop a **macvtap** network binding plugin.
+2. In case option [1] fails, develop a **bridge** network binding plugin.
+3. In case option [2] fails, enhance the existing **bridge** core binding.
```
maiqueb, line 503: It is really not clear to me why macvtap is preferred to bridge, especially when deployed as a plugin.

The order of the priorities here is not what I would expect to be honest, after going through the table above.

design-proposals/primary_network_binding_for_seamless_migration.md
maiqueb: I am assuming the former, but it should definitely be clarified. If you wonder about it, more people will.

maiqueb (changes requested):
Thank you; I need more probe related information, and I overall disagree with the priority order of the proposals.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+addresses to the guest.
+
+Compared to the bridge binding (as a plugin), it is simpler and with
+probably better performance.
+In addition, macvtap devices have a device representation and can be
+relabeled (SELinux) post creation. However, as a device, there is a need
+to adjust it to the virt-launcher compute cgroup.
+
+> *Warning*: Relabeling of a macvtap device and cgroup adjustment needs
+> PoC confirmation.
```
enp0s3, line 404: @EdDev Can you please elaborate on the cgroup thing? I am curious which cgroup controller needs to be updated w.r.t. this

design-proposals/primary_network_binding_for_seamless_migration.md
enp0s3: @EdDev Can you please elaborate on the meaning of Kubevirt support? how do you measure it?

enp0s3: @EdDev By `high` do you mean the boilerplate of SELinux re-labeling and the cgroup adjustment?

design-proposals/primary_network_binding_for_seamless_migration.md
EdDev: I also updated the results in the specific sections.

Done

design-proposals/primary_network_binding_for_seamless_migration.md
EdDev: IMO the bridge one has lower performance compared to the macvtap passthru.
If we did not encounter the problems with creating a macvtap through the CNI, we would even had this out-of-tree, reducing the maintenance cost.

Even now, after discovering all the problem with cgroups and the yet unclear reason it does not work for ingress traffic, I would have not prioritize this differently.

Anyway, as the current attempt failed with macvtap and the realization that cgroup manipulation is really problematic in Kubevirt today, we ended up with choosing the option of the bridge plugin (which seems to work as expected).

EdDev (commented):
Thank you all for the review and sorry for the delay in responding.

Main [changes](https://github.com/kubevirt/community/compare/4f132f131d94469d962882450d9c949da458b2ba..d9d9d329e956ef5fb9877e76d655015f75e72b19): Responded to the review comments and re-arranged the proposal such that the solution and summary is presented first. All lower details are given later. Hopefully that should ease the readability of the important parts.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+sometimes only partial functionality compared to the primary pod network.
+
+To cope with the limitations network providers introduce to secondary
+networks, Kubevirt needs to provide an alternative and work with the
+primary pod network and still support core features like the seamless
+migration.
+
+### Goals
+
+- Support seamless migration through the primary pod network.
```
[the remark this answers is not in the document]
maiqueb: Seamless, which means no traffic disruption .

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,520 @@
...
+      support various guest OS with a potential need to increase/change
+      the options it serves.
+
+##### Required Development
+
+- The live-migration support indicator options:
+  - Drop the existing limitation that blocks the migration when bridge
+    binding is used with the primary pod network. I.e. always allow
+    migration.
+    - Not recommended because the VM may lose connectivity after migration.
```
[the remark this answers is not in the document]
maiqueb: > > how is this any different from what you have in masquerade ?
> 
> The masq one keeps the VM with the same IP internally, while externally the IP changes. This allows connections to be re-establish with no actions from the VM admin or guest user.

I don't follow; in masquerade, the pods are reached using the pod IP. On a migration (assuming the CNI does not implement IP stickiness), the pod IP changes, and the client has no way on knowing what's the new IP unless it checks the VMI object status (I'm assuming we report the pod IP in the VMI status; I don't remember the details). How can connections be re-establishes with no actions ? Surely I am missing something.

> 
> With bridge core binding, the scenarios in which the IP is changed will not be reflected to the guest until a DHCP client refresh occurs. Until then, it will have an IP which cannot communicate and therefore the connectivity will not be restored in a reasonable time.

I am assuming the plugin used implements IP stickiness for bridge binding. That's what I mean below with the admin should choose a network / configuration that fulfills the live-migration requirements. As they must do with storage (RWX block volumes, etc).

> 
> > It would be the admin's responsibility to choose a network / provide configuration that fulfills the live migration requirements.
> 
> Kubevirt attempts to provide and own functionality and not blindly pass it to its users. In general, the network bindings are suppose to work with a variety of CNI solutions, especially the core ones. Specialized bindings can be introduces as plugins, however, even then we would prefer them to be as generic as possible in order to 1) allow their re-usability with as many CNIs as possible and 2) to make them simple to integrate into the Kubevirt core (e.g. comply with status report).
> 
> > Out of curiosity... what do we do with storage ? AFAIU, there are also requirements on storage for the VM to live-migrate. Do they prevent the migration if the requirements are not met ?
> 
> AFAIK yes. But I am not in the details on what is known in advance and what is discovered when the target pod is not able to be scheduled.

So no news here - the admin must choose components that fulfill the feature's requirements.

ormergi (commented):
Looks good to me overall, really nice overview on networking aspects in migration.

I need to go over this once more and the discussion though.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,596 @@
...
+  * [As a Binding Plugin](#as-a-binding-plugin)
+  * [As a DomainAttachmentType](#as-a-bridge-domainattachmenttype)
+* [Macvtap Binding](#macvtap-binding)
+  * [As a DomainAttachmentType](#as-a-macvtap-domainattachmenttype)
+* [Other L2 Bindings](#other-l2-bindings)
+<!-- TOC -->
+
+#### The Binding Winner
+
+The bridge binding plugin option has been chosen after a successful
```
xpivarc, line 198: It wasn't clear to me that the `As a bridge DomainAttachmentType` is the winner.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,596 @@
...
+
+### Motivation
+
+Kubevirt has been using the primary pod network with two main bindings:
+masquerade and bridge. In order to support migration,
+masquerade binding is required as it is the only binding that ensures
+the guest IP remains the same while the Pod IP changes.
+The primary pod network binding requirements allow migration, but it
+does not promise any seamless traffic continuity on the primary pod
+network (i.e. existing connectivity passing through the primary pod
```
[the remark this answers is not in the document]
EdDev: Done

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,596 @@
...
+
+Seamless traffic continuity is assured only for secondary networks with
+bridge bindings (or any alternative L2 connectivity). Usually the
+secondary networks come without IPAM support, i.e. Kubevirt has no part
+in preserving or managing the IP addresses for these networks. In other
+words, it is up to the users to manage IP assignment and management for
+secondary networks to allow seamless connectivity on migration.
+
+> **Note**: On the primary pod network, classic IPAM solutions would
+> assign IP addresses from different subnets (one per node).
```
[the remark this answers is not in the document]
EdDev: Done

design-proposals/primary_network_binding_for_seamless_migration.md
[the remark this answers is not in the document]
EdDev: This is the goal of this proposal. If we did not had this goal in mind, we could have used the regular masquerade binding. So I am not sure about what you mean by your comment here.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,596 @@
...
+also comes with a limitation of not attempting to manage IP on secondary
+networks, due to dependencies on sticky IP allocations for the pods up
+to the ability to pass the IP allocation down to the VM itself.
+
+The proposed solutions need to take into account the challenges the
+existing known bindings have with the primary pod network and with migration.
+
+## Proposal
+
+The following proposal attempts to provide a solution from VMs that use
```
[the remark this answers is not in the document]
EdDev: Done

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,596 @@
...
+As we are focusing here on the primary pod network, these requirements
+translate to several pre-requirements from the network provider & CNI:
+
+- Sticky IP: An IPAM plugin assigns an IP for each pod. In order to
+  support migration and preserve seamless connectivity, the pod IP on
+  the target node must use the same IP.
+  The network-provider/CNI/IPAM needs to assure such an option exists so
+  the same IP used for the primary pod network can be moved to the
+  target pod.
+- Flat Cluster Subnet: In order to allow the same IP to be assigned on a
```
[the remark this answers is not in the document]
EdDev: > All I am asking here is to clarify what subnet you are talking about.

Done.

> Additionally, do we really need flat subnet? I could imagine a working scenario (hacky for sure) where you could persist the IP across nodes without flat subnet. So this point is just specification of the first one and my question would be if that is really needed.

If we just preserve the IP but not the subnet, then the pod on the other node will use an IP that is not included in that node subnet, therefore it will not be able to communicate.
In this context, a subnet is equivalent to a LAN.

design-proposals/primary_network_binding_for_seamless_migration.md
EdDev: Done

EdDev:
[change](https://github.com/kubevirt/community/compare/d9d9d329e956ef5fb9877e76d655015f75e72b19..e9edb4a2317d633cea209b9e2cb7d81b9899ad34): Answered latest review comments.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,596 @@
...
+  different node when migration occurs, only a single subnet needs to
+  exist from the entire cluster on the primary pod network.
+- Traffic hand-off on migration: With the need to keep the same IP on
```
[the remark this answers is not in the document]
EdDev: >   same IP on both the source and target nodes/pods, the network-provider needs to
>   assure traffic is correctly directed to the active VM. I.e.during the
>   migration process, the source VM is the active one (which may take
>   several tens of seconds), afterward when the hand-off is moved to the target

This expresses that the two pods co-exist. If you prefer some other clearer wording, could you please suggest something?

> and that the switch is also a problem.

This is the network-provider business, not ours. I mean, the pre-requirment is for it to assure traffic flows correctly to the two pods. I tried to avoid using specifics as much as possible, just because I do not want to hint about any specific implementation.
I do know that for SDN providers (e.g. OVN), they can duplicate traffic to both ports.

design-proposals/primary_network_binding_for_seamless_migration.md
```diff
@@ -0,0 +1,597 @@
...
+The core bindings are embedded into the Kubevirt codebase and usually
+are the ones frequently used by users (e.g. masquerade, bridge and
+SR-IOV bindings).
+The plugin bindings are not part of the Kubevirt codebase, they
+integrate with Kubevirt through an API. Some are developed and supported
+by the project (e.g. passt binding) and some by 3rd party contributors.
+
+The plugin bindings provide the ability to develop a binding
+out-of-tree, innovating and examining usage without committing for long
+term support in advance.
```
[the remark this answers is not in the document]
vladikr: @EdDev ^^

vladikr:
/approve
I don't know why did we need to mention the nature of the network plugins support in this proposal. We could simply point to a centralized description of of out of tree network plugins.
If we are planning to mention this, please also make sure that it is clear that the support for these plugins is best effort only. 
These plugins don't use an officially supported KubeVirt API.

/hold

design-proposals/primary_network_binding_for_seamless_migration.md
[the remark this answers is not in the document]
EdDev: @vladikr , The intention of this text is to express the diff between solving the need here with a core binding or a plugin. It this sense, it basically says that we can introduce the binding as a plugin now, support it (as the vendors of it) and if found to be best solution, to consider it as a core.

In any case, the plugins we (Kubevirt sig-network) create, can be consider supported by us.
But I agree about the more general idea: Binding plugins are not supported by Kubevirt as expressed [here](https://github.com/kubevirt/user-guide/pull/842).

vladikr:
> > If we are planning to mention this, please also make sure that it is clear that the support for these plugins is best effort only.
> > These plugins don't use an officially supported KubeVirt API.
> 
> This plugin is fully supported because Kubevirt SIG-Network authored it. It uses a Beta feature, so it is valid while the API is there. As Beta, it can either GA or drop (and with it all the binding plugins).
> 
> Is this acceptable?

Yes, no issues with this plugin. I am only concerned about the general statement on line 79-80

design-proposals/primary_network_binding_for_seamless_migration.md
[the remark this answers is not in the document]
vladikr: Thanks, I think I misread this. Thanks for clarifying.

PR #318, opened 2024-08-20
design-proposal: instancetype.kubevirt.io - Support custom resource sizing

lyarwood:
**What this PR does / why we need it**:

Instance types and preferences provide a way to define common VirtualMachine workload resource sizing and preferences.

The common-instancetypes project within KubeVirt currently provides a set of standardized instance types and preferences that can be deployed by `virt-operator` alongside all of the other core KubeVirt components.

While the resources provided by the common-instancetypes project are a great starting point they are generalized and could never possibly cover all workload resourcing and preference use cases. 

As such this design proposal seeks to make it as easy as possible for both cluster admins and users to create their own customized versions of these commonly deployed resources specific to their workloads. Additionally the instance type API itself will be made more flexible by making vCPU and memory resource sizing optional within an instance type, allowing users to provide their own sizing within their VirtualMachine while still retaining the remaining benefits of using an instance type.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**

This checklist is not enforcing, but it's a reminder of items that could be relevant to every PR.
Approvers are expected to review this list.

- [ ] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [ ] PR: The PR description is expressive enough and will help future contributors
- [ ] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [ ] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [ ] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
- [ ] Testing: New code requires [new unit tests](https://github.com/kubevirt/kubevirt/blob/main/docs/reviewer-guide.md#when-is-a-pr-good-enough). New features and bug fixes require at least on e2e test
- [ ] Documentation: A [user-guide update](https://github.com/kubevirt/user-guide/) was considered and is present (link) or not required. You want a user-guide update if it's a user facing feature / API change.
- [ ] Community: Announcement to [kubevirt-dev](https://groups.google.com/g/kubevirt-dev/) was considered

**Release note**:

```release-note
NONE
```

PR #326, opened 2024-09-24
design-proposal: instancetype.kubevirt.io - Expand during submission

lyarwood:

The initial design of instance types and preferences captured a point in time copy of the original resource for future use by the VirtualMachine to ensure we always generate the same VirtualMachineInstance at runtime.

This additionally allows users to easily switch between classes, sizes and generations of these resources at some point in the future. However this flexibility comes at the cost of complexity within KubeVirt itself especially handling more advanced VirtualMachine lifecycle flows.

This design proposal aims to set out a new cluster configurable to control how instance types and preferences are referenced from a VirtualMachine.

Fixes #

**Checklist**

**Release note**:

```release-note
NONE
```

### DIFF
--- a/design-proposals/instancetype.kubevirt.io/expand-during-submission.md
+++ b/design-proposals/instancetype.kubevirt.io/expand-during-submission.md
@@ -1,0 +1,255 @@
+# Overview
+
+The initial design of instance types and preferences captured a point in time
+revision of the original resource for future use by the `VirtualMachine` to
+ensure we always generate the same `VirtualMachineInstance` at runtime.
+
+This additionally allows users to easily switch between classes, sizes and
+generations of these resources at some point in the future. However this
+flexibility comes at the cost of complexity within KubeVirt itself when handling
+more advanced `VirtualMachine` lifecycle flows.
+
+This design proposal aims to set out new alternative and configurable behavior
+where these revisions are no longer captured and are instead expanded into the
+`VirtualMachine`.
+
+## Motivation
+
+The complexity of managing these point in time revisions of instance types and
+preferences has grown over time as more features and versions of the CRDs have
+landed.
+
+Some of this complexity is exposed to users and third party integrations
+such as back-up tooling or user-interfaces built on top of KubeVirt. This can
+take the form of differing entry points for certain functionality such as hot
+plug to requiring knowledge of the stored revisions to allow for a valid back up
+and eventual restore of a given VirtualMachine to happen.
+
+## Goals
+
+* Provide a simple cluster configurable to control how
+  instance types and preferences are referenced from a VirtualMachine
+
+## Non Goals
+
+* The default behavior will not change as part of this work
+
+## User Stories
+
+* As a cluster owner I want to control how instance types and preferences are
+  referenced from VirtualMachines within my environment
+
+## Repos
+
+* kubevirt/kubevirt
+
+## Design
+
+A new KubeVirt configurable will be introduced to control how instance types and
+preferences are referenced from VirtualMachines.
+
+This configurable will provide an `InstancetypeReferencePolicy` that
+encapsulates this behaviour. The following policies will be initially provided:
+
+* `reference` (default) - This is the original reference behaviour of instance
+  types and preferences where a ControllerRevision is captured and referenced
+  from the VirtualMachine.
+* `expand` - This is a new behaviour where any instance type and preferences are
+  expanded into the VirtualMachine if a ControllerRevision hasn't already been
+  captured and referenced.
+* `expandAll` - The same behaviour as expand but regardless of a
+  ControllerRevision being captured already.
+
+## Concerns
+
+### Exposing users to API complexity within VirtualMachines
+
+One of the original design goals with instance types and preferences was to
+simplify creation by reducing a users exposure to the core
+`VirtualMachineInstanceSpec` API and thus their overall decision matrix.
+
+While this proposal doesn't change the ability to simplify creation of
+`VirtualMachines` it can result in a fully flattened `VirtualMachine` exposing
+all of this complexity at that level once again.
+
+### Breaking declarative management of VirtualMachines using instance types
+
+This expansion behavior will break any declarative management of these
+`VirtualMachines` as they will substantially change after initial submission.
+VM owners will need to explicitly request to not expand their `VirtualMachines`
+referencing instance types or preferences to avoid this.
+
+## Alternatives
+
+### Immutable instance types and preferences
+
+The need to retain point in time revisions of instance types and preferences is
+due to the simple fact that in the current implementation these resources are
+mutable and can change over time. Thus to ensure we always get the same
+`VirtualMachineInstance` at runtime revisions need to be taken and referenced
+from the `VirtualMachine`.
+
+We could possibly remove this requirement by making these object immutable and
+thus dropping the need capture and reference `ControllerRevisions` from the
+`VirtualMachines` at all.
+
+This however still retains the need for additional logic in more complex
+`VirtualMachine` lifecycle operations where we need to expand these now
+immutable resources in the `VirtualMachine`.
+
+### Expand by default and deprecate revision references
+
+We could alter the default `policy` to `expandAll` and in doing so deprecate the
+revision `reference` behaviour for eventual removal from the project ahead of
+`instancetype.kubevirt.io` finally making it to `v1`.
+
+### Keep existing behaviour
+
+Ultimately we can also decide not to implement the core proposal or any of the
+above alternatives and continue to support the original revision based flows.
+Supporting users and third-party integrators with better documentation and
+tooling for VirtualMachines referencing instance types or preferences.
+
+## API Examples
+
+The default `policy` will be `reference` and as such there should be no change
+in behavior when this configurable is not provided.
+
+A cluster admin can default the policy of all new `VirtualMachines` by setting
+`expand` within the `KubeVirt` `CR`:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+metadata:
+  name: kv
+spec:
+  configuration:
+    instancetype:
+      referencePolicy: expand
+```
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: example
+spec:
+  instancetype:
+    name: foo
+  preference:
+    name: bar
+```
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: example
+spec:
+  template:
+    spec:
+      domain:
+        cpu:
+          sockets: 1
+          cores:   1
+          threads: 1
+```
+
+A cluster admin can also use the `expandAll` policy to have all VirtualMachines
+expanded regardless of `revisionNames` already being captured.
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+metadata:
+  name: kv
+spec:
+  configuration:
+    instancetype:
+      referencePolicy: expandAll
+```
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: example
+spec:
+  instancetype:
+    name: foo
+    revisionName: revision-foo
+  preference:
+    name: bar
+    revisionName: revision-bar
+```
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: example
+spec:
+  template:
+    spec:
+      domain:
+        cpu:
+          sockets: 1
+          cores:   1
+          threads: 1
+```
+
+```go
+// KubeVirtConfiguration holds all kubevirt configurations
+type KubeVirtConfiguration struct {
+[..]
+    // Instancetype configuration
+    Instancetype *InstancetypeConfiguration `json:"instancetype,omitempty"`
+}
+
+type InstancetypeConfiguration struct {
+ // ReferencePolicy defines how an instance type or preference should be referenced by the VM after submission, supported values are:
+ // reference (default) - Where a copy of the original object is stashed in a ControllerRevision and referenced by the VM.
+ // expand - Where the instance type or preference are expanded into the VM during submission with references removed.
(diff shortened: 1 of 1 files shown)


design-proposals/instancetype.kubevirt.io/expand-during-submission.md
```diff
@@ -0,0 +1,105 @@
...
+* Provide simple configurables at the cluster and VirtualMachine level to expand instance types and preferences during submission
+
+## Non Goals
+
+* The default behavior will not change as part of this work
+
+## User Stories
+
+* As a cluster owner I want to expand all new VirtualMachines referencing instance types and preferences during submission  
+* As a VirtualMachine owner I want to optionally expand my VirtualMachine referencing an instance type or preference
```
0xFelix, line 24: I think both user stories should state that it is an option that can be enabled?

> As a cluster owner I want to configure whether all new VirtualMachines referencing instance types and preferences are expanded during submission

> As a VirtualMachine owner I want to choose whether my VirtualMachine referencing an instance type or preference is expanded during submission

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
```diff
@@ -0,0 +1,105 @@
...
+    Expand *bool `json:"expand,omitempty"`
+}
+
+```
+
+## Scalability
+
+The mutation and validation webhooks both already expand instance types and preferences into a local copy of the `VirtualMachine` during submission.
+
+The additional call to `Update` the `VirtualMachine` will introduce additional load across the control plane.
```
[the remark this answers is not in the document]
lyarwood: Yeah given the size of the patch I don't think that would get us anything.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
```diff
@@ -0,0 +1,301 @@
...
+
+While this proposal doesn't change the ability to simplify creation of
+`VirtualMachines` it can result in a fully flattened `VirtualMachine` exposing
+all of this complexity at that level once again.
+
+### Breaking declarative management of VirtualMachines using instance types
+
+This expansion behavior will break any declarative management of these
+`VirtualMachines` as they will substantially change after initial submission.
+Users will need to explicitly request to not expand their `VirtualMachines`
```
dankenigsberg, line 86:
```suggestion
VM owners will need to explicitly request to not expand their `VirtualMachines`
```

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
```diff
@@ -0,0 +1,301 @@
...
+
+A possible alternative to this design proposal is to have the VirtualMachine
+controller eventually expand all VirtualMachines referencing instance types or
+preferences regardless of a revision already being stored.
+
+This behaviour could be controlled by the same cluster-wide and VirtualMachine
+specific configurables as suggested by this proposal.
+
+This could however result in drastic changes to existing VirtualMachine objects
+without user interaction.
```
dankenigsberg, line 101: Not sure I follow: a user interaction took place if the configurable was set to `expand`.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
```diff
@@ -0,0 +1,301 @@
...
+
+We would also still need to retain webhook validation logic to ensure that a
+given `VirtualMachine` using an instance type or preference is valid during
+submission.
```
dankenigsberg, line 105: I find it recommended for KubeVirt to move the validation logic from the webhook into the controller. Validating VM definition against cluster-level objects such is instance types or preferences is raceful by design and [breaks eventual consistency](https://docs.google.com/presentation/d/1EIuMwabaOcBw6AqcZCqz8Mip0HuhOcJtOsqVV0lVbEM/).

But I do not see how this is related to the design proposal. Can you elaborate?

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
```diff
@@ -0,0 +1,301 @@
...
+
+This would also put additional load on the controller if the cluster-wide
```
dankenigsberg, line 107:
```suggestion
This would also put one-off load on the controller if the cluster-wide
```
I think that on-startup load is reduced by approximately same amount. This option just makes it happen sooner.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
```diff
@@ -0,0 +1,301 @@
...
+  name: cirros
+spec:
+  instancetype:
+    name: foo
+    revisionName: bar
+```
+
+A user can explicitly ask for the associated instance type to be expanded into
+the `VirtualMachine` by removing `revisionName` and setting `policy` to
+`expand`:
```
dankenigsberg, line 208: Isn't adding `policy:expand` explicit enough? Why removing `revisionName` is required?

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
```diff
@@ -0,0 +1,301 @@
...
+## Update/Rollback Compatibility
+
+Existing `VirtualMachines` referencing instance types or preferences will not be
+changed as a result of this behaviour.
+
+There will be no ability to automatically rollback new `VirtualMachines` once
+they have their instance type or preference expanded by this new functionality.
+
+Users will also be unable to resize their `VirtualMachines` by making a singular
+choice of instance type in the future.
```
dankenigsberg, line 293: Why not? It is still OK to define instance-typed VM `spec`s, why would it not be ok to modify an existing VM to be instance typed? In properly eventual-consistent systems, the intention of the user is what matters most, not the history of the resource.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
```diff
@@ -0,0 +1,301 @@
...
+* kubevirt/kubevirt
+
+## Design
+
+A new `KubeVirt` configurable will be introduced to have `VirtualMachines`
+referencing instance types or preferences expanded.
+
+Additional `VirtualMachine` configurables will also be introduced to allow a
+specific newly submitted VirtualMachine referencing instance types or
+preferences to be expanded.
```
dankenigsberg, line 55: Can we consider alternatives where this is not available? My question stems form me not being aware of the second user story (a VM owner wanting to control this). I heard only of cluster admins.

dankenigsberg (changes requested):

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
lyarwood: I can capture that in alternatives but personally I think that VM Owners should have the ability to expand or not regardless of the admins choice.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
lyarwood: I'll reword to explicitly state that this can happen without the VM owner interacting with the VM thanks.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
lyarwood: I understand but while there are still validations taking place in the webhooks we will need to expand any provided instance type and preference to retain the same behaviour as a flat VirtualMachine. Once these validations are moved into the control loop then yes the need for this is also removed.

I wanted to mention this in passing as it's logic we would still need to carry in the project that isn't removed by eventually expanding later.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
lyarwood: ACK yes fair point, changed.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
lyarwood: The current proposal stops short of expanding all VirtualMachines and would use `revisionName` to determine if the VirtualMachine had already been seen by the controller and revisions stored. If we went with the alternative of expanding all regardless of a revision already being captured then `policy` being set to `expand` would be enough to trigger the behaviour.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
lyarwood: At present instance types conflict if the VirtualMachine already provides an amount of vCPU or memory.

We have talked about changing this behaviour in https://github.com/kubevirt/community/pull/318 and in doing so could allow a user to resize flat VirtualMachines by selecting an instance type but that might require a follow up design proposal and discussion to work out the details.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
dankenigsberg: Thanks. But if I understand correctly, the biggest issue today is the complexity for third parties. If each VM owner can choose whatever they want, the complexity remains.

I would prefer have an opinionated solution, with less knobs and options; if we see a use case for the VM owners having this freedom to work against their cluster admin, we can add per-VM knob later. Removing knobs is practically impossible.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
dankenigsberg: does the following represents what you are saying?
```suggestion
Note that this proposal does not affect the validation logic to ensure that a
given `VirtualMachine` using an instance type or preference is valid during
submission.
```

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
dankenigsberg: I do not follow... Assume that I created a flat VM with 7 CPUs. I stopped my VM and then changed my mind. I would like it to be much larger and more standard and edit the VM to use `cx1.8xlarge`. Would KubeVirt attempt to block this? Why?

dankenigsberg (changes requested):

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
lyarwood: At present yes it would unless you also removed the 7 vCPUs you originally requested within the VirtualMachine.

We just don't allow you to request vCPUs and memory through the VirtualMachineInstanceSpec *and* instance type at the same time.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
lyarwood: That complexity will remain still even if this configurable is limited to cluster admins surely, especially given that we are not changing the default here.

Making this a cluster admin only configurable will mean that the end user experience of using a VirtualMachine initially defined with instance types and preferences can change drastically between environments and there's nothing a user can do to avoid this.

By providing an additional VM owner configurable they can be explicit and know that the resulting VirtualMachine will behave as expected.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
dankenigsberg: > At present yes it would unless you also removed the 7 vCPUs you originally requested within the VirtualMachine.
> 
> We just don't allow you to request vCPUs and memory through the VirtualMachineInstanceSpec _and_ instance type at the same time.

this is clear and expected. I just don't understand it from "Users will also be unable to resize their `VirtualMachines` by making a singular choice of instance type in the future". Would you agree to rephrase the text?

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
dankenigsberg: > Users will also be unable to resize their `VirtualMachines` by making a singular
choice of instance type in the future

Why? A cluster admin that would like to interact with a third party can configure `expand` on her cluster, and that's that. If we complicate the solution and let the VM owner choose his own semantic, the cluster admin is left with uncertainty.

> Making this a cluster admin only configurable will mean that the end user experience of using a VirtualMachine initially defined with instance types and preferences can change drastically between environments and there's nothing a user can do to avoid this.

Correct. And that's exactly what I, as a cluster admin, want. I need to interact with a third party that is unaware of instance types, so I want my cluster to comply.

> By providing an additional VM owner configurable they can be explicit and know that the resulting VirtualMachine will behave as expected.

Right, but this adds complexity to the code, which I am not aware was requested by any user, increases development time, expands test matrix and add possibility of bugs, and cannot be dropped out of the API in the future.

lyarwood: Okay these are fair points, I'll remove the user configurables for now and we can always follow up in the future if requested.

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
lyarwood: Ah apologies I see your point, I'll rephrase now.

lyarwood:
It does help if I push to the correct branch when attempting to update a PR, apologies!

design-proposals/instancetype.kubevirt.io/expand-during-submission.md
```diff
@@ -0,0 +1,255 @@
...
+  instance types and preferences are referenced from a VirtualMachine
+
+## Non Goals
+
+* The default behavior will not change as part of this work
+
+## User Stories
+
+* As a cluster owner I want to control how instance types and preferences are
+  referenced from VirtualMachines within my environment
```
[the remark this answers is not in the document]
lyarwood: The suggested configurable still allows a VirtualMachine to reference an instance type and preference so to me it doesn't make sense to focus on expansion here.

vladikr:
/approve
I went over this again. It reflects the reality as the implementation has already been merged.

@dankenigsberg do you still have concerns about the current text?

dankenigsberg:
> /approve I went over this again. It reflects the reality as the implementation has already been merged.
> 
> @dankenigsberg do you still have concerns about the current text?

Yes - I don't like the fact that we have an `expand` option. We needed only `expandAll`.
But unfortunately the code was merged with this option, so it does not matter any more.

/lgtm

to reflect reality.

PR #329, opened 2024-09-30
Removing Alona, cwilkers, & jobbler from reviewer/approver

aburdenthehand:
Sadly @AlonaKaplan  @cwilkers  and @jobbler are no longer active in the project. It has been a bit over 6 months since their last contribution and we should remove them from our reviewer/approver list.

Thank you for your past contributions, and for your help in making this project what it is today. 

(Also minor housekeeping: moving jed to alphabetical order)

I'm using the same criteria that we apply for [project maintainers](https://github.com/kubevirt/community/blob/main/GOVERNANCE.md). As discussed in this PR, we should document a process and criteria for retiring reviewers and approvers.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,19 +1,15 @@
 reviewers:
   - aburdenthehand
-  - cwilkers
   - dhiller
   - jean-edouard
-  - jobbler
 
 approvers:
   - aburdenthehand
-  - AlonaKaplan
-  - cwilkers
   - davidvossel
   - fabiand
+  - jean-edouard
   - rmohr
   - vladikr
-  - jean-edouard
 
 emeritus_approvers:
   - karmab # 10 july 2019
@@ -26,3 +22,5 @@ emeritus_approvers:
   - alosadagrande # April 2020
   - codificat # Jan 2022
   - mazzystr # Jan 2022
+  - AlonaKaplan # Sep 2024
+  - cwilkers # Sep 2024


jean-edouard (commented):
/cc @AlonaKaplan @cwilkers @jobbler
Please take a second to LGTM if you see this

EdDev:
> @EdDev I am okay with a script handling this in the future however to the best of my knowledge the only script we have today targets members in the org after 12 months of inactivity and removes them. I'm not sure if a person removed from the org would affect their being tagged as a reviewer based on these owners files.

I think it removed them from the org, based on the CNCF policy of inactivity.

If we remove specific members, we need to mention what policy exactly we follow, how we detect these members and what are the steps needed to politely retire them.

But if there is an agreement on all of the details, don't let me be in the way.

aburdenthehand:
This doesn't remove them from the org, just as reviewers and approvers with special privileges. They will presumably be removed after 12 months of inactivity but at the moment they are inactive in the community and being tagged on reviews which limits visibility on issues/prs.
I'm using the same criteria that we apply for [project maintainers](https://github.com/kubevirt/community/blob/main/GOVERNANCE.md) but I agree that it would be worth being explicit about how we remove inactive reviewers and approvers in the future. Looking through the history of this file it seems as though there is a general agreement that when someone leaves the project they are removed from these roles.

OWNERS
```diff
@@ -1,19 +1,15 @@
...
 
 approvers:
   - aburdenthehand
-  - AlonaKaplan
-  - cwilkers
   - davidvossel
   - fabiand
+  - jean-edouard
   - rmohr
   - vladikr
```
iholder101, line 12: nit: this is redandunt

iholder101 (commented):
This makes sense to me.
@dhiller's script is valuable, but I don't think we've committed to rely solely on it. If we know for a fact that approvers/reviewers are no longer active, opening such a PR is the right choice IMO. In addition, these members aren't being completely removed, but being moved to `emeritus_approvers` which tracks the fact that they were once active approvers and opens the door to come back again in the future.

@AlonaKaplan @cwilkers @jobbler - this PR is opened for a while now, if you have any objections please let us know.

From my perspective this is good to go.
/lgtm

EdDev:
> I'm using the same criteria that we apply for [project maintainers](https://github.com/kubevirt/community/blob/main/GOVERNANCE.md) but I agree that it would be worth being explicit about how we remove inactive reviewers and approvers in the future. Looking through the history of this file it seems as though there is a general agreement that when someone leaves the project they are removed from these roles.

As long as we can ref a policy like the one you shared and the means (I guess a github query?) that have been taken to identify these personel, we should be good. I thought this is exactly what @dhiller script does.
If we do not share the query or follow a specific script outcome, it becomes hard to assure the names are picked correctly.

@aburdenthehand , can you please share the query/script that identified the members and ref the policy you mentioned above in the PR description?

aburdenthehand:
@EdDev I've added the link to the description. 
These were not identified by a script but rather from me hearing that they had largely moved on from kubevirt, and observing that they were no longer reviewing or approving PRs to which they were being assigned. 
To my knowledge our devstats metrics do not allow us to silo contributions specifically to this repo; iirc I went through their account activity.

OWNERS
aburdenthehand: Respectfully disagree - it returns the list to an alphabetised state.

dhiller:
@EdDev as stated by @aburdenthehand I confirm that the script I wrote [1] only cleans up the org itself. We have a blind spot in detecting inactive reviewers and approvers on repos though. After the **broken devstats for KubeVirt are repaired** [2] we might be able to leverage them in a similar way.

[1]: https://github.com/kubevirt/project-infra/blob/main/hack/remove-users-without-contributions.sh
[2]: https://github.com/kubevirt/community/pull/332

dhiller:
To emphasize again: even the org cleanup can't work currently since devstats for kubevirt are broken.

iholder101:
> Personally I am not comfortable with this criteria. I would prefer an actual script or filter to confirm a specific user is no longer active for a pre-defined period of time.

My 2 cents on this:
@EdDev I absolutely agree with you that this would be preferred. Ideally, it would be great if we had the bot automatically create such PRs whenever it detects inactivity.

However, since we currently don't have this automation nor a clear policy on how to remove members, I think this PR makes sense. And in general, any policy and automation have their exceptional cases, and nothing can replace human judgement. If we know for a fact that members are inactive and not going to be active soon, be very vocal about wanting to remove them, give them enough time to object, and moving them to `emeritus_approvers` with a chance to get back to being approvers/reviewers, I think it's enough. I would even say this process should be a part of the removal policy as an alternative to the automatic approach.

iholder101:
> Pull requests that are marked with `lgtm` should receive a review from an approver within 1 week.
> 
> After that period the bot marks them with the label `needs-approver-review`.
> 
> /label needs-approver-review

ping @aburdenthehand @jean-edouard

jean-edouard:
@EdDev do you still have concerns about this?
Affected users were pinged more than a month ago, I think this is ready to be approved.

EdDev:
> @EdDev do you still have concerns about this? Affected users were pinged more than a month ago, I think this is ready to be approved.

I have a problem with not providing "proof" and picking random users.
If you give me "proof" like [here](https://github.com/kubevirt/kubevirt/pull/13249), then I will be ok with this.

vladikr:
> > @EdDev do you still have concerns about this? Affected users were pinged more than a month ago, I think this is ready to be approved.
> 
> I have a problem with not providing "proof" and picking random users. If you give me "proof" like [here](https://github.com/kubevirt/kubevirt/pull/13249), then I will be ok with this.

yeah, we need to do a better job showing why we are removing users. 

I wrote this quickly today (so no judging ;) ) https://gist.github.com/vladikr/29ff700eefd1f9388d2268c51461236b
@dhiller perhaps you can improve this?

running this:

```
python inactive_members.py --repo kubevirt/community --auth_token [token]
All inactive members: ['mazzystr', 'codificat', 'iranzo', 'prnaraya', 'jparrill', 'jberkus', 'suomiy', 'scollier', 'ptrnull', 'fgimenez', 'AlonaKaplan', 'phoracek', 'tripledes', 'ShellyKa13', 'yasker', 'vasiliy-ul', 'starbops', 'akalenyu', 'alromeros', 'chandrajr', 'cwilkers', 'daghaian', 'dependabot[bot]', 'salanki', 'maya-r', 'maryshak1996', 'ksimon1', 'jseguillon', 'hrittikhere', 'doron-fediuck', 'dbbhat', 'xmulligan', 'alexpilotti', 'adithyaakrishna']
All inactive reviewers: ['cwilkers', 'jobbler']
All inactive approvers: ['AlonaKaplan', 'cwilkers']
```

aburdenthehand:
@EdDev not sure why you think it was random but until we have a better alternative do you accept the output from @vladikr's script that these users are in fact inactive so we can remove them and prevent them being tagged on new PRs?

vladikr:
Ok, I'd like to get it in. I provided evidence that this change is not random. 
/approve

PR #331, opened 2024-10-07
DRA in Kubevirt

alaypatel07:

Fixes #

**Checklist**

**Release note**:

```release-note

```

### DIFF
--- a/design-proposals/dra.md
+++ b/design-proposals/dra.md
@@ -1,83 +1,344 @@
 # Overview
+
 This proposal is about adding supporting DRA (dynamic resource allocation) in KubeVirt.
-DRA allows vendors fine-grained control of devices.  The device-plugin model will continue
+DRA allows vendors fine-grained control of devices. The device-plugin model will continue
 to exist in Kubernetes, but DRA will offer vendors more control over the device topology.
 
 ## Motivation
+
 DRA adoption is important for KubeVirt so that vendors can expect the same
 control of their devices using Virtual Machines and Containers.
 
 ## Goals
+
+- Align on the API changes needed to consume DRA enabled devices in KubeVirt
 - Align on how KubeVirt will consume external and in-tree DRA drivers
 - Align on what drivers KubeVirt will support in tree
-- Align on the API changes needed to consume DRA enabled devices in KubeVirt
+TODO: @alayp Q: Should the decision around in-tree drivers be in scope for this proposal?
 
 ## Non Goals
+
 - Replace device-plugin support in KubeVirt
 
 ## Definition of Users
+
 - A user is a person that wants to attach a device to a VM
 - An admin is a person who manages the infrastructure and decide what kind of devices will be exposed via DRA
-- An developer is a person familiar with CNCF ecosystem that can develop automation using the APIs discussed here  
+- A developer is a person familiar with CNCF ecosystem that can develop automation using the APIs discussed here
+
+## Assumptions
+
+- KubeVirt will support DRA API from kubernetes version 1.31.
+- The example demonstrate a pgpu driver, but the same mechanism could be used for vgpu
+TODO: @alayp Q: Can device plugins and DRA drivers co-exists in the same cluster?
 
 ## User Stories
+
 - As a user, I would like to use my GPU dra driver with KubeVirt
 - As a user, I would like to use KubeVirt's default driver
 - As a developer, I would like APIs to be extensible so I can develop drivers/webhooks/automation for custom use-cases
 
+## Use Cases
+
+### Supported Usecases
+
+1. Devices where the DRA driver that set agreed upon attributes for resources in ResourceSlice.
+   1. PCIe Bus address for pGPU
+   2. Device uuid for vGPU
+1. Devices have to either be of the type GPU or HostDevices. 
+
+### Unsupported Usecases
+
+1. Devices where the DRA driver does not set the attributes needed to configure libvirt dom XML for the devices will not
+   be supported.
+   1. In the future a standardization could be envisioned for a PCI device where the attributes are set
+      automatically through the DRA framework. When this is achieved, any DRA device should be available in KubeVirt VMs
+
 ## Repos
+
 kubevirt/kubevirt
 
 # Design
 
-At the high level there are two sets of changes needed for integrating DRA with KubeVirt:
-1. How DRA resources can be consumed by users of KubeVirt?
-2. What changes are needed for existing device-plugin drivers to be DRA compatible?
-
-This design focuses on part 1 of the problem. 
-
 For allowing users to consume DRA devices, there are two main changes needed:
-1. API changes
-2. Driver Implementation
+
+1. API changes and the plumbing required in KubeVirt to read env var and convert it into right domxml
+2. Driver Implementation to set the env var
+
+This design focuses on part 1 of the problem.
 
 ## API Changes
 
 ```
 type VirtualMachineInstanceSpec struct {
+    ..
+    ..
     // ResourceClaims defines which ResourceClaims must be allocated
-    // and reserved before the virt-launcher pod and hence the VMI is allowed to start. The resources
-    // will be made available to the domain which consume them
-    // by name.
-    //
-    // This is an alpha field and requires enabling the
-    // DynamicResourceAllocation feature gate in kubernetes
-    //  https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/
-    //
-    // This field is immutable.
-    //
-    // +featureGate=DynamicResourceAllocation
-    ResourceClaims []k8sv1.PodResourceClaim `json:"resourceClaims,omitempty"`
+	// and reserved before the VMI and hence virt-launcher pod is allowed to start. The resources
+	// will be made available to the domain which consume them
+	// by name.
+	//
+	// This is an alpha field and requires enabling the
+	// DynamicResourceAllocation feature gate in kubernetes
+	//  https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/
+	//
+	// This field is immutable.
+	//
+	// +listType=map
+	// +listMapKey=name
+	// +optional
+	ResourceClaims []k8sv1.PodResourceClaim `json:"resourceClaims,omitempty"`
 }
 
 type GPU struct {
 	// Name of the GPU device as exposed by a device plugin
+	Name string `json:"name"`
+	// DeviceName is the name of the device provisioned by device-plugins
+	DeviceName string `json:"deviceName,omitempty"`
+	// Claim is the name of the claim that is going to provision the DRA device
+	Claim             *k8sv1.ResourceClaim `json:"claim,omitempty"`
+	VirtualGPUOptions *VGPUOptions         `json:"virtualGPUOptions,omitempty"`
+	// If specified, the virtual network interface address and its tag will be provided to the guest via config drive
+	// +optional
+	Tag string `json:"tag,omitempty"`
+}
+
+type HostDevice struct {
+	Name string `json:"name"`
+	// DeviceName is the name of the device provisioned by device-plugins
+	DeviceName string `json:"deviceName,omitempty"`
+	// Claim is the name of the claim that is going to provision the DRA device
+	Claim *k8sv1.ResourceClaim `json:"claim,omitempty"`
+	// If specified, the virtual network interface address and its tag will be provided to the guest via config drive
+	// +optional
+	Tag string `json:"tag,omitempty"`
+}
+
+type VirtualMachineInstanceStatus struct {
+    ..
+    ..
+	// DeviceStatus reflects the state of devices requested in spec.domain.devices. This is an optional field available
+	// only when DRA feature gate is enabled
+	// +optional
+	DeviceStatus *DeviceStatus `json:"deviceStatus,omitempty"`
+}
+
+// DeviceStatus has the information of all devices allocated spec.domain.devices
+// +k8s:openapi-gen=true
+type DeviceStatus struct {
+	// GPUStatuses reflects the state of GPUs requested in spec.domain.devices.gpus
+	// +listType=atomic
+	// +optional
+	GPUStatuses []DeviceStatusInfo `json:"gpuStatuses,omitempty"`
+	// HostDeviceStatuses reflects the state of GPUs requested in spec.domain.devices.hostDevices
+	// DRA
+	// +listType=atomic
+	// +optional
+	HostDeviceStatuses []DeviceStatusInfo `json:"hostDeviceStatuses,omitempty"`
+}
+
+type DeviceStatusInfo struct {
+	// Name of the device as specified in spec.domain.devices.gpus.name or spec.domain.devices.hostDevices.name
+	Name string `json:"name"`
+	// DeviceResourceClaimStatus reflects the DRA related information for the degive
+	DeviceResourceClaimStatus *DeviceResourceClaimStatus `json:"deviceResourceClaimStatus,omitempty"`
+}
+
+// DeviceResourceClaimStatus has to be before SyncVMI call from virt-handler to virt-launcher
+type DeviceResourceClaimStatus struct {
+	// ResourceClaimName is the name of the resource claims object used to provision this resource
+	// +optional
+	ResourceClaimName *string `json:"resourceClaimName,omitempty"`
+	// DeviceName is the name of actual device on the host provisioned by the driver as reflected in resourceclaim.status
+	// +optional
+	DeviceName *string `json:"deviceName,omitempty"`
+	// DeviceAttributes are the attributes published by the driver running on the node in
+	// resourceslice.spec.devices.basic.attributes. The attributes are distinguished by deviceName
+	// and resourceclaim.spec.devices.requests.deviceClassName.
+	// +optional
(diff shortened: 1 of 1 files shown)


design-proposals/dra.md
```diff
@@ -0,0 +1,529 @@
...
+  - name: pgpu-claim-name
+    source:
+      resourceClaimTemplateName: pgpu-claim-template
+  domain:
+    devices:
+      gpus:
+      - name: pgpu
+        claim: 
+          name: pgpu-claim-name
+          request: pgpu-request-name
```
alicefr, line 228: I don't quite understand why we cannot simply reference the name of the resourceClaim, it is done in the same way for the volumes/disks. Why do we need an additional level with the name and the request?

design-proposals/dra.md
```diff
@@ -0,0 +1,529 @@
...
+	ResourceClaimTemplateName string `json:"resourceClaimTemplateName"`
+}
+
+```
+
+This design misses the use-case where more than one DRA device is specified in the claim template, as each
+device will have its own template in the API.
+
+This design also assumes that the deviceName will be provided in the ClaimParameters, which requires the DRA drivers
+to have a ClaimParameters.spec.deviceName in their spec.
```
alicefr, line 520: From this description you are distinguishing the `name` and the `deviceName` which also refer to my previous questions. Maybe it will be clearer to add the difference between the 2 at the beginning

design-proposals/dra.md
```diff
@@ -0,0 +1,529 @@
...
+1. virt-controller needs a mechanism to watch for virt-launcher pods, resourceclaims and resourceslices to populate the 
+   vmi.status.deviceStatus using the steps mentioned in above section that has all the attributes (for example the 
+   pciAddress for the gpu device)
+
+### Virt launcher changes
+
+1. For devices generated using DRA, virt-launcher needs to use the vmi.status.deviceStatus to generate the domxml
+   instead of environment variables as in the case of device-plugins
+
+# Alternate Designs
```
alicefr, line 467: I tend to prefer the first design since it follow the same schema as the volumes/disks and the network/interfaces where we have the section for the pod resource and how they are mapped to the VM

design-proposals/dra.md
```diff
@@ -0,0 +1,529 @@
...
+
+1. If devices are requested using DRA, virt controller needs to render the virt-launcher manifest such that 
+   pod.spec.resourceClaims and pod.spec.containers.resources.claim sections are filled out.
+1. virt-controller needs a mechanism to watch for virt-launcher pods, resourceclaims and resourceslices to populate the 
+   vmi.status.deviceStatus using the steps mentioned in above section that has all the attributes (for example the 
+   pciAddress for the gpu device)
+
+### Virt launcher changes
+
+1. For devices generated using DRA, virt-launcher needs to use the vmi.status.deviceStatus to generate the domxml
```
alicefr, line 464: +1

design-proposals/dra.md
```diff
@@ -0,0 +1,529 @@
...
+   instead of environment variables as in the case of device-plugins
```
alicefr, line 465: +1 for avoiding the env variable, I like the status!

alicefr:
@alaypatel07 the proposal looks great. 
One question. You mention in the goal section to allow external DRA drivers. Do we want to have a mechanism to enable certain dra driver or is this unnecessary?  For CSI storge classes we don't have, but for the external device plugin we do.

design-proposals/dra.md
alicefr, line 465: How do we populate the deviceStatus inormation with an external DRA driver?

iholder101:
/cc

Please add to the PR description that it's a continuation of https://github.com/kubevirt/community/pull/293.

design-proposals/dra.md
```diff
@@ -0,0 +1,529 @@
...
+	// +listMapKey=name
+	// +optional
+	ResourceClaims []k8sv1.PodResourceClaim `json:"resourceClaims,omitempty"`
+}
+
+type GPU struct {
+	// Name of the GPU device as exposed by a device plugin
+	Name string `json:"name"`
+	// DeviceName is the name of the device provisioned by device-plugins
+	DeviceName string `json:"deviceName,omitempty"`
```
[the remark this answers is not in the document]
alaypatel07: @alicefr This is not a new API being added, it is how the existing device-plugins work.  The idea is that if DeviceName is populated, then its a device provisioned by device-plugins, if the Claim field below is not nil then its a DRA device. We have an alternate API suggestion that will make this choice to the user much more obvious.

design-proposals/dra.md
alaypatel07: @alicefr as far as this proposal concerns, the assumption is that drivers publish information needed to generate the right domxml for the device. For instance in case of a gpu, it could either be pciAddress(pGPU), or mdev uuid (vGPU).  The drivers need to publish this information in ResourceSlice as attributes https://pkg.go.dev/k8s.io/api@v0.31.0/resource/v1alpha3#BasicDevice.

Virt-controller will find the device for it and put this attribute in the vmi status using the steps mentioned here: https://github.com/kubevirt/community/blob/30a2005d8ef72cbae0d8fb1d818861b14d8a5d88/design-proposals/dra.md#dra-api-for-reading-device-related-information

So there could be two cases here:
- either the device has standard attributes, like pciAddress and uuid which will be supported in tree or
- the device needs additional attributes, for which a sidecar will be needed to read that attribute from vmi status and define it in the domxml

alaypatel07:
> You mention in the goal section to allow external DRA drivers. Do we want to have a mechanism to enable certain dra driver or is this unnecessary? For CSI storge classes we don't have, but for the external device plugin we do.

Answered it inline, the proposal will implement the conversion logic of standard attributes in vmi status (pciAddress and uuid) for other attributes, a extension in form of side car is needed.

design-proposals/dra.md
```diff
@@ -0,0 +1,541 @@
...
+2. allowing the devices to be used as a host device (spec.domain.device.hostDevices)
+
+The status section of the VMI will contain information of the allocated devices for the VMI when the information is
+available in DRA APIs. The same information will be accessible in virt-handler, virt-launcher and sidecar containers. 
+This allows for device information to flow from DRA APIs into KubeVirt stack. 
+
+The virt-launcher will have the logic of converting a GPU device into its corresponding domxml. For use-cases that are 
+not handled in-tree, a sidecar container could be envisioned which will convert the information available in status to 
+the corresponding domxml.
+
```
vladikr, line 185: 1. Could you please provide a specific use case for the proposed use of a sidecar? It would also help to understand why this configuration cannot be handled directly in the virt-launcher itself.
2. Currently, a sidecar can't access the VMI object without changes to the virt-controller. How do you plan to differentiate between DRA-related sidecars and other sidecars?
3. I assume that the proposed sidecar would rely on KubeVirt's existing hooking mechanism. It's worth noting that this mechanism is provided on a best-effort basis and is not fully supported. It uses gRPC with a defined API, and rather than adding DRA-specific data to the VMI status, a new API call could be integrated into the existing gRPC used by virt-launcher. This way, the DRA-specific data could be handled only between virt-launcher and the sidecar and avoid replication of the DRA data across multiple components...
4. Building a solution based on the current hooking mechanism, or modifying the libvirt XML outside of virt-launcher, is not recommended. The libvirt XML is not a supported external API—only the KubeVirt API is. Please carefully consider how to maintain API compatibility between DRA, libvirt, and KubeVirt. If supporting a sidecar is necessary, we should develop a stable interface between compute and the DRA sidecar containers.

design-proposals/dra.md
vladikr: From my perspective, there is no need to store driver internal data in the public, user facing,  VMI status.
This information is only required by the virt-launcher internals, no other kubevirt components benefit from this.  
I fail to understand how polluting the VMI status is superior to providing specific data to the virt-launcher's compute container directly.

vladikr: Same way as we don't keep the PVC status in VMI we should keep DRA specific data out of the VMI status.

vladikr: As I mentioned above, please provide more usecases for the sidecar scenario. I've provided an alternative take.

design-proposals/dra.md
```diff
@@ -0,0 +1,541 @@
...
+            int: 0
+          model:
+            string: LATEST-GPU-MODEL
+          uuid:
+            string: gpu-8e942949-f10b-d871-09b0-ee0657e28f90
+          pciAddress: 
+            string: 0000:01:00.0
+        deviceName: gpu-0
+        resourceClaimName: virt-launcher-vmi-fedora-9bjwb-gpu-resource-claim-m4k28
+      name: pgpu
```
vladikr, line 248: At the moment, I am not convinced that storing this information in the VMI status is needed. 
What other Kubevirt components require this information?
Why CDI cannot provide this info directly to the consumer container?

design-proposals/dra.md
alaypatel07: After last week’s discussion and another offline conversation with @xpivarc, its looks like currently we do not have enough use-cases for allowing users to configure these devices in sidecar. It seems like we would like encourage building the support for these devices in the core, rather than allowing them to be configured by a sidecar, I will modify this proposal to change the language to reflect this

design-proposals/dra.md
alaypatel07: As I said in one of the above comments, will be updating the proposal to remove sidecar scenario due to lack of usecases, updated my previous comment to reflect it.

varunrsekar: > From my perspective, there is no need to store driver internal data in the public, user facing, VMI status.
This information is only required by the virt-launcher internals, no other kubevirt components benefit from this.
I fail to understand how polluting the VMI status is superior to providing specific data to the virt-launcher's compute container directly.

This is in part to address an architectural difference between DPs and DRA:
* DPs pre-define a static resource which is usually identified by a device model: Eg: `nvidia.com/GP102GL_Tesla_P40
` . A VMI spec would reference this name directly for its device and it gives easily inferrable information to the user that the VMI will run on a nvidia Tesla P40 GPU.
* In DRA, this information is lost as the resource claim that we provide in our spec masks all details about what kind of device that claim is allocating. A resourceclaim is a dynamic object where you would need to look into its spec to know what piece of hardware you're getting. For the user, this means that they can no longer look into the vmi spec to immediately tell what their VMI is running on.

Given this regression in the user expectation (conscious or unconscious), we should provide atleast the bare-minimum identifiable information about a device like the hardware model etc.

varunrsekar: > Same way as we don't keep the PVC status in VMI we should keep DRA specific data out of the VMI status.

In DRA, once a resource claim is allocated, you'll need to take the following steps to identify what device is allocated:

* look at the vmi spec to see the referenced claim (template or name) and request.
* look at virt-launcher pod status to match the claim from vmi spec and infer the resource claim k8s object
* look at the resource claim k8s object to find out the driver and device names for each request.
* look at the resource slice object for the VMI's node published by the corresponding driver and know all the relevant attributes needed

A PVC needs similar hoops to find the volume information but is a lot less complex than DRA. To tackle this problem, PVC information is currently posted onto the VMI status so consumers (users & kubevirt) can easily find the info about a volume. (See https://pkg.go.dev/kubevirt.io/api/core/v1#PersistentVolumeClaimInfo)

Given that DRA ResourceClaims are modeled after PVCs and we have precedence for publishing volume information in the VMI status to make visilibity/access easier, I believe we should follow the same approach for DRA devices.

alaypatel07: @vladikr the particular case that @varunrsekar mentioned in the comment was discussed in the meeting on 10/15 ([recording](https://drive.google.com/file/d/1RClJoetDy5rE2M9DjZiqv1iNDBcXU92m/view?usp=drive_link)) and documented in detail here: https://docs.google.com/document/d/1bQdLoxwSC1ILvyIb4ljSm5RZmsVpvqfWKIbN0KOBKsw/edit#heading=h.wjmanqk7qzd9

Additionally, we came across another use-case of why the information is needed in vmi-status:

Usecase Title: Debuggability of VMI in failed state with DRA device 
- K8s has a feature of releasing the resource claims once the pod has gone to a terminal state(Failed or Succeeded). This is to allow for expensive resources to be used for other pods
- in case where VMI has failed and the virt-launcher has gone to a Failed state, what hardware was attempted to be provisioned for this VMI is lost

It is documented in depth here: https://docs.google.com/document/d/1bQdLoxwSC1ILvyIb4ljSm5RZmsVpvqfWKIbN0KOBKsw/edit#heading=h.3z09ita4dzuz

Please let use know if this answers the question about why extending the VMI status with this information is useful to the user

alaypatel07:
@alicefr From our discussion in the meeting on 10/15 on lifecycle of resourceclaim and VM/VMI

@varunrsekar and I looked at the nvme implementation and it seems that it was developed on an older version of the DRA apis (v1alpha2). With the newer version (k8s 1.31 and v1alpha3 of the ResourceClaim api), if the pod goes to failed state:
1. if the resourseclaim is being created by template, the resource claim will be deleted
2. if the resourceclaim is created by the user, the resource claim goes from allocated to pending state where the allocations are cleared

Here is the running demo from 1.31 cluster
1. 
```
$ k version
Client Version: v1.31.1
Kustomize Version: v5.4.2
Server Version: v1.31.0
```
2. create the vmi with manual resource claim
```
$ k apply -f demo-with-manual-claim.yaml
namespace/gpu-test1 unchanged
resourceclaim.resource.k8s.io/virt-launcher-vmi-fedora created
Warning: metadata.finalizers: "foregroundDeleteVirtualMachine": prefer a domain-qualified finalizer name to avoid         accidental conflicts with other finalizer writers
virtualmachineinstance.kubevirt.io/vmi-fedora created
```
        
3. get the status of virt-launcher pod to check the claim it used:
```
$ k get pods -n gpu-test1 -oyaml| grep resourceClaimName
      resourceClaimName: virt-launcher-vmi-fedora
```
4. check the pod is in running state
```
$ k get pods -n gpu-test1 -oyaml | grep phase
    phase: Running
```
5. check the resource claim status:
```
$ k get resourceclaim -n gpu-test1 -oyaml | grep -A 5 status
  status:
    allocation:
      devices:
        config:
        - opaque:
            driver: gpu.nvidia.com
```
6. wait for pod to go to failed state
```
$ k get pods -n gpu-test1 -oyaml -w | grep phase
  phase: Running
  phase: Running
  phase: Failed
```
7. check the resourceclaim status:
```
$ k get resourceclaim -n gpu-test1 -oyaml | grep -A 5 status
  status: {}
```

This is further seen in the KEP: https://github.com/kubernetes/enhancements/pull/4709/files#diff-8fa237d276346416c2aafa209f721e43c9762f59cabec234eafafe01694de3eeR1221 

I wonder if the use-cases that require the lifecycle of the ResourceClaim not attached to the lifecycle of the pod, be deferred to the future versions of DRA where it is possible for the allocation to persist after pod is deleted.

alaypatel07:
@rthallisey done, I have couple of updates to be pushed from our last sync up. Will plan on pushing those soon.

alicefr:
> I wonder if the use-cases that require the lifecycle of the ResourceClaim not attached to the lifecycle of the pod, be deferred to the future versions of DRA where it is possible for the allocation to persist after pod is deleted.

@alaypatel07 sorry I missed the ping. I'm fine to tackle all the use case gradually. It needs just to be properly documented

design-proposals/dra.md
```diff
@@ -0,0 +1,703 @@
...
+metadata:
+  name: pci-nvme-claim-template
+spec:
+  spec:
+    devices:
+      requests:
+        - name: pci-nvme-request-name
+          deviceClassName: pci-nvme.kubevirt.io
+          allocationMode: ExactCount
+          count: 1
```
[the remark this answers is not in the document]
alaypatel07: @alicefr with the alpha3 API the count field is set by default, ref: https://github.com/kubernetes/api/blob/b7783abfc99c11f3b56fee4e5cf99023fcc8120a/resource/v1alpha3/types.go#L446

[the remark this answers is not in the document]
varunrsekar: And it has carried over to the beta APIs too: https://github.com/kubernetes/api/blob/master/resource/v1beta1/types.go#L457 :)

design-proposals/dra.md
```diff
@@ -0,0 +1,703 @@
...
+---
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  labels:
+    kubevirt.io/vm: vm-cirros
+  name: vm-cirros
+spec:
+  running: false
```
RamLavi, line 311: nit: please use the new API: `runStrategy: Halted`

alaypatel07:
@xpivarc @vladikr @alicefr @rthallisey Updated the document as per discussion in "DRA Support sync" on Feb 18th. 

Can I have another rounds of reviews on this?

akalenyu:
Sorry to jump in so late. I might have mentioned this in some other PR (or closed forums)

Basically, with the way volume hotplug (specifically, block volume hotplug) is implemented today, I fear that we are going to lose DRA-added devices in the cgroupsv2 allow list (rendering them unusable on subsequent open() calls)
This is because today, there is no way to append to the device allowlist https://github.com/opencontainers/runc/issues/3196

alaypatel07:
@akalenyu thanks for bring this point regarding cgroupv2, I will be sure to capture this in the doc.

alaypatel07:
closing in favor of https://github.com/kubevirt/enhancements/pull/11

PR #363, opened 2024-12-04
design-proposal: VM delete protection

jcanocan:

The proposal introduces a mechanism to protect VMs from being deleted. By adding the annotation `kubevirt.io/vm-delete-protection` and with a value of `true` or `True` to the VM object, the VM cannot be deleted by any means. This would prevent accidental deletions in VMs that users considers crucial for their applications/clusters.

Fixes # [CNV-48696](https://issues.redhat.com/browse/CNV-48696)

**Checklist**

**Release note**:

```release-note
None
```


design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,147 @@
...
+| Baseline   | 78.97                            | 6.03                            | 198.52                          |
+| Option \#1 | 78.96                            | 6.23                            | 199.61                          |
+| Option \#2 | 79.06                            | 6.16                            | 199.8                           |
+
+Please note that these tests are not aimed to compare exact deletion times between different experiment sets, but to
+compare them to a baseline and check if any option implies a degradation and if there is a faster option than the other.
+Differences between options are negligible in any scenario analyzed. Moreover, **no performance degradation has been
+observed.**
+
+Given the pros and cons, **OPTION \#1 is preferred in this case.**
```
0xFelix, line 130: Will we go with this option then?

jcanocan: IMHO, yes, we should go with this option.

design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+### Backend
+
+#### Option \#1: ValidatingAdmissionPolicy
+
+The VM protection is implemented by adding
+a  [ValidationAdmissionPolicy and ValidationAdmissionPolicyBinding](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/)
+objects. On every delete action against a VM object, the ValidatingAdmissionPolicy will check for presence of the
+annotation. If it is present and its value equals “true” or “True” the request will be rejected. The
```
geetikakay, line 63: Would it be useful to explain its behavior in scenarios where it is  misconfigured?

jcanocan: It might create a really verbose and noisy outputs IMO. Documentation should cover how to properly configure the feature, and it should make clear that other configurations/values, will allow the deletion.

jcanocan:
v2: 
- Included Section about Finalizers.

jean-edouard (commented):
While I'm not against this (as long as option 1 is used), I don't quite see the point.
Are we trying to protect PVCs that would have the VM as a parent? In that case, I think we should take steps to protect the PVCs directly instead.
The VM spec itself really isn't all that precious, and would cost just a few kilobytes of storage to backup somewhere if it really did contain special data.

This proposal sounds like there are interesting field stories associated with it, if so could you maybe share a concrete example of what happened?

codingben:
> While I'm not against this (as long as option 1 is used), I don't quite see the point. Are we trying to protect PVCs that would have the VM as a parent?

No. We're trying to protect VMs from deletion by accident, e.g. by sysadmins that may run business critical applications or services (e.g. SQL server).

design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,155 @@
...
+- It works with any Kubernetes version.
+
+Cons:
+
+- It will increase the maintenance effort.
+
+A proof of concept implementation can be found
+at:  [https://github.com/kubevirt/kubevirt/commit/2fafcf649019c26c8802838f2bbe673a9e93f04c](https://github.com/kubevirt/kubevirt/commit/2fafcf649019c26c8802838f2bbe673a9e93f04c)
+
+#### Option 3\#: Finalizers
```
codingben, line 107: After understanding the requested requirements, I don't think we should consider this as an option. Maybe just add a note about it that isn't releavnt to us.

design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,155 @@
...
+for backwards compatibility details.
+
+Pros:
+
+- Kubernetes built-in feature which is straightforward to maintain, implement and configure.
+- ValidationAdmisssionPolicy is [GA](https://kubernetes.io/blog/2024/03/12/kubernetes-1-30-upcoming-changes/) since Kuberentes version 1.30.
+
+Cons:
+
+- It will not work in deployments using Kubernetes versions \< 1.30
```
codingben, line 84: I'd not consider this as a con because this is not a requirement to work in older Kubernetes versions.

design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,155 @@
...
+metadata:
+  annotations:
+    kubevirt.io/vm-delete-protection: "true"
+spec:
+[...]
+```
+
+### Backend
+
+#### Option \#1: ValidatingAdmissionPolicy
```
codingben, line 58: I'd name it as a `#### Solution` instead of an option from the listed options. It should be clear to the reader that this is the chosen solution, and if you want to keep the second option, name it as `#### Alternative Solution`.

IMO, it would be also helpful for you as the owner to push towards implementing a single solution, and us as a reviewers to review it.

design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,155 @@
+# Overview
+
+VirtualMachine Delete Protection prevents the accidental deletion of VMs from a cluster. As long as deletion protection
+is activated for certain VM objects, they cannot be deleted unless the protection is intentionally removed.
+
+## Motivation
+
+KubeVirt clusters are often managed by automations, CLI commands, 3rd party tools or UI actions, for example to deploy,
+reconfigure or delete VMs. These automations may result in accidental deletion of VMs that should have not been deleted.
```
codingben, line 9: This is not accurate. There is no UI in KubeVirt upstream.

Since both CLI and UI will interact with the extended Kubernetes API to deploy, configure or delete VMs, I'd suggest just to mention that API call to delete the VM(s) will be rejected to avoid accidental deletions of the VMs by sysadmins (as an example of the user scenario).

design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,155 @@
...
+Affected VMs could be critical to application/s the cluster is running, or they may contain crucial data. Therefore, the
+VM deletion protection aims to avoid accidental degradation of applications or loss of data.
```
codingben, line 11: I'd remove `they may contain crucial data` and `or loss of data` from here. We don't want to have any reference to storage, since it may raise questions about the protection of PVC from deletion.

Our motivation is simple: Protect VMs from deletion that may run business critical applications or services, that deletion of them could cause out of service (e.g. SQL server is not available to handle requests).

design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,155 @@
+# Overview
+
+VirtualMachine Delete Protection prevents the accidental deletion of VMs from a cluster. As long as deletion protection
+is activated for certain VM objects, they cannot be deleted unless the protection is intentionally removed.
```
codingben, line 4: Can you add a section that explains how the user is expected to remove the protection (remove the annotation)? e.g. by patch operation?

design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,155 @@
...
+## Motivation
+
+KubeVirt clusters are often managed by automations, CLI commands, 3rd party tools or UI actions, for example to deploy,
+reconfigure or delete VMs. These automations may result in accidental deletion of VMs that should have not been deleted.
+Affected VMs could be critical to application/s the cluster is running, or they may contain crucial data. Therefore, the
+VM deletion protection aims to avoid accidental degradation of applications or loss of data.
+
+## Goals
+
+- Establish a delete protection mechanism for VirtualMachine objects.
```
codingben, line 15: I'd make it similar to non-goals:
```
Create a delete protection for VirtualMachine objects.
```
Because "Establish" could mean to start a new service to have a "mechanism" to prevent deletion (also IMO "mechanism" is an unclear complicated word for it).

design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Non Goals
+
+- Create a delete protection for VirtualMachineInstance objects.
+- Create a delete protection for virt-launcher Pod objects.
+
+## Definition of Users
+
+Everyone who has the permissions to delete VMs.
```
codingben, line 24: I'm not fan of "everyone", wouldn't you prefer to refer specifically to the user role, VM owner or cluster admins?

design-proposals/vm-delete-protection.md
0xFelix: Technically, you can install upstream KubeVirt through HCO on OCP/OKD. :)

design-proposals/vm-delete-protection.md
0xFelix: I think it's valid to mention this. If a PVC is deleted as a result of a cascaded delete that would result in loss of data.

design-proposals/vm-delete-protection.md
jcanocan: The discussion about finalizers has been triggered in multiple forums. Therefore, I've considered it relevant to add a specific Section explaining why we cannot implement this feature thought finalizers.

design-proposals/vm-delete-protection.md
jcanocan: I will keep it as it is for the reason provided.

design-proposals/vm-delete-protection.md
jcanocan: You are right, moreover, in the best case this feature will be onboarded in version 1.5, and according to https://github.com/kubevirt/sig-release/blob/main/releases/k8s-support-matrix.md, unless exception, we will support K8S ranging from 1.30 to 1.32.
Dropped.

design-proposals/vm-delete-protection.md
jcanocan: All right. Changed.

design-proposals/vm-delete-protection.md
jcanocan: Thanks for your inputs. Based on your comments, I've improved the text.
I find out Felix's argument valid, so I will keep the mention to storage.

design-proposals/vm-delete-protection.md
jcanocan: Done.

design-proposals/vm-delete-protection.md
jcanocan: Fair. Changed.

jean-edouard:
> > While I'm not against this (as long as option 1 is used), I don't quite see the point. Are we trying to protect PVCs that would have the VM as a parent?
> 
> No. We're trying to protect VMs from deletion by accident, e.g. by sysadmins that may run business critical applications or services (e.g. SQL server).

Sure but what do you mean by "VMs"? What, if not disks, are we afraid to lose?
From your answer, it sounds like we're more trying to protect against VM shutdown rather than VM deletion, maybe that's what the annotation should be about?

jcanocan:
> > > While I'm not against this (as long as option 1 is used), I don't quite see the point. Are we trying to protect PVCs that would have the VM as a parent?
> > 
> > 
> > No. We're trying to protect VMs from deletion by accident, e.g. by sysadmins that may run business critical applications or services (e.g. SQL server).
> 
> Sure but what do you mean by "VMs"? What, if not disks, are we afraid to lose? From your answer, it sounds like we're more trying to protect against VM shutdown rather than VM deletion, maybe that's what the annotation should be about?

As stated in the design document, what we are trying to protect is the VirtualMachine object from deletion.
Please note that this is a costumer request: https://issues.redhat.com/browse/CNV-46398?focusedId=25412651&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25412651
So, what we would like to achieve is a deletion protection for VM objects similar to RHV or other cloud providers have.
While I think your point about protecting VM from shutting down is valid, users look more concerned about accidental deletions.

jean-edouard:
> > > > While I'm not against this (as long as option 1 is used), I don't quite see the point. Are we trying to protect PVCs that would have the VM as a parent?
> > > 
> > > 
> > > No. We're trying to protect VMs from deletion by accident, e.g. by sysadmins that may run business critical applications or services (e.g. SQL server).
> > 
> > 
> > Sure but what do you mean by "VMs"? What, if not disks, are we afraid to lose? From your answer, it sounds like we're more trying to protect against VM shutdown rather than VM deletion, maybe that's what the annotation should be about?
> 
> As stated in the design document, what we are trying to protect is the VirtualMachine object from deletion. Please note that this is a costumer request: https://issues.redhat.com/browse/CNV-46398?focusedId=25412651&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25412651 So, what we would like to achieve is a deletion protection for VM objects similar to RHV or other cloud providers have. While I think your point about protecting VM from shutting down is valid, users look more concerned about accidental deletions.

Thank you for the pointer. Please note that access to the comment you linked to is restricted. The important part of it is that `Having some form of delete protection would provide peace of mind against sysadmin accidents.`
With great power comes great responsibility... No matter what we do, `kubectl delete -f` will bypass it. We're also talking about protecting something that can easily be re-created (a VM spec).
That being said, I don't feel strongly against this, as the maintenance burden seem minimal.
@vladikr WDYT?

jcanocan:
> > > > > While I'm not against this (as long as option 1 is used), I don't quite see the point. Are we trying to protect PVCs that would have the VM as a parent?
> > > > 
> > > > 
> > > > No. We're trying to protect VMs from deletion by accident, e.g. by sysadmins that may run business critical applications or services (e.g. SQL server).
> > > 
> > > 
> > > Sure but what do you mean by "VMs"? What, if not disks, are we afraid to lose? From your answer, it sounds like we're more trying to protect against VM shutdown rather than VM deletion, maybe that's what the annotation should be about?
> > 
> > 
> > As stated in the design document, what we are trying to protect is the VirtualMachine object from deletion. Please note that this is a costumer request: https://issues.redhat.com/browse/CNV-46398?focusedId=25412651&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25412651 So, what we would like to achieve is a deletion protection for VM objects similar to RHV or other cloud providers have. While I think your point about protecting VM from shutting down is valid, users look more concerned about accidental deletions.
> 
> Thank you for the pointer. Please note that access to the comment you linked to is restricted. The important part of it is that `Having some form of delete protection would provide peace of mind against sysadmin accidents.`

Yes, this is a good summary. Thanks!

>  With great power comes great responsibility... No matter what we do, `kubectl delete -f` will bypass it. 

Just one note, both approaches mentioned in this design document does not allow the force deletion either. You could check it in the PoCs linked.

We're also talking about protecting something that can easily be re-created (a VM spec). That being said, I don't feel strongly against this, as the maintenance burden seem minimal. @vladikr WDYT?

I agree, but from the user point of view, they will have the guarantee that the VM will not be deleted no matter what.

design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,171 @@
...
+kubevirt/kubevirt
+
+# Design
+
+The design is divided into two sections. The first section shows how the VM protection can be enabled by a user, the
+second section depicts possible implementations of the feature in the backend.
+
+### Enable the VM Object Protection
+
+To protect VMs against deletion, they need to be annotated with  `kubevirt.io/vm-delete-protection` set to `true` or
```
stu-gott, line 44: Personally, I'd much rather have a first-class setting in the VM. Documentation of annotations tends to be fragmented and doesn't lead to a very good user experience. Making a dedicated setting in the spec means it will be properly documented in swagger.

stu-gott (commented):
Overall I don't see why this is needed. I've seen reference to tekton pipelines as a justification, but I don't understand what the real problem is there.

jcanocan:
> Overall I don't see why this is needed. I've seen reference to tekton pipelines as a justification, but I don't understand what the real problem is there.

IMHO, Jed summed it up really well here: https://github.com/kubevirt/community/pull/363#issuecomment-2528259452
Looks like users accidentally delete VMs more that they would like to. Therefore, they are missing some delete protection in the VM objects. Most cloud providers have this option, e.g., https://cloud.google.com/compute/docs/instances/preventing-accidental-vm-deletion

design-proposals/vm-delete-protection.md
jcanocan: Many thanks for your feedback, much appreciated.

While I'm not against this idea, I can see potential challenges in the future. IMHO, a good documentation entry should be enough. The API field vs an annotation has the following advantages:
- Low maintenance, no need to maintain a new API field.
- Extensible, it would be just a matter of adding new `resourcesRules` to the existing VAP if we would like to protect other KubeVirt objects. Creating a dedicated API field will force us to complicate the VAP and even to move the API setting from different places (creating room for potential backwards compatibility challenges).

design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,171 @@
...
+- Create a delete protection for virt-launcher Pod objects.
+
+## Definition of Users
+
+VM owners with permissions to delete VMs.
+Cluster administrators with permissions to delete VMs.
+
+## User Stories
+
+As a user who has the permissions to delete VMs, I want to protect my VMs from accidental deletion.
```
lyarwood, line 31: I'd simplify this and just say `As a VM Owner,` the fact the owner has permissions to delete the VM isn't relevant to the fact that they want to protect it from accidental deletion by them or a third party.

design-proposals/vm-delete-protection.md
```diff
@@ -0,0 +1,171 @@
...
+spec:
+[...]
+```
+
+### Disable the VM Object Protection
+There are two possible ways to disable the VM protection:
+
+1. Removing the `kubevirt.io/vm-delete-protection` annotation with the following patch operation:
+```bash
+$ kubectl patch vm <vm-name> --type=json -p='[{"op": "remove", "path": "/metadata/annotations/kubevirt.io~1vm-delete-protection"}]'
```
lyarwood, line 64: This ultimately means that anyone with the ability to update/patch the VM can enable or disable this functionality. Have we considered placing it in status controlled by a subresource with finer grain RBAC contol?

lyarwood (commented):
Might be an invalid use case but what about using status and a subresource API to control this behaviour to allow finer grain RBAC control?

design-proposals/vm-delete-protection.md
jcanocan: Correct me if I'm wrong, but I assume this is the same concern raised in the [top comment of your review](https://github.com/kubevirt/community/pull/363#pullrequestreview-2491965394).

I didn't think about that TBH, but it feels a bit overkill IMHO.  However, what would be the ultimate goal of doing that? I'm not sure if this particular use case requires it. Please, elaborate a bit if you think otherwise.

In any case, if we find out this is required, we could implement it in a follow-up.

design-proposals/vm-delete-protection.md
jcanocan: Done.

design-proposals/vm-delete-protection.md
lyarwood: The ultimate goal would be to allow a smaller subset of users (super admins?) to have access to the delete protection toggle than have the ability to update the VM object itself. It's extra work but given the use case here I'm not sure we want to give everyone access to this toggle right?

vladikr:
Thank you for taking the time to write this proposal @jcanocan.

At first, I thought that this might be a good idea. Then I became conflicted as it seems that ultimately we are protecting these VMs from the cluster admins. 
Those who have the ultimate responsibility in the cluster...
Aren't we going to ruin the hierarchy in the cluster if we allow users to prevent an admin operation?

If we really need such a layer of protection, I would imagine this feature working from another direction.
Perhaps a user could provide a known "hint" (in the form of a label) and the admin could create a list of protected VMs that have label X.

jcanocan:
> Thank you for taking the time to write this proposal @jcanocan.
> 
> At first, I thought that this might be a good idea. Then I became conflicted as it seems that ultimately we are protecting these VMs from the cluster admins. Those who have the ultimate responsibility in the cluster... Aren't we going to ruin the hierarchy in the cluster if we allow users to prevent an admin operation?
> 
> If we really need such a layer of protection, I would imagine this feature working from another direction. Perhaps a user could provide a known "hint" (in the form of a label) and the admin could create a list of protected VMs that have label X.

We could enable the protection by setting a label like:

```yaml
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  labels:
    kubevirt.io/vm-delete-protection: "true"
```

In this way, as you mentioned, the cluster-admin could get a list of VMs with the protection enabled with something like:

```bash
kubectl get vm --selector=kubevirt.io/vm-delete-protection=true
```

WDYT? Did I get your idea right?

Thanks for your input.

design-proposals/vm-delete-protection.md
lyarwood: @dominikholler well it's any user with DELETE privileges against the VM surely so not just an owner but anyone in the same namespace or admin. It also doesn't protect against that same user accidentently removing the label and thus the protection before requesting the deletion.

Moving the toggle into the status would avoid this as users wouldn't be able to touch it without going through the sub resource API that itself could be limited to a specific owner and/or super set of admins.

vladikr:
@jcanocan What I had in mind is something closer to `MigrationPolicy` or a CR configuration similar to `PermittedHostDevices`

The idea is to separate the concerns of each persona.
The VM owner would "hint" by setting a label:

```yaml
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  labels:
    kubevirt.io/please-avoid-deleting: "true"
```

Then the admin would list the relevant labels to reject the deletion of the associated VMs:

```yaml
preventVMsDeletion:
  selector: kubevirt.io/please-avoid-deleting
```

dominikholler:
> @jcanocan What I had in mind is something closer to `MigrationPolicy` or a CR configuration similar to `PermittedHostDevices`
> 
> The idea is to separate the concerns of each persona. The VM owner would "hint" by setting a label:
> 
> ```yaml
> apiVersion: kubevirt.io/v1
> kind: VirtualMachine
> metadata:
>   labels:
>     kubevirt.io/please-avoid-deleting: "true"
> ```
> 
> Then the admin would set the rule that would prevent the deletion of the associated VMs with the listed label :
> 
> ```yaml
> preventVMsDeletion:
>   selector: kubevirt.io/please-avoid-deleting
> ```

@vladikr Isn't this already working [out-of-the-box](https://issues.redhat.com/browse/CNV-46398?focusedId=26067149&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26067149) using ValidatingAdmissionPolicy ?

This feature is about having a fixed marking (e.g. label or annotation) so that the UI can configure a marking for a pre-defined rule.

dominikholler:
> > Might be an invalid use case but what about using status and a subresource API to control this behaviour to allow finer grain RBAC control?
> 
> It's an interesting idea. How would this work with newly created VMs? Doesn't it have to be only a secondary action?

In which way would help this idea to help the admin with "fat-fingers" ?

vladikr:
> > @jcanocan What I had in mind is something closer to `MigrationPolicy` or a CR configuration similar to `PermittedHostDevices`
> > The idea is to separate the concerns of each persona. The VM owner would "hint" by setting a label:
> > ```yaml
> > apiVersion: kubevirt.io/v1
> > kind: VirtualMachine
> > metadata:
> >   labels:
> >     kubevirt.io/please-avoid-deleting: "true"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then the admin would set the rule that would prevent the deletion of the associated VMs with the listed label :
> > ```yaml
> > preventVMsDeletion:
> >   selector: kubevirt.io/please-avoid-deleting
> > ```
> 
> @vladikr Isn't this already working [out-of-the-box](https://issues.redhat.com/browse/CNV-46398?focusedId=26067149&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26067149) using ValidatingAdmissionPolicy ?
> 

Please elaborate on what is working out of the box. :)
I mentioned CEL (and I meant `ValidatingAdmissionPolicy` ) in my previous [comment](https://github.com/kubevirt/community/pull/363#issuecomment-2539065085)

I'm only trying to emphasize that we shouldn't now allow users to block admin operations without the admins' consent or/and a way to override this.

We achieve this separation either with `ValidatingAdmissionPolicy` or with a policy CRD that the admin would install or a CR config. 



> This feature is about having a fixed marking (e.g. label or annotation) so that the UI can configure a marking for a pre-defined rule.

This proposal is for KubeVirt which doesn't have a UI... but in all cases, I see a label or an annotation being used to request this feature.

vladikr:
> > > Might be an invalid use case but what about using status and a subresource API to control this behaviour to allow finer grain RBAC control?
> > 
> > 
> > It's an interesting idea. How would this work with newly created VMs? Doesn't it have to be only a secondary action?
> 
> In which way would help this idea to help the admin with "fat-fingers" ?

Not by itself, but this is just another way to limit the configuration of this feature to cluster admins and vm owners.

design-proposals/vm-delete-protection.md
codingben: > Technically, you can install upstream KubeVirt through HCO on OCP/OKD. :)

Is it officially supported?

Let's keep this discussion open because there is no UI in KubeVirt, I just want to get attention of others reading it, altough I understand the importance of mentioning it.

codingben:
> This proposal is for KubeVirt which doesn't have a UI...

I think we should get rid of mentioning "UI" in the design proposal. I've already opened discussion about it earlier this week: https://github.com/kubevirt/community/pull/363#discussion_r1874761854

lyarwood:
> > > > Might be an invalid use case but what about using status and a subresource API to control this behaviour to allow finer grain RBAC control?
> > > 
> > > It's an interesting idea. How would this work with newly created VMs? Doesn't it have to be only a secondary action?

Yeah I can't think of a clean way of providing it with the initial definition as you'd always have to remove it once populated in status to ensure a single source of truth.

> > In which way would help this idea to help the admin with "fat-fingers" ?
> 
> Not by itself, but this is just another way to limit the configuration of this feature to cluster admins and vm owners.

Yup it makes the act more deliberate by forcing an admin or whomever it is with access to the subresource to first toggle delete protection there and then fire off the delete.

Without this any user with the ability to update the VM object can toggle delete protection and then nuke the VM.

jcanocan:
So far, it looks like @vladikr's idea about activating the protection first in the KubeVirt CR is well received, and it is still technically possible using VAP. Therefore, let's take that direction.

Moreover, @lyarwood's subresource concern is valid, @vladikr @dominikholler do you agree that the "vm delete protection" subresource idea should be included in this design?

Also, all mentions to UI in the design document will drop.

Many thanks for all your ideas, feedback and reviews, much appreciated.

design-proposals/vm-delete-protection.md
dominikholler: Since there seems to be a demand here to delve into the details of the wording, let's do it.
In Wikipedia, the definition of the term [user interface](https://en.wikipedia.org/wiki/User_interface) begins as follows
> In the industrial design field of human–computer interaction, a user interface (UI) is the space where interactions between humans and machines occur. The goal of this interaction is to allow effective operation and control of the machine from the human end, while the machine simultaneously feeds back information that aids the operators' decision-making process.


Contrary to the beginning of the definition of the term [API](https://en.wikipedia.org/wiki/API):
> An application programming interface (API) is a connection between computers or between computer programs. It is a type of software interface, offering a service to other pieces of software.

> There is no UI in KubeVirt upstream.

Yes, the kubevirt/kubevirt repo only provides an API. Never the less there might be humans using kubevirt via the ways which are mentioned in the design proposal, even if the code is in repos other than kubevirt/kubevirt.

From my personal point of view, it is essential that this feature, which is intended to help humans, is seen from the point of view of human users.

dominikholler:
> 
> Please elaborate on what is working out of the box. :) I mentioned CEL (and I meant `ValidatingAdmissionPolicy` ) in my previous [comment](https://github.com/kubevirt/community/pull/363#issuecomment-2539065085)
> 
> I'm only trying to emphasize that we shouldn't now allow users to block admin operations without the admins' consent or/and a way to override this.
> 
> We achieve this separation either with `ValidatingAdmissionPolicy` or with a policy CRD that the admin would install or a CR config.
> 

According to my limited technical understanding, the cluster admin can already right now create a ValidatingAdmissionPolicy like this:

```
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicy
metadata:   
  name: "vm-delete-protection"
spec:   
  failurePolicy: Fail
  matchConstraints:     
    resourceRules:     
    - apiGroups:   ["kubevirt.io"]
      apiVersions: ["*"]
      operations:  ["DELETE"]
      resources:   ["virtualmachines"]
  variables: 
    - expression: string('kubevirt.io/vm-delete-protection')
      name: annotation
  validations: 
    - expression: (!(variables.annotation in oldObject.metadata.annotations) || !oldObject.metadata.annotations[variables.annotation].matches('^(true|True)$'))
      messageExpression: "'Object ' + string(oldObject.metadata.name) + ' cannot be deleted, remove delete protection'"
---
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicyBinding
metadata:   
  name: "vm-delete-protection-binding"
spec:   
  policyName: "vm-delete-protection"
  validationActions: [Deny]
  matchResources: 
```
Now, if the VM owner adds the annotation `kubevirt.io/vm-delete-protection: "true"` to any VM object, it will be protected against deletion. You can do it by applying the following path to your VM object:

```
kubectl patch <vm-name> -n <namespace> --type=json -p '[{"op": "add", "path": "/metadata/annotations/kubevirt.io~1vm-delete-protection", "value": "true"}
```

To reverse the operation and allow deletion, the following command can be used:

```
kubectl patch vm <vm-name> -n <namespace> --type=json -p '[{"op": "remove", "path": "/metadata/annotations/kubevirt.io~1vm-delete-protection"}]'
```

the same should work if a label is used instead of the annotation.

This feature is about providing a easy to use, simple and testable interface to the already existing solution. 
For this reason the feature would loose its value, if it is introducing more complexity than what already exists.

But I see the point, that there should be a way to introduce a way for the admin to control the deletion protection. 

> >  In which way would help this idea to help the admin with "fat-fingers" ?

> Not by itself, but this is just another way to limit the configuration of this feature to cluster admins and vm owners.

I see, thanks for the hint!

design-proposals/vm-delete-protection.md
dominikholler: >  It also doesn't protect against that same user accidentently removing the label and thus the protection before requesting the deletion.

This would require two steps, this feature is just protecting from a single unintended step, e.g. a `kubectl delete vm/xxx" .

dominikholler:
> Without this any user with the ability to update the VM object can toggle delete protection and then nuke the VM.

Where do you see the issue with this? Isn't it expected that if a user has the vm owners permissions is able to delete the VM?

xpivarc:
@dominikholler The policy you wrote is exactly what you want. Now why should this opinionated (subjective at least to me) policy be part of core Kubevirt? Especially when it is so easy to implement as you demonstrated? Maybe more opinionated component could introduce this?

jcanocan:
Update:
- Added new chosen solution, deploy VAP in SSP.
- Changed the mechanism to enable the protection from annotation to label
- Changed affected repos from kubevirt to SSP.

iholder101:
I agree with others' comments from above.

I'm not comfortable with this proposal. Kubernetes provides mechanisms to manage permissions regarding who's able to delete the different objects, like RBAC and VAP. By definition, admins are the trusted persona to perform dangerous actions, and the admin can easily add an opinionated VAP as demonstrated above if he wants to be extra paranoid.

Thank you @jcanocan for your efforts on this one, but unless there's a compelling argument for it, I tend to advocate against it.

iholder101:
> Please note that this is a costumer request: [issues.redhat.com/browse/CNV-46398?focusedId=25412651&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25412651](https://issues.redhat.com/browse/CNV-46398?focusedId=25412651&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25412651) 

Kubevirt has users, not customers :)

jcanocan:
Hi all! This feature has been merged in https://github.com/kubevirt/ssp-operator/pull/1199, and it is already available though SSP operator. In order to keep the room clean, could we get this merged if we find it appropriate? If not, I can just close it. 
Thanks in advance!

PR #361, opened 2024-12-02
cli-tool: generate github contribution reports using the Github GraphQL api

dhiller:
**What this PR does / why we need it**:

We have had several cases leading to extensive discussions over removing users from either the org or from an OWNERS file. We are constantly struggling with providing evidence about whether a user is really inactive whenever we have the gut feeling that they are.

We want to avoid having inactive reviewers and inactive org members, since:

1. inactive reviewers are holding up contributors from getting their PRs reviewed and approved, forcing them to ping people. In the worst case those aren't active any more and might even ignore the ping.
2. org membership comes with elevated privileges (i.e. automatic runs of test lanes and being able to lgtm PRs), which we want to reserve for people with active contribution status.

This tool tries to help with the above problems, offering to tackle three use cases:
- fetching contribution data for a specific user
- checking activity from contribution data for the members of an org (given by orgs.yaml)
- checking activity from contribution data for reviewers and approvers (given an OWNERS file)

Detailed information in the README: https://github.com/dhiller/community/tree/create-user-activity-tool/generators/cmd/contributions/README.md

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #359

**Special notes for your reviewer**:

PR #365, opened 2024-12-10
Updating membership_policy to inc removal of reviewer/approvers/chairs

aburdenthehand:

Adds guidelines for removing/demoting reviewers/approvers/chairs/leads. Primarily this makes the recommendation that they should step down of their own accord and remove themselves; when that does not happen, inactivity of 6 months or more can trigger their removal.

The need for this became apparent during #329

/hold because it is dependent on #361 as the tool to confirm inactivity. 
@dhiller @EdDev @vladikr 

Fixes #

```release-note
Updating our membership policy to include how reviewers/approvers/chairs/leads should step down, or otherwise be removed/demoted.
```

### DIFF
--- a/membership_policy.md
+++ b/membership_policy.md
@@ -117,6 +117,8 @@ The following apply to the part of project for which one would be an approver in
   * Nominated by an approver
     * With no objections from other approvers
     * Done through PR to update the top-level OWNERS file
+    
+In exceptional circumstances, such as when a repo is left without sufficient approvers, the project maintainers can pass a majority vote to add approvers that do not fulfill the requirements.
 
 ### Responsibilities and privileges
 
@@ -162,7 +164,7 @@ The KubeVirt project is organized primarily into SIGs, each with a common purpos
 
 ## SIG Subproject Lead
 
-Specific work efforts within SIGs can be divided into subprojects. A SIG Subproject Lead is required to help guide and coordinate the subproject.  
+Specific work efforts within SIGs can be divided into subprojects. A SIG Subproject Lead is required to help guide and coordinate the subproject.
 
 ### Requirements
 
@@ -208,7 +210,11 @@ Working groups are primarily used to facilitate topics of discussion that cross 
   * Gives updates to respective sponsoring SIG Chairs
   * The broader community
 
-## Inactive members
+## Inactivity
+
+Impermanence is one of the true constants in life: no one is expected to be with the project forever. 
+
+### Inactive members
 
 [_Members are continuously active contributors in the community._](#member)
 
@@ -225,7 +231,7 @@ will be removed from the KubeVirt GitHub Organization and will be required to
 go through the org membership process again after re-familiarizing themselves
 with the current state.
 
-### How inactivity is measured
+#### How member inactivity is measured
 
 Inactive members are defined as members of the KubeVirt Organization
 with **no** contributions within the KubeVirt organization within 12 months. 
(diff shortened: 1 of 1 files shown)


membership_policy.md
```diff
@@ -239,6 +243,24 @@ After an extended period away from the project with no activity
...
+Considering that they have elevated permissions and responsibilities in the project, their active engagement with the project is greatly appreciated.
+
+If a reviewer and/or approver steps down from their position in a repository or the project entirely, we trust that they will notify the community and create PRs removing them as reviewers/approvers in the applicable repositories.
+This will help others identify and grow into the role so as to not leave a gap for any length of time. 
+It will also ensure they are not erroneously tagged to review/approve PRs, which is both an inconvenience to them but also an obstacle for the project.
+
+When a SIG or WG Chair or Subproject Lead needs to step down from their position, it is expected that they work with the SIG/Subproject/WG to identify and mentor a willing candidate to grow into the role.
+They should then open a PR to replace them in [sigs.yaml], and move themselves to emeritus status (if applicable).
+
+A reviewer/approver/chair/lead may also be demoted or removed from their role if they are inactive in their area for 6 months or more.
```
dhiller, line 258: I think we also should think about the case of a "vanishing" person. In that case there's no candidate, so the impacted group or the community should select one.

This could happen maybe by approaching the candidate - either by maintainers of the affected group or the project maintainers in the worst case - and then suggesting them to the group or community in a PR against the `sigs.yaml`. A quorum of people with authority should then be enough to accept the suggestion.

aburdenthehand:
Removing hold as #361 has merged

membership_policy.md
iholder101: +1.
In general I like the idea of having an "escape hatch" to such defined policies. After all, it makes sense that due to circumstances sometimes someone will vanish with no notice and won't have the time to fulfill this process or even communicate the situation. In such a scenario, I would say that if the majority of SIG chairs vote to remove them because they know they won't be active for the near future, there is no need to wait for 6 months just to get the "proof" of inactiveness.

No policy can replace common sense 100%, but IMO is mostly there to serve as a general guideline.

iholder101:
Added a small comment above, but in general looks great to me.
Thank you @aburdenthehand!

membership_policy.md
dhiller: @aburdenthehand my suggestion could also be handled in a follow up, I will not block on this.

dhiller (approved):
Thank you @aburdenthehand !

membership_policy.md
aburdenthehand: Thanks for the feedback folks, and sorry about the delay. 
I've added #L121 to cover a vote for approvers who don't meet the requirements, and added to this line. I think this covers the suggested cases but let me know if they do not. 
+1 that common sense should prevail, but always good to cover these cases.

dhiller (approved):

membership_policy.md
```diff
@@ -117,6 +117,8 @@ The following apply to the part of project for which one would be an approver in
   * Nominated by an approver
     * With no objections from other approvers
     * Done through PR to update the top-level OWNERS file
+    
+In exceptional circumstances, such as when a repo is left without suffcient approvers, the project maintainers can pass a majority vote to add approvers that do not fulfil the requirements.
```
EdDev, line 121: IMO the bar is low enough as it is. Why should we accept exceptions?
Perhaps we can focus on providing SIG approvers the ability to nominate repo subproject approvers, that sounds a bit more loose.

EdDev (commented):
Thank you!
This looks great.

I only request to remove the exception addition which feels a bit problematic to me.
Is that possible?

membership_policy.md
jean-edouard, line 121:
```suggestion
In exceptional circumstances, such as when a repo is left without sufficient approvers, the project maintainers can pass a majority vote to add approvers that do not fulfill the requirements.
```

membership_policy.md
```diff
@@ -225,7 +231,7 @@ will be removed from the KubeVirt GitHub Organization and will be required to
 go through the org membership process again after re-familiarizing themselves
 with the current state.
 
-### How inactivity is measured
+#### How inactivity for members is measured
```
jean-edouard, line 234:
```suggestion
#### How members inactivity is measured
```

membership_policy.md
```diff
@@ -239,6 +245,24 @@ After an extended period away from the project with no activity
 those members would need to re-familiarize themselves with the current state
 before being able to contribute effectively.
 
+### Inactive Reviewers/Approvers/Chairs/Leads
+
+Reviewers, approvers, chairs and leads are pillars of the community and are fundamental to maintaining the momentum of the project.
+Considering that they have elevated permissions and responsibilities in the project, their active engagement with the project is greatly appreciated.
+
+If a reviewer and/or approver steps down from their position in a repository or the project entirely, we trust that they will notify the community and create PRs removing them as reviewers/approvers in the applicable repositories.
```
jean-edouard, line 253:
```suggestion
If a reviewer and/or approver steps down from their position in a repository or the project entirely, we trust that they will notify the community and create PRs to remove themselves as reviewers/approvers in the applicable repositories.
```

jean-edouard (commented):
@aburdenthehand I found a couple typos and suggested some rewordings.
This PR otherwise looks good to me, please just address remaining comments. Thanks!

aburdenthehand:
@jean-edouard Thanks for the catches :)

membership_policy.md
EdDev: Well, if a project has no active maintainers, the "council" should either find one or archive the project.
I am more concerned that this 1% of the cases which you aim here will become in some projects a common norm.
But fine, I understand better from where you come with this.

EdDev (approved):

membership_policy.md
```diff
@@ -239,6 +245,24 @@ After an extended period away from the project with no activity
...
+It will also ensure they are not erroneously tagged to review/approve PRs, which is both an inconvenience to them but also an obstacle for the project.
+
+When a SIG or WG Chair or Subproject Lead needs to step down from their position, it is expected that they work with the SIG/Subproject/WG to identify and mentor a willing candidate to grow into the role.
+They should then open a PR to replace them in [sigs.yaml], and move themselves to emeritus status (if applicable).
+
+A reviewer/approver/chair/lead may also be demoted or removed from their role if they are inactive in their area for 6 months or more, or if there is a majority vote by the remaining approvers/chairs/leads or maintainers depending on the situation (for instance, if a SIG chair abruptly leaves one remaining SIG chair).
+
+#### How inactivity for Reviewers/Approvers/Chairs/Leads is measured
+
+The KubeVirt community repo has a [project activity tool](https://github.com/kubevirt/community/blob/main/generators/cmd/contributions/README.md) that can be used to demonstrate that a reviewer/approver has been inactive for 6 months.
```
jean-edouard, line 264: @dhiller don't we have a periodic execution of this, with public-facing output? I think a link to such output would be more valuable than a link to the readme of the tool.

jean-edouard (commented):
/approve
/hold
Just one question in-line, feel free to unhold.

membership_policy.md
dhiller: > @dhiller don't we have a periodic execution of this, with public-facing output? I think a link to such output would be more valuable than a link to the readme of the tool.

No, we don't. My first question would be how we would want to present this though.

I was trying to tackle this in a different way: by automating the removal of accounts who don't show activity - this way the process is democratized. However the PR didn't get much feedback, not sure why that is.

https://github.com/kubevirt/project-infra/pull/3832

dhiller: ^^ @jean-edouard

PR #369, opened 2024-12-18
Proposal for inject/eject CD-ROM

mhenriks:

Currently, a VM has to be rebooted to change the CD-ROM media. Let's leverage the hotplug volume API to fix that gaping hole in functionality

Fixes #

**Checklist**

**Release note**:

```release-note
Proposal for inject/eject CD-ROM
```

### DIFF
--- a/design-proposals/cdrom-inject-eject.md
+++ b/design-proposals/cdrom-inject-eject.md
@@ -1,0 +1,198 @@
+# Overview
+Proposal for for adding inject/eject CD-ROM support to KubeVirt
+
+## Motivation
+In a world where all of our knowledge has been collected, indexed, and can be served up via a network connection, the humble CD-ROM is still a "thing." And to the dedicated hangers-on of this once ubiquitous technology, KubeVirt's CD-ROM support is lacking in a very fundamental way. Today, a user must reboot a VM in order to change the media in a CD-ROM. That is just. not. right. Fortunately, the virtualization technology that underpins KubeVirt has had this functionality for a long time and we can easily plug it into the KubeVirt API to unleash the full power of the CD-ROM
+
+## Goals
+- Extend existing APIs to support inject/eject CD-ROM operations
+- Extend virtctl to have cdrom specific commands
+- Support for PersistentVolumeClaim/DataVolume based disk images
+
+## Non Goals
+- Support for any [volume type](https://kubevirt.io/user-guide/storage/disks_and_volumes/#volumes) other than DavaVolume/PersistentVolumeClaim
+
+## Definition of Users
+KubeVirt end user, may not be a namespace/cluster admin
+
+## User Stories
+- As a VM owner, I want to be able to attach a CD-ROM for the purposes of installing an operating system, then be able eject the CD and perform a guest-side reboot to use the operating system
+- As a VM Owner, I want to attach an ISO/CD-ROM from my desktop to the VM without the need to reboot the VM
+- As a VM Owner, I want to attach a PVC which contains an ISO to my VM without the need to reboot the VM
+- As a VM Owner, I want to be able to "eject" (remove) the ISO/CD from my VM without the need to reboot the VM
+
+## Repos
+kubevirt/kubevirt
+
+# Design
+Like most new features in KubeVirt, the building blocks already exist, just have to tweak a couple things and wire it all together
+
+## VirtualMachineInstance API
+Currently, every disk in a KubeVirt VM must have a corresponding volume. That restriction will be lifted for CD-ROM disks, i.e., it will be possible to define a VMI that has no corresponding volume for a CD-ROM. For Example:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+...
+spec:
+  domain:
+    devices:
+      disks:
+      - cdrom:
+        bus: sata
+      name: cdrom
+      - disk:
+        bus: virtio
+      name: root
+...
+  volumes:
+  - dataVolume:
+      name: fedora-root
+    name: root
+  networks:
+...
+```
+
+## Volume Hotplug Subresource API
+This feature leverages the Volume Hotplug API in order to add/remove disk images in running Pods. Hence, the restriction on PersistentVolumeClaim and DataVolume volume types.
+
+Hotplug operations are wrapped in the `VirtualMachineVolumeRequest` struct which looks like this:
+
+```golang
+type VirtualMachineVolumeRequest struct {
+  // AddVolumeOptions when set indicates a volume should be added. The details
+  // within this field specify how to add the volume
+  AddVolumeOptions *AddVolumeOptions `json:"addVolumeOptions,omitempty" optional:"true"`
+  // RemoveVolumeOptions when set indicates a volume should be removed. The details
+  // within this field specify how to add the volume
+  RemoveVolumeOptions *RemoveVolumeOptions `json:"removeVolumeOptions,omitempty" optional:"true"`
+}
+```
+
+One of `AddVolumeOptions` or `RemoveVolumeOptions` is expected to be set
+
+### AddVolumeOptions
+`AddVolumeOptions` as currently defined:
+
+```golang
+type AddVolumeOptions struct {
+  // Name represents the name that will be used to map the
+  // disk to the corresponding volume. This overrides any name
+  // set inside the Disk struct itself.
+  Name string `json:"name"`
+  // Disk represents the hotplug disk that will be plugged into the running VMI
+  Disk *Disk `json:"disk"`
+  // VolumeSource represents the source of the volume to map to the disk.
+  VolumeSource *HotplugVolumeSource `json:"volumeSource"`
+  DryRun []string `json:"dryRun,omitempty"`
+}
+```
+
+At present, the `disk` field is required. That restriction will be removed. However, when the `addVolume` request is received, we will validate that the corresponding disk is in fact a CD-ROM. The following request will be valid for the VirtualMachine defined above:
+
+```yaml
+addVolumeOptions:
+  name: cdrom
+  volumeSource:
+    persistentVolumeClaim:
+       claimName: windows11-iso
+       hotpluggable: true
+```
+
+### RemoveVolumeOptions
+`RemoveVolumeOptions` as currently defined:
+
+```golang
+type RemoveVolumeOptions struct {
+  // Name represents the name that maps to both the disk and volume that
+  // should be removed
+  Name string `json:"name"`
+  DryRun []string `json:"dryRun,omitempty"`
+}
+```
+
+Since we don't want to remove the CD-ROM when ejecting a disk, a new field will be added to `RemoveVolumeOptions` to specify whether to keep/remove a disk.
+
+```golang
+// DiskRetentionPolicy defines how removevolume subresource API
+// requests should handle the corresponding disk
+type DiskRetentionPolicy string
+
+// DiskRetentionDelete specifies that removevolume should
+// delete the matching disk in the VirtualMachineInstance spec
+const DiskRetentionDelete = "delete"
+
+// DiskRetentionKeep specifies that removevolume should
+// keep the matching disk in the VirtualMachineInstance spec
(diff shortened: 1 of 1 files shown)


design-proposals/cdrom-inject-eject.md
```diff
@@ -0,0 +1,173 @@
...
+- As a VM owner, I want to be able to attach a CD-ROM for the purposes of installing an operating system, then be able eject the CD and perform a guest-side reboot to use the operating system.
+- As a VM Owner I want to attach an ISO/CD-ROM from my desktop to the VM without the need to reboot the VM
+- As a VM Owner I want to attach a PVC which contains an ISO to my VM without the need to reboot the VM
+- As a VM Owner I want to be able to "eject" (remove) the ISO/CD from my VM without the need to reboot the VM
+
+## Repos
+kubevirt/kubevirt
+
+# Design
+Like most new features in KubeVirt, the building blocks already exist, just have to wire everything together
```
dankenigsberg, line 28: Please consider alternative designs and their pros/cons. In particular, can you consider following https://github.com/kubevirt/community/blob/e4645eb/design-proposals/vm-rollout-strategy/vm-rollout-strategy.md to enable a simpler, declarative, API. According to it, the user expresses a VM spec that includes a CDROM with a specific PVC and the VM controller makes that happen.

dankenigsberg (changes requested):
Thanks for caring for this much-requested functionality.

mw-0:
this would be great. this is a real user missing feature moving from alternatives.

design-proposals/cdrom-inject-eject.md
mhenriks: I can address that in the design, but just to answer you inline, I think the functionality you describe can be generalized and included in an all encompassing "Declarative Volume Hotplug" design

jean-edouard:
I would add an important non-goal about "config volumes", typically for service accounts.
Config volumes are ISOs *but* they're exposed as disks, not cdroms for some reason, so they can't be ejected/inserted.

design-proposals/cdrom-inject-eject.md
dankenigsberg: > I can address that in the design,

Thanks.

> I think the functionality you describe can be generalized and included in an all encompassing "Declarative Volume Hotplug" design

Right, I'd love to see that. But I'd be satisfied in seeing CDROM-only as a first step.

mhenriks: I think just CD-ROM being declarative would be confusing to users. 

To me, what is proposed here is a good first step in making this long awaited functionality available to users soon. Going full declarative is a lot more complicated. But is something I do agree that we should tackle eventually

mhenriks:
> I would add an important non-goal about "config volumes", typically for service accounts. Config volumes are ISOs _but_ they're exposed as disks, not cdroms for some reason, so they can't be ejected/inserted.

The fact that they they are isos is more of an implementation detail to me. Do end users really know/care when it is exposed as a disk?

I state as a non-goal `Support for any volume type other than DataVolume/PersistentVolumeClaim`. To me that is pretty clear, but maybe I can link to all the volume types we support

jean-edouard:
> > I would add an important non-goal about "config volumes", typically for service accounts. Config volumes are ISOs _but_ they're exposed as disks, not cdroms for some reason, so they can't be ejected/inserted.
> 
> The fact that they they are isos is more of an implementation detail to me. Do end users really know/care when it is exposed as a disk?

They do know, since they are directly consuming them. And they probably do care, since that limitation prevents service account data from getting updated.

> 
> I state as a non-goal `Support for any volume type other than DataVolume/PersistentVolumeClaim`. To me that is pretty clear, but maybe I can link to all the volume types we support

You're right that's probably enough, thanks!

design-proposals/cdrom-inject-eject.md
dankenigsberg: Please consider pros and cons each of these options in the text of the design.

stu-gott:
/hold

A request was made to compare pros/cons of doing just CD-ROM vs a more generalized API. This hold can be removed once that's addressed.

design-proposals/cdrom-inject-eject.md
alicefr: Adding my 2 cents here, declarative is a nice to have, however, this type of installation operation is imperative and kind of "manual". Do we really expect this to be used with git ops?

alicefr:
@mhenriks aren't we missing the setting the boot order for the other disks? I mean, if we eject the cdrom, how does the VM know from which disk it has to boot the next time it is restarted?

mhenriks:
> /hold
> 
> A request was made to compare pros/cons of doing just CD-ROM vs a more generalized API. This hold can be removed once that's addressed.

@stu-gott @dankenigsberg I don't think that it makes sense to debate the pros and cons of declarative vs subresource APIs in this design. Declarative volume hotplug is bigger topic, not just for CD-ROMs, and will be a separate design proposal. 

This proposal only describes two changes to existing APIs which are specific to inject/eject CD-ROM support:

1. Update the `VirtualMachineInstance` API to support a CD-ROM disk without a matching volume (empty cd-rom tray)
2. Allow `VirtualMachineVolumeRequest` to support adding/removing a volume without adding/removing the corresponding disk

That's it. Those two changes allow us to support inject/eject CD-ROM in short order by leveraging the existing hotplug machinery. How about we judge this proposal based on its merits and with the understanding that there will be a declarative implementation eventually? I kindly ask that the hold on this PR be removed.

mhenriks:
> @mhenriks aren't we missing the setting of the boot order for the other disks? I mean, if we eject the cdrom, how does the VM know from which disk it has to boot the next time it is restarted?

Hey @alicefr qemu seems to handle an empty cd-rom with `bootOrder: 1` just fine in my testing. It will go to disk 2

dankenigsberg:
> > /hold
> > A request was made to compare pros/cons of doing just CD-ROM vs a more generalized API. This hold can be removed once that's addressed.
> 
> @stu-gott @dankenigsberg I don't think that it makes sense to debate the pros and cons of declarative vs subresource APIs in this design. Declarative volume hotplug is bigger topic, not just for CD-ROMs, and will be a separate design proposal.
> 
> This proposal only describes two changes to existing APIs which are specific to inject/eject CD-ROM support:
> 
> 1. Update the `VirtualMachineInstance` API to support a CD-ROM disk without a matching volume (empty cd-rom tray)
> 2. Allow `VirtualMachineVolumeRequest` to support adding/removing a volume without adding/removing the corresponding disk
> 
> That's it. Those two changes allow us to support inject/eject CD-ROM in short order by leveraging the existing hotplug machinery. How about we judge this proposal based on its merits and with the understanding that there will be a declarative implementation eventually? I kindly ask that the hold on this PR be removed.

@mhenriks I agree with you that the cost of implementation is an important consideration when choosing a course of action, but it cannot be the only one. I believe that there's a lot of sense, of having an API proposal consider alternatives, describing the benefits and costs of each, and finally choosing one.

mhenriks:
> @mhenriks I agree with you that the cost of implementation is an important consideration when choosing a course of action, but it cannot be the only one. I believe that there's a lot of sense, of having an API proposal consider alternatives, describing the benefits and costs of each, and finally choosing one.

@dankenigsberg with regards to declarative vs subresource I don't see why we have to choose one. The subresource case is described here. The declarative case will be handled implicitly by declarative volume hotplug

mhenriks:
Inspired by feedback from the passionate CD-ROMers in the community and the [k8s kep template](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/README.md#drawbacks) to add `Drawbacks` and `Alternatives`  section

design-proposals/cdrom-inject-eject.md
```diff
@@ -0,0 +1,198 @@
...
+### Not GitOps compatible
+Subresource APIs are not helpful to users that want to communicate all VM state changes in a declarative/GitOps compatible way. In fact, GitOps users have to be extra careful when using this feature to avoid conflicts with volume names/definitions. Unfortunately, KubeVirt does not support declarative volume updates for any disk type at this time.
+
+### Complicated for users to invoke directly
+kubectl does not give users an easy way to invoke a subresource API. Which means, as a practical matter, virtctl support is necessary and not just "nice to have."
+
+# Alternatives
+
+## Declatative API
+Given the drawbacks of the subresource API, why not propose a declarative way to inject/eject CD-ROMs in this document? The advantages of a declarative API are obvious. It is "the Kubernetes way." Here's why:
```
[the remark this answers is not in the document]
dankenigsberg: I truly apologize! you already have this written most of this in https://github.com/kubevirt/community/pull/369/files#diff-1eb0bbc806ec90076d40390805959b650e939e77e432f9a149b28736a80e34bbR180-R184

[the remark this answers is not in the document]
mhenriks: > It can be made race free. With the suggested API, a client that wants to eject the CDROM has to check if the VM is running and choose whether to patch the VM spec or use the subresource accordingly. During this period, the VM may have been stopped/crashed/lost connectivity.

The subresource API may be called at any point in the lifecycle of a VM

design-proposals/cdrom-inject-eject.md
```diff
@@ -0,0 +1,198 @@
...
+
+### It's bigger than just CD-ROM
+Declarative inject/eject CD-ROM falls under the umbrella of Declarative Volume Hotplug. CD-ROMs are just one of the four [disk types](https://kubevirt.io/user-guide/storage/disks_and_volumes/#disks) that KubeVirt supports. The work required for supporting one disk type is roughly equal to supporting all four, requires no special considerations for CD-ROMs, and is worthy it's own design proposal. Ultimately we should support both declarative and subresource. Supporting just subresource initially would allow us to release this feature sooner. But that is not the only reason to support subresource.
```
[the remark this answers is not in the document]
mhenriks: > I see. I hoped that we can start CD-ROM afresh with a good declarative API. Would you share a bit of history? Since CD-ROMs are just another disk, why did we not consider them in the original disk hotplug API?

I don't think it was a conscious decision to exclude cd-roms. But cd-roms are not "just another disk" when it comes to the subresource API. Otherwise there would not be changes to the API proposed in this document. Currently, the subresource API deletes the disk when `removeVolume` is called. But you don't want to delete the disk when ejecting a cd-rom

> To me, changing a CDROM seems closer to storage migration than adding a new disk. Implementation-wise, maybe we can piggy-back on top of the new storage migration API (which is declarative) rather than on the disk hotplug API?

Yes, the storage migration API is declarative, which we all want eventually, but storage migration triggers a live migration which could be considered a quite heavyweight when the same outcome could be achieved with the hotplug disk pod. Live migrations could be useful for supporting hotplug of non PVC based storage though, like containerdisks.

design-proposals/cdrom-inject-eject.md
```diff
@@ -0,0 +1,198 @@
...
+
+### VirtualMachineInstances
+If declarative is the only way to go for inject/eject, `VirtualMachineInstances` will be neglected. They can stand on their own without a `VirtualMachine` owner and currently support Subresource Volume Hotplug. VMIs are also immutable to regular users and we have no intention of changing that.
```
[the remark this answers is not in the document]
mhenriks: > But let us assume this use case is important. I am not sure I understand the statement about immutability of VMI, though. The proposed API does mutate VMI, albeit via a subresource. Can you explain why it ok to modify a VMI in one way but not the other?

I think the indirection is to limit mutations to fields that can actually be updated in real time. Rather than have a complicated webhook that validates all combinations of changes, we have trusted components that are allowed to make only the supported updates to the VMI.

This is not an uncommon pattern in Kubernetes. Pod specs are immutable as well. But there are subresources like `/ephemeralcontainers` and `/resize`.  There is an alpha feature to allow declarative resizing however

Honestly, maybe we will have mutable VMIs someday but there is a lot of work required to have declarative volume update for VMIs. Much easier for VMs

mhenriks:
Closing in favor of: https://github.com/kubevirt/enhancements/pull/12

PR #371, opened 2024-12-26
design-proposal: VM video device type configuration

dasionov:
This PR introduces a design proposal to discuss about the optimal implementation for introducing a dedicated field for a video device to be used.

We want to allow VM-owners to explicitly set the video device type when needed.

Fixes #

**Checklist**

**Release note**:

```release-note
None
```

### DIFF
--- a/design-proposals/vm_video_device_type_field.md
+++ b/design-proposals/vm_video_device_type_field.md
@@ -1,0 +1,99 @@
+# Overview
+We want to allow VM-owners to explicitly set the video device type when needed.
+
+## Motivation
+Currently, the default video device type for AMD64 in KubeVirt is **VGA** or **Bochs** in some cases.
+Using `vga` video type can limit the video functionality available to virtual machines.
+For example, VGA restricts display resolution settings for Windows VMs and may not support modern workloads effectively.
+However, it is necessary to maintain compatibility for older guests that depend on VGA (e.g., RHEL 5/6).
+To address this, we propose introducing a new configuration field to allow users to set the video device type.
+
+## Goals
+The VM spec template resource should have the ability to set the video device type when needed.
+
+## Non Goals
+Change the defaulting logic
+
+## Definition of Users
+VM-owners who require specific configurations for VMIs in their environments.
+
+## User Stories
+- As a VM owner, I want to be able to set a better video device like `virtio` instead of the standard `vga`.
+
+## Repos
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### VM-Level Configuration Field
+
+#### Description
+Add a `Video` struct under `spec.template.spec.domain.devices` in the VM template schema.
+This struct will include a single field, `type`, to specify the desired video device type.
+
+The current behavior relies on the `autoattachGraphicsDevice` field, where:
+1. If `autoattachGraphicsDevice` is not specified or is set to `true`, the current logic in the virt-launcher determines the video device type.
+2. If `autoattachGraphicsDevice` is set to `false`, no graphics or video devices are attached.
+
+The proposed change ensures that users can explicitly set the `type` for the video device **only if `autoattachGraphicsDevice` is not explicitly set to `false`**. This constraint will be enforced via the validation webhook and ensures that the new field does not conflict with existing configurations.
+
+The architecture-specific logic will continue to manage the video device type configurations.
+
+#### Implementation Logic
+1. If `autoattachGraphicsDevice` is set to `false`, the `Video` field must not exist. This will be validated via the webhook.
+
+at the `addGraphicsDevice` method:
+2. If `autoattachGraphicsDevice` is set to `false`, no graphics or video devices are attached.
+3. If `Video` is explicitly specified in the VM spec, the video type from the spec is used.
+4. If `autoattachGraphicsDevice` is not specified or is `true`:
+   - Use Bochs for EFI guests on AMD64.
+   - Use VGA for BIOS guests on AMD64.
+
+The same hierarchy will apply for other architectures.
+
+#### Pros and Cons
+
+**Pros:**
+- **Granular Control:** Enables precise configuration for each VM, avoiding cluster-wide changes that could affect unrelated workloads.
+- **Default Logic:** Ensures reasonable defaults based on architecture while allowing user overrides for legacy compatibility.
+- **Scalability:** Provides flexibility for administrators and VM owners in mixed environments with diverse requirements.
+
+**Cons:**
+- **API Complexity:** Requires schema changes to the VM template API, introducing additional fields that need maintenance and testing.
+- **Administrative Effort:** Per-VM configuration may lead to higher administrative overhead if many VMs require toggling.
+
+### API Examples
+
+#### Example 1: Video Device Configured Explicitly 
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+   name: vm
+spec:
+   template:
+      spec:
+         domain:
+            devices:
+               video:
+                  type: virtio
+```
+
+## Scalability
+Overhead of the virt-launcher/qemu may be impacted.
+
+## Update/Rollback Compatibility
+Currently, the virt-launcher arch-converter manages the default video type.
+We only provide a field for the user to specify explicitly if they want something else like virtio, so backward compatibility isn't affected.
+
+## Functional Testing Approach
+* Create a VM with video.type set to `virtio`, Validate that the guest contains virtio drivers, and expect launch successfully.
+
+# Implementation Phases
+1. **Phase 1: Introduce New Field to Allow Users to Explicitly Set Their Desired Video Type**
+    - Add logic to the validation webhook to prevent the creation of VMs where `Video` is specified and `autoattachGraphicsDevice` is explicitly set to `false`.
+    - Add logic in the virt launcher to check the field's existence at `addGraphicsDevice`, before setting video device (virtio/vga).
+    - Write unit and functional tests for the new behavior.
+
+2. **Phase 2 Documentation**
+    - Update KubeVirt documentation to include examples of configuring the video device type.


design-proposals/toggle_vga_video_device.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Definition of Users
+Cluster admins managing KubeVirt deployments who need to standardize the default video device type across their cluster.
+vm-owners who require specific default configurations for VMIs in their environments.
+
+## User Stories
+- As a cluster administrator, I want to ensure my old VMs continue using VGA as the default video device type
+while transitioning new workloads to use virtio by default.
+
+- As a VM owner, I want to ensure compatibility with older VMs that require VGA by marking the kv resource accordingly before upgrading.
```
dankenigsberg, line 26: VM owners do not have permission to touch the kv resource.

Can the VM owner currently select the non-default virtio type for her VM?
I'm not convinced that a cluster-wide config makes sense here, because we are likely to have clusters that have legacy OSes that do not have the required virtio drivers, and newer OSes that would benefit from virtio.

design-proposals/toggle_vga_video_device.md
```diff
@@ -0,0 +1,96 @@
...
+
+# Design
+
+### Option 1: Annotation-Based Toggle
+**Description:** Introduce an annotation on the KubeVirt CR to specify the default video device type. The annotation kubevirt.io/default-video-device: vga can be used to default all VMs to VGA unless overridden in individual VMIs.
+
+**Pros:**
+
+- Minimal changes to the API schema.
+- Easy to implement and maintain.
```
iholder101, line 39: TBH I don't understand these pros.

Such an annotation would demand backward compatibility anyhow, and it's not much easier to implement AFAICT. To me it's equivalent to adding a non-visible field.

design-proposals/toggle_vga_video_device.md
```diff
@@ -0,0 +1,96 @@
...
+spec:
+  configuration:
+    vgaVideoDevice: true
+```
+
+## Scalability
+The proposed changes have no anticipated impact on scalability capabilities of the KubeVirt framework
+
+## Update/Rollback Compatibility
+cluster admins and vm-owners that have old vms that supports only vga video device.
```
iholder101, line 82: Can VMs change video type between reboots / live-migrations? Will it break the guest?

iholder101 (commented):
Thanks @dasionov!

design-proposals/toggle_vga_video_device.md
dasionov: AFAIK as long as the domain capabilities supports virtio as video type, switching between `vga` to `virtio` should not break the guest at all,  (e.g in upgrade scenario)
unless of course the launcher image doesn't have the `qemu-kvm-device-display-virtio-gpu` (which provide support for `virtio`) installed, in that case the guest will break immediately at the first sync.

design-proposals/vm_video_device_type_field.md
```diff
@@ -0,0 +1,73 @@
...
+which limits the functionality available to virtual machines.
+For example, VGA restricts display resolution settings for Windows VMs and may not support modern workloads effectively.
+However, it is necessary to maintain compatibility for older virtual machines that depend on VGA (e.g., RHEL 5/6).
+To address this, we propose introducing a new configuration field to allow users to set the video device type.
+
+## Goals
+The VM spec template resource should have the ability to set the video device type when needed.
+
+## Non Goals
+(Limitations to the scope of the design)
```
jean-edouard, line 15: Please populate or remove.
I think a good non-goal could be "Change the defaulting logic", to highlight the fact that virtio will be opt-in only.

design-proposals/vm_video_device_type_field.md
```diff
@@ -0,0 +1,73 @@
...
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: vm
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          videoDeviceType: virtio
```
jean-edouard, line 54: I have a feeling that we'll want more video options in the future, like acceleration settings or vram amount.
So I think it would make sense to introduce a `video` device struct, with just a `type` for now.

design-proposals/vm_video_device_type_field.md
dasionov: Thanks, I was having hard time thinking about what is a good non-goal for this, and you are absolutely right!

design-proposals/toggle_vga_video_device.md
dasionov: Thanks, I have adjusted the design accordingly, and removed the cluster-wide toggle options.

design-proposals/toggle_vga_video_device.md
```diff
@@ -0,0 +1,96 @@
+# Overview
+We want to allow cluster administrators and VM owners to toggle the video device type.
+between VGA (for compatibility with older VMs) and virtio (that will be used by default).
+
+## Motivation
+Currently, the default video device type for AMD64 in KubeVirt is VGA.
```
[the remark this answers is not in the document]
dasionov: Thanks, I adjusted the design to mention we use `bochs` too sometimes.

design-proposals/vm_video_device_type_field.md
dasionov: done

jean-edouard (approved):

design-proposals/vm_video_device_type_field.md
```diff
@@ -0,0 +1,77 @@
...
+With this change, users can explicitly set the video type, and the arch-converter will respect it.
+
+**Pros:**
+- Granular Control: Enables precise configuration for each VM, avoiding cluster-wide changes that could affect unrelated workloads.
+- Default Logic: Ensures reasonable defaults based on architecture while allowing user overrides for legacy compatibility.
+- Scalability: Provides flexibility for administrators and VM owners in mixed environments with diverse requirements.
+
+**Cons:**
+- API Complexity: Requires schema changes to the VM template API, introducing additional fields that need maintenance and testing.
+- Administrative Effort: Per-VM configuration may lead to higher administrative overhead if many VMs require toggling.
```
iholder101, line 42: BTW, perhaps this con can be addressed with instance types & preferences
FYI @lyarwood

design-proposals/vm_video_device_type_field.md
```diff
@@ -0,0 +1,77 @@
...
+kind: VirtualMachine
+metadata:
+  name: vm
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          video:
+            type: virtio
```
iholder101, line 57: Libvirt supports multiple video devices [[1](https://libvirt.org/formatdomain.html#video-devices)]:
```xml
<devices>
  <video>
    <model type='vga'/>
  </video>
  <video>
    <model type='virtio'/>
  </video>
</devices>
```

They also have the "primary" field to indicate which is the primary volume.
Since that's the case, I think we want to turn the `video` field to a slice, and perhaps add the `primary` field. In libvirt, if the primary field is not supplied, than the first video device is set as the primary, perhaps we can follow this approach.

iholder101 (commented):
Thank you @dasionov! Looks good in general

A bit off topic, but I think this PR raises an interesting question around defaults. Do we want to generally default to the most-compatible option, or the most optimized one? In any case, I think that in some point in the future we'd need to consider defaulting to virtio as I think it makes sense to do so when supported on vast majority of guests.

design-proposals/vm_video_device_type_field.md
```diff
@@ -0,0 +1,77 @@
...
+The VM spec template resource should have the ability to set the video device type when needed.
+
+## Non Goals
+Change the defaulting logic
+
+## Definition of Users
+VM-owners who require specific configurations for VMIs in their environments.
+
+## User Stories
+- As a VM owner, I want to be able to set a better video device like `virtio` instead of the standard `vga`.
```
xpivarc, line 21: This is not really an use case imho. There should be why one needs virtio or anything other than vga/bosh(partial compatibility of vga)

design-proposals/vm_video_device_type_field.md
```diff
@@ -0,0 +1,77 @@
...
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+
+### VM-Level Configuration Field
+
+**Description:** Add a `Video` struct under `spec.template.spec.domain.devices` in the VM template schema.
+This struct for now, will include a single field, type, to specify the desired video device type.
+Currently, the virt-launcher arch-converter manages the video type.
+With this change, users can explicitly set the video type, and the arch-converter will respect it.
```
xpivarc, line 33: Please describe how this behave with current behavior. Such as what happens with existing VMs in cluster, what happens if not specified, what happens when `autoattachGraphicsDevice` is set to false...

design-proposals/vm_video_device_type_field.md
```diff
@@ -0,0 +1,77 @@
...
+    spec:
+      domain:
+        devices:
+          video:
+            type: virtio
+          disks:
+```
+
+## Scalability
+The proposed changes have no anticipated impact on the scalability capabilities of the KubeVirt framework.
```
xpivarc, line 62: Overhead of the virt-launcher(or rather qemu I should say) may be impacted.

design-proposals/vm_video_device_type_field.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Update/Rollback Compatibility
+Currently, the virt-launcher arch-converter manages the default video type.
+We only provide a field for the user to specify explicitly if they want something else like virtio, so backward compatibility isn't affected.
+
+## Functional Testing Approach
+* Create a VM with video.type set to `virtio`, Validate that the domain contains virtio as video device type, and expect launch successfully.
```
xpivarc, line 69: ` Validate that the domain contains virtio as video device type` ? Can we validate this from within guest?

xpivarc (commented):
Kubevirt doesn't strive to provide any/every option available (be it by qemu/libvirt...) and rather we have sidecars that allows you to do what we don't make explicitly available. I am not convinced we should support this unless good use cases are provided and available alternative (sidecars) shows major negatives compare to have this in core.

iholder101:
> Kubevirt doesn't strive to provide any/every option available (be it by qemu/libvirt...) and rather we have sidecars that allows you to do what we don't make explicitly available. I am not convinced we should support this unless good use cases are provided and available alternative (sidecars) shows major negatives compare to have this in core.

Sidecars are problematic in many ways. They're alpha, not really supported, hard to use and do not fit production use-cases. As most virtio drivers, I'd guess that also in this case the `virtio` driver would be faster, consume less memory and would fit better to virtualized workloads in most cases.

On the other hand, I agree this proposal lacks the use-cases and research necessary to back up these assumptions.
@dasionov can you please try to find more use-cases or data that shows if and how virtio is better? I vaguely remember you mentioned something about being able to achieve better video resolutions with virtio, this sounds like a good start to me.

design-proposals/vm_video_device_type_field.md
```diff
@@ -0,0 +1,77 @@
+# Overview
+We want to allow VM-owners to explicitly set the video device type when needed.
+
+## Motivation
+Currently, the default video device type for AMD64 in KubeVirt is **VGA** or **Bochs** in some cases.
+Using `vga` video type can limit the video functionality available to virtual machines.
+For example, VGA restricts display resolution settings for Windows VMs and may not support modern workloads effectively.
+However, it is necessary to maintain compatibility for older virtual machines that depend on VGA (e.g., RHEL 5/6).
```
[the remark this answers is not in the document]
dasionov: fixed

design-proposals/vm_video_device_type_field.md
dasionov: does it make sense to allow `Video` configuration to exist while `autoattachGraphicsDevice` is set to false?
if not, then we should probably add validation rule to the vm-validation web-hook for that. 
otherwise, we can just leverage the current behavior, 

```
	if vmi.Spec.Domain.Devices.AutoattachGraphicsDevice == nil || *vmi.Spec.Domain.Devices.AutoattachGraphicsDevice {
		c.Architecture.addGraphicsDevice(vmi, domain, c)
		domain.Spec.Devices.Graphics = []api.Graphics{
			{
				Listen: &api.GraphicsListen{
					Type:   "socket",
					Socket: fmt.Sprintf("/var/run/kubevirt-private/%s/virt-vnc", vmi.ObjectMeta.UID),
				},
				Type: "vnc",
			},
		}
	}
```

The `c.Architecture.addGraphicsDevice` is the unit responsible for the video device type based on the arch, 
we can do:

```
func (archConverterAMD64) addGraphicsDevice(vmi *v1.VirtualMachineInstance, domain *api.Domain, c *ConverterContext) {

	if vmi.Spec.Domain.Devices.Video != nil {
		domain.Spec.Devices.Video = []api.Video{
			{
				Model: api.VideoModel{
					Type:  vmi.Spec.Domain.Devices.Video.Type,
					Heads: pointer.P(graphicsDeviceDefaultHeads),
				},
			},
		}
	}
	// For AMD64 + EFI, use bochs. For BIOS, use VGA
	else if c.BochsForEFIGuests && util.IsEFIVMI(vmi) {
		domain.Spec.Devices.Video = []api.Video{
			{
				Model: api.VideoModel{
					Type:  "bochs",
					Heads: pointer.P(graphicsDeviceDefaultHeads),
				},
			},
		}
	} else {
		domain.Spec.Devices.Video = []api.Video{
			{
				Model: api.VideoModel{
					Type:  "vga",
					Heads: pointer.P(graphicsDeviceDefaultHeads),
					VRam:  pointer.P(graphicsDeviceDefaultVRAM),
				},
			},
		}
	}
}
```

design-proposals/vm_video_device_type_field.md
dasionov: `readlink /sys/class/drm/card0/device/driver` command within guest outputs this: `../../../../bus/virtio/drivers/virtio_gpu` 

for cirros vm with virtio video for example

design-proposals/vm_video_device_type_field.md
xpivarc: >does it make sense to allow Video configuration to exist while autoattachGraphicsDevice is set to false?

Imho no, it does not make sense.

design-proposals/vm_video_device_type_field.md
xpivarc: Yes this is what I was trying to steer to. Please adjust the design

xpivarc:
> > Kubevirt doesn't strive to provide any/every option available (be it by qemu/libvirt...) and rather we have sidecars that allows you to do what we don't make explicitly available. I am not convinced we should support this unless good use cases are provided and available alternative (sidecars) shows major negatives compare to have this in core.
> 
> Sidecars are problematic in many ways. They're alpha, not really supported, hard to use and do not fit production use-cases. As most virtio drivers, I'd guess that also in this case the `virtio` driver would be faster, consume less memory and would fit better to virtualized workloads in most cases.

About the side-cars, I would you say you just point out that we should work towards graduation of it and improving it. For the virtio, well it is The virtualization "hw" so it better be better than emulating whatever ;) 

> 
> On the other hand, I agree this proposal lacks the use-cases and research necessary to back up these assumptions. @dasionov can you please try to find more use-cases or data that shows if and how virtio is better? I vaguely remember you mentioned something about being able to achieve better video resolutions with virtio, this sounds like a good start to me.

design-proposals/vm_video_device_type_field.md
dasionov: adjusted the design to enforce this in validation web-hook as well

design-proposals/vm_video_device_type_field.md
dasionov: done

dasionov:
/close
closing this, since it's not relevant at the moment.
Also it  has low priority, and diffs between `vga` to `virtio` are not fully clear

PR #377, opened 2025-01-16
design-proposal: limit Live Migration initial permissions to cluster-admin

tiraboschi:
In kubernetes we have two distinct pre-defined administrative roles:
- cluster-admin that is allowed to to perform any action on any resource
- admin that is allowed to read/write/amend almost all the resources in a namespace

We can assume that that granting to a developer/devops person the admin role to a namespace that he "owns" is a common practice.

In order to enqueue o cancel a live migration request, the user is supposed to create a `VirtualMachineInstanceMigration` object that is a namespaced resource that lives in
the same namespace of the VM object.
Till now, a namespace admin is allowed to create/delete `VirtualMachineInstanceMigration` objects in his namespaces.

The issue that this proposal is trying to address is that live migration is also a critical feature for cluster wide tasks such as upgrades, node drains and so on.
So today an unprivileged user can already easily delay a cluster upgrade or a node drain simply continuously enqueuing migrations or even block it by continuously deleting `VirtualMachineInstanceMigration` objects as soon as they appear on the cluster.

This proposal is about allowing, by default, only cluster-admin users to create/delete `VirtualMachineInstanceMigration` objects as a hardening measure (*principle of least privilege*).
*Please notice that this is not an API change.*

A cluster-admin will still be allowed to grant the migration role to selected users or groups.
Other actions like hotplugging that are only implicitly relying on a live migration are not going to be affected by this change.

Fixes #

Please notice that this is not an API change but simply hardening  (*principle of least privilege*) on the default KubeVirt roles.
A cluster-admin will still be allowed to grant the migration role to selected users or groups or even restore the previous behaviour.

**Checklist**

- [X] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [X] Upgrade: Impact of this change on upgrade flows was considered and addressed if required

**Release note**:

```release-note
limit Live Migration initial permissions to cluster-admin
```

### DIFF
--- a/design-proposals/hardening-vmim.md
+++ b/design-proposals/hardening-vmim.md
@@ -1,0 +1,159 @@
+# Overview
+The request to live migrate a VM is represented by an instance of VirtualMachineInstanceMigration kind.
+VirtualMachineInstanceMigration is a namespaced CRD and its instances are expected to be in the namespace of the VM they are referring to.
+As for today, by default, a namespace admin (usually a namespace "owner" in a less formal definition) is able to create VMs and also VirtualMachineInstanceMigrations (VMIMs) objects to enqueue a live migration request for a VM within his namespace.
+On the other side, live migration is also a building block for infrastructure critical operations like node drains or upgrades, so potentially every namespace admin allowed to create or delete VMIMs objects can delay or even prevent a cluster critical operation such as an upgrade.
+This proposal is about amending KubeVirt default RBAC roles according to the principle of least privilege so that namespace admins will be allowed to create or delete VirtualMachineInstanceMigrations objects only if explicitly allowed by a cluster admin.
+The present proposal is not an API change since the API will remain exactly the same, the change is only about who is allowed by default to call the API.
+
+## Motivation
+Live migration is an essential building block for cluster critical operations like node drains or upgrades.
+A less privileged user should not be allowed by default to be able to potentially interfere with infrastructure critical operations.
+
+## Goals
+- Namespace admins will not be allowed anymore by default to enqueue or delete migration request for the VMs in their namespaces
+- If needed, the right to enqueue or delete migration requests could be still granted by cluster admins to individual users or groups
+
+## Non Goals
+- This is only affecting explicit live migration requests, other operations like hot-plugging a device to a VM can rely on live migration but, in this case, the migration request will be handled by a KubeVirt controller and so is not going to be limited by this change.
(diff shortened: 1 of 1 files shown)


design-proposals/hardening-vmim.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Motivation
+Live migration is an essential building block for cluster critical operations like node drains or upgrades.
+A less privileged user should not be allowed by default to be able to potentially interfere with infrastructure critical operations.
+
+## Goals
+- Namespace admins will not be allowed anymore by default to enqueue or delete migration request for the VMs in their namespaces
+- If needed, the right to enqueue or delete migration requests could be still granted by cluster admins to individual users or groups
+
+## Non Goals
```
xpivarc, line 17: I would add that building a priority system is also out of scope as that would address some of the problems mentioned in this proposal.

design-proposals/hardening-vmim.md
```diff
@@ -0,0 +1,125 @@
...
+
+## API Examples
+No API changes
+
+## Scalability
+No concerns, it's only amending the default KubeVirt cluster roles restricting them.  
+
+## Update/Rollback Compatibility
+After an upgrade, KubeVirt operator will amend `kubevirt.io:admin`, `kubevirt.io:edit` cluster roles removing the rules that allow namespace admin to manage VMIMs so an upgraded cluster will behave exactly a fresh deployed one.
+ A cluster admin could still restore the previous behavior simply relabeling the new `kubevirt.io:migrate` clusterrole. The KubeVirt operator will ignore custom labels there honoring the cluster-admin configuration.
```
xpivarc, line 115: Note worthy would be that for smooth upgrades the admin is required to pre-deploy the cluster role and bind it.

xpivarc (commented):
Looks good overall

design-proposals/hardening-vmim.md
tiraboschi: done

design-proposals/hardening-vmim.md
tiraboschi: Personally I think that aiming to have clusters deployed with an old version and then upgraded behaving by default exactly as fresh clusters is more than desiderable in terms of maintainability and harmonization.
Otherwise a cluster admin managing two different clusters deployed starting with different versions could be surprised by a different behaviour that is hard to track and justify.
A really conscious cluster admin could still pre-deploy a temporary clusterrole to grant migration rights to all the namespace admins (labelling it with `rbac.authorization.k8s.io/aggregate-to-admin=true` ) and remove it only  when he is absolutely sure that all the users that needs to be able to trigger live migrations got bound to the new `kubevirt.io:migrate` cluster role.
I added it to the proposal.

xpivarc:
/lgtm 
@tiraboschi thank you

PR #385, opened 2025-02-07
design-proposal: Object Graph subresource

alromeros:

This Pull Request introduces a design proposal for an Object Graph API in KubeVirt, which aims to provide an authoritative way to represent VM and VMI dependencies and their relationships.

**POC:** https://github.com/kubevirt/kubevirt/pull/13875

**Release note**:

```release-note
Object Graph design proposal
```

### DIFF
--- a/design-proposals/object-graph.md
+++ b/design-proposals/object-graph.md
@@ -1,0 +1,245 @@
+## **Overview**
+
+This is a proposal to include an `Object Graph` API in KubeVirt to represent VM and VMI dependencies and their relationships.
+
+## **Motivation**
+
+As new features continue to be added to KubeVirt, the graph of objects related to VMs (DataVolumes, PersistentVolumeClaims, InstanceTypes, Preferences, Secrets, ConfigMaps, etc.) continues to expand. Identifying all the objects that a VM depends on for tasks like backup, disaster recovery, or migration can be error-prone. We should simplify this process for users and partners by creating an authoritative way to retrieve a structured object graph.
+
+## **Goals**
+
+- Define an Object Graph API to represent the dependency list of VMs and VMIs.
+- Create an API subresource to expose the dependency list.
+- Allow filtering of resources in some way.
+- Ensure extensibility for future enhancements.
+
+### **Non-Goals**
+
+- Introducing a redundant API that reimplements the VM or VMI Spec.
+
+## **User Stories**
+
+1. As a KubeVirt user, I want a clear way to retrieve all VM and VMI-related dependencies.
+2. As a backup partner, I want a way to identify a list of a VM's related objects so I can comprehensively backup and restore everything a VM needs.
+3. As a VM owner, I want to easily define an ACM-discovered application and protect my VM with disaster recovery software.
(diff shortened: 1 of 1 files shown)


design-proposals/object-graph.md
```diff
@@ -0,0 +1,245 @@
+## **Overview**
+
+This is a proposal to include an `Object Graph` API in KubeVirt to represent VM and VMI dependencies and their relationships.
+
+## **Motivation**
+
+As new features continue to be added to KubeVirt, the graph of objects related to VMs (DataVolumes, PersistentVolumeClaims, InstanceTypes, Preferences, Secrets, ConfigMaps, etc.) continues to expand. Identifying all the objects that a VM depends on for tasks like backup, disaster recovery, or migration can be error-prone. We should simplify this process for users and partners by creating an authoritative way to retrieve a structured object graph.
```
dankenigsberg, line 7: > We should simplify this process

Would you add a few word about the current process?
Could it be that we already use `ownerReferences` to reflect this information? It would be a problem to duplicate the information.

alromeros: It's not like there’s a "standardized" process, currently backup or migration solutions have to manually determine and include the objects they believe are related to a VM, without access to an authoritative source of truth.

Regarding ownerReferences, there are several VM dependencies that may lack them, like PVCs or DVs referenced in the Volumes section but not created using datavolumeTemplates.

dankenigsberg: Thanks. If it is not standard, it is even more important to review the current process. We need to see if we can phase it out when introducing a new API, as we do not want to have two possibly-contradicting processes.

alromeros: Hey @dankenigsberg, sorry for the late response. Yeah I agree, for example, we've been doing something similar in the kubevirt velero plugin and we plan to deprecate the current method in favor of this structured graph. @awels is also planning to integrate this in his cluster migration feature, so this would provide a unified standard instead of duplicating logic across implementations. We are not aware if other software is using its own thing to fetch the vm dependencies, but the idea is to provide this api as the single source of truth.

design-proposals/object-graph.md
```diff
@@ -0,0 +1,245 @@
...
+
+This format visualizes dependencies in a tree structure, showing parent-child relationships between resources.
+
+```go
+// ObjectGraphNode represents an individual resource node in the graph.
+type ObjectGraphNode struct {
+	ObjectReference k8sv1.TypedObjectReference `json:"objectReference"`
+	Labels          map[string]string          `json:"labels,omitempty"`
+	Optional        bool                       `json:"optional"`
+	Children        []ObjectGraphNode          `json:"children,omitempty"`
```
mhenriks, line 52: I know VM object graphs probably won't be huge but I wonder about this getting too nested. Maybe have Children be []ObjectReference instead?

mhenriks:
This should probably be in the enhancements repo

alromeros:
Migrated to https://github.com/kubevirt/enhancements/pull/42 as suggested by @mhenriks.

alromeros:
Closing this PR, will continue the discussion in https://github.com/kubevirt/enhancements/pull/42. Thanks for the reviews!

PR #401, opened 2025-04-14
sigs.yaml: add sig-compute meetings

iholder101:
Adds sig-compute meetings to sig.yaml.

Also, add the "sig-" prefix to sig names, like sig-network or sig-storage.

**Release note**:

```release-note
sigs.yaml: add sig-compute meetings
```

### DIFF
--- a/sigs.yaml
+++ b/sigs.yaml
@@ -1,5 +1,5 @@
 sigs:
-    - name: sig-compute
+    - name: compute
       dir: sig-compute
       label: sig/compute
       leadership:
--- a/sigs.yaml
+++ b/sigs.yaml
@@ -10,6 +10,14 @@ sigs:
             - github: jean-edouard
               name: Jed Lejosne
               company: Red Hat
+      meetings:
+        - description: KubeVirt Compute SIG Meeting
+          day: Wednesday
+          time: 2:30 PM
+          tz: UTC
+          frequency: weekly
+          url: https://zoom.us/j/98091854677
+          archive_url: https://docs.google.com/document/d/1VTeHa-QxppNcDQbpoddrucBz1vXjxWB0JLARfS5Uuqs
       contact:
         slack: kubevirt-dev
         mailing_list: https://groups.google.com/forum/#!forum/kubevirt-dev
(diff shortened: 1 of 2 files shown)


sigs.yaml
```diff
@@ -81,7 +89,7 @@ sigs:
             - https://raw.githubusercontent.com/kubevirt/kubevirt.github.io/main/OWNERS
             - https://raw.githubusercontent.com/kubevirt/user-guide/main/OWNERS
             - https://raw.githubusercontent.com/kubevirt/kubevirt-tutorial/main/OWNERS
-    - name: storage
+    - name: sig-storage
```
dhiller, line 92: nit: since this is inside the `sigs:` section the sig repetition feels kind of redundant, however I don't feel strongly either way.

iholder101: Thanks for explaining the reasoning behind this! I've missed the sigs section :)

I mainly tried to keep things consistent. Would you prefer if I'd remove the `sig-` prefix from compute?

aburdenthehand: It's probably better if we do drop the sig prefix from that column, mostly just to make the label clearer. With the prefix, it's a single character difference.

sigs.yaml
```diff
@@ -10,6 +10,14 @@ sigs:
...
               name: Jed Lejosne
               company: Red Hat
+      meetings:
+        - description: KubeVirt Compute SIG Meeting
+          day: Wednesday
+          time: 2:30 PM
+          tz: UTC
+          frequency: weekly
+          url: https://zoom.us/j/98091854677
+          archive_url: https://docs.google.com/document/d/1VTeHa-QxppNcDQbpoddrucBz1vXjxWB0JLARfS5Uuqs/edit?usp=sharing
```
jean-edouard, line 20: Nit:
```suggestion
          archive_url: https://docs.google.com/document/d/1VTeHa-QxppNcDQbpoddrucBz1vXjxWB0JLARfS5Uuqs
```

jean-edouard (commented):
Added my own nit, but PR looks good, feel free to unhold.
/approve
/hold

sigs.yaml
iholder101: Done! :)
PTAL

sigs.yaml
iholder101: Thanks for catching that! :)
/unhold

aburdenthehand:
/lgtm
Thanks @iholder101 :+1:

PR #420, opened 2025-09-10
Removing dvossel and adding missing maintainers from community approver list

aburdenthehand:
David is no longer with the project, and it seems only right that the project maintainers have permissions to approve PRs in the community repo, since it is primarily concerned with shaping the processes and direction of the project. 

Modifying the approver list to match this. 

FYI @rthallisey @stu-gott @xpivarc in case you disagree :)

**Release note**:

```release-note
NONE
```

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -6,11 +6,13 @@ aliases:
 
   approvers:
   - aburdenthehand
-  - davidvossel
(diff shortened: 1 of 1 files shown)


jean-edouard (approved):
Makes sense to me!

dhiller (approved):

PR #444, opened 2026-05-28
Adopt documentation-first approach for AI context

iholder101:
Replaces the `AGENTS.md`/vendor-specific prompt file approach with an agent-agnostic, documentation-first strategy for providing context to AI tools.

Repositories are encouraged to maintain a documentation index file linking existing docs by topic [1], keeping project knowledge in standard markdown that serves both humans and AI tools equally. A clear separation between agent-agnostic documentation and vendor-specific tool configuration is established, and local development tool choice is left to individual contributors.

[1] This pattern is adopted by other projects (e.g., Anthropic's [docs map](https://code.claude.com/docs/en/claude_code_docs_map.md)).

**Release note**:

```release-note
NONE
```

### DIFF
--- a/ai-contribution-policy.md
+++ b/ai-contribution-policy.md
@@ -73,11 +73,9 @@ acceptable and need not be changed to `Assisted-by`.
 Authors **MUST** still adhere to KubeVirt's Developer's Certificate of Origin
 (DCO) requirements and sign off commits.
 
-This will be aided through the use of the emerging
-[AGENT.md](https://ampcode.com/AGENT.md) standard with symlinks provided to the
-in project prompt configuration files of various agents. An example file will
-be created within the [`kubevirt/.github`](https://github.com/kubevirt/.github)
-repository for projects to use as a base.
+Disclosure requirements are documented in project contribution guidelines,
+making them discoverable by both human contributors and AI tools alike (see
+[Providing Context to AI Tools](docs/ai-tooling-guidelines.md)).
 
 ### Scope of Disclosure
 
@@ -205,6 +203,5 @@ repository](https://github.com/kubevirt/community)
 - [QEMU Code Provenance
   Policy](https://www.qemu.org/docs/master/devel/code-provenance.html#use-of-ai-content-generators)
 - [Ghostty AI Policy](https://github.com/ghostty-org/ghostty/blob/main/AI_POLICY.md)
-- [AGENT.md](https://ampcode.com/AGENT.md)
 - [KubeVirt Developer's Certificate of
   Origin](https://github.com/kubevirt/kubevirt/blob/main/DCO)
--- a/docs/ai-tooling-guidelines.md
+++ b/docs/ai-tooling-guidelines.md
@@ -1,0 +1,56 @@
+# Providing Context to AI Tools
+
+## Documentation-First Approach
+
+KubeVirt adopts an agent-agnostic, documentation-first approach for providing
+context to AI tools. Project knowledge - contribution guidelines, coding
+patterns, architectural decisions, review policies - lives in standard
+documentation files readable by both humans and AI agents.
+
+This approach is preferred over maintaining agent-specific configuration files
+(e.g., vendor-specific prompt files or rule sets) for several reasons:
+
+- The community's expertise is in writing good documentation, not in
+  maintaining agentic configurations
+- Good documentation serves both humans and AI tools simultaneously
+- It avoids vendor lock-in to any specific AI tool or platform
+- It reduces maintenance burden by having a single source of truth
+
+## Repository Documentation Index
+
+Repositories SHOULD provide a documentation index file that serves as a table
+of contents, linking relevant documentation by topic. AI tools can load this
+single file to gain awareness of available project documentation and pull in
+specific files as needed for the task at hand.
+
+This pattern is adopted by other projects (e.g., Anthropic's
+[docs map](https://code.claude.com/docs/en/claude_code_docs_map.md)) and keeps
+project knowledge in a form that is useful regardless of which AI tool - or
+human - accesses it.
+
+Repository maintainers may choose to additionally maintain an `AGENTS.md`
+file, but it is strongly encouraged that it link to a documentation index or
+other form of pointers to human-first documentation rather than encoding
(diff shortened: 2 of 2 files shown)


EdDev (commented):
Thank you for this direction, replacing the `AGENT.md` reference with a documentation-first approach is a cleaner long-term strategy.

One structural point worth discussing: the new "Providing Context to AI Tools" section feels like it belongs in a dedicated file rather than inline here.

The existing `ai-contribution-policy.md` is consistently contributor-facing: it tells people what to disclose, what they're responsible for, and how reviewers should evaluate AI-assisted work. Every section addresses the same actor: a contributor submitting a PR.

The new section addresses a different actor and a different moment in time: it's repository-maintainer guidance for how to set up a repo to work well with AI tools. That's a meaningful shift in audience and scope.

There's also a practical concern: the section names concrete artifacts, i.e. `AGENTS.md`, `.coderabbit.yaml`, documentation index files. These conventions are still evolving and will need updates as tooling changes. Embedding them in a policy document means the policy gets touched for what are really tooling-hygiene updates, which is friction and noise. A dedicated `ai-tooling-guidelines.md` can evolve at its own cadence.

The PR already does the right structural thing on the contributor-policy side, the cross-reference (see [Providing Context to AI Tools](#...)) in the
Disclosure section. It just needs to point outward to a separate file rather than downward within the same document.

How about this:
1. Move the new section into a new dedicated file (e.g., `ai-tooling-guidelines.md`)
2. Update the cross-reference in the Disclosure section to point to that file.

This keeps the policy document focused on its original purpose, and gives the tooling guidance room to grow without polluting the policy's change history.

ai-contribution-policy.md
```diff
@@ -173,6 +171,63 @@ reviewers should evaluate:
...
+- **Tool configuration** (vendor-specific): operational settings such as noise
+  control, review profiles, and file routing. These live in tool-specific
+  configuration files (e.g., `.coderabbit.yaml`) and control only tool
+  behavior.
+
+Tool configuration files SHOULD reference documentation files for
+project-specific context rather than duplicating knowledge inline.
+
+Repository-level tool configuration applies only to tools that the repository
+has officially adopted (e.g., CodeRabbit for automated code review).
```
aburdenthehand: Do we have the officially adopted tools listed somewhere we can link to?

ai-contribution-policy.md
```diff
@@ -173,6 +171,63 @@ reviewers should evaluate:
...
+
+- The community's expertise is in writing good documentation, not in
+  maintaining agentic configurations
+- Good documentation serves both humans and AI tools simultaneously
+- It avoids vendor lock-in to any specific AI tool or platform
+- It reduces maintenance burden by having a single source of truth
+
+### Repository Documentation Index
+
+Repositories SHOULD provide a documentation index file that serves as a table
```
aburdenthehand: Are we expecting consistency with the file names and locations across the repos? If so, we should be explicit about that.

ai-contribution-policy.md
```diff
@@ -173,6 +171,63 @@ reviewers should evaluate:
...
+of contents, linking relevant documentation by topic. AI tools can load this
+single file to gain awareness of available project documentation and pull in
+specific files as needed for the task at hand.
+
+This pattern is adopted by other projects (e.g., Anthropic's
+[docs map](https://code.claude.com/docs/en/claude_code_docs_map.md)) and keeps
+project knowledge in a form that is useful regardless of which AI tool - or
```
aburdenthehand: In the old days I would view this as a major maintenance concern. Am I right in guessing that the plan is to use an LLM to generate this map based on a comprehensive docs scan? What does maintaining it look like?

aburdenthehand (commented):
@iholder101 Thanks for submitting this. This was new for me and I think I would need to explore it more in order to provide better feedback, but I did have a couple of questions after reading through it. 

Edit: I think I agree with @EdDev about separation of concerns however I think this is also important for the contributor using AI to understand as well. It needs to be easily discoverable, which I think is what the AGENTS.md convention was aimed at iiuc.

ai-contribution-policy.md
iholder101: It is being discussed as part of https://github.com/kubevirt/enhancements/issues/297, where it's planned to replace SourceryAI by CodeRabbit. I don't think there's any other list

ai-contribution-policy.md
iholder101: I imagine something much simpler. We can decide that documentation files should container a "frontmatter" metadata section with a short description/summary of the content (about a paragraph long).

Let's say that we have theoretical a file under `./docs/developer-guidelines/coding-style-and-conventions.md`. It would look like:

```
---
description: "This file defines the Go conventions
and coding style that should be used by KubeVirt developers."
---

<Actual full content>

```

Then, a simple script could update the documentation map, which can look something like:
```
## docs

### developer-guidelines

This file defines the Go conventions
and coding style that should be used by KubeVirt developers.
./docs/developer-guidelines/coding-style-and-conventions.md

This is another file with a different
summary that captures its contents
./docs/developer-guidelines/another-file.md
```

(this is of course just an example, we can discuss the exact format).

So the file path should be used to define the doc's position in the subject tree, which paths should already reflect anyways. And the frontmatter metadata can be used in order to provide the short summary. While the doc writer can obviously use LLMs to help generate the summary, eventually the responsibility of creating it is (like anything else) on the PR owner and reviewers. We should probably eventually also enforce this with an automation.

Does this make sense? WDYT?

EdDev (approved):
Thank you!

ai-contribution-policy.md
iholder101: According to https://github.com/kubevirt/community/pull/444#discussion_r3503828431 I don't think that's necessary.

In addition, this document is more of a general statement of how we should pursue making KubeVirt better for AI agents (and humans as well). I wonder if this should be followed up with a tracker issue / VEP / something else to actually track and plan the progress of this effort.

enp0s3 (approved):
/approve

@iholder101 Thank you, makes total sense to me.

@aburdenthehand @jean-edouard @dhiller Would you be able to review?

